PR #338, opened 2022-09-22
Use protobuf and gRPC

bartekpacia:
- fix #316
- fix #322
- fix #357
- fix #321

PR #345, opened 2022-09-22
Implement enable/disable dark mode on iOS 

### LINKED ISSUE #320: Toggle dark mode on iOS

bartekpacia:
fix #320 

TODO:
- [x] simulator
- [x] physical device
- [x] return to the app under test after enabling/disabling dark mode

### DIFF
--- a/AutomatorServer/ios/AutomatorServerUITests/PatrolServer.swift
+++ b/AutomatorServer/ios/AutomatorServerUITests/PatrolServer.swift
@@ -34,7 +34,9 @@ struct Selector: Codable {
   var text: String
 }
 
-struct EnableDar
+struct DarkModeCommand: Codable {
+  var appId: String
+}
 
 class PatrolServer {
   private static let envPortKey = "PATROL_PORT"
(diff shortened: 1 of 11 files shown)


AutomatorServer/ios/AutomatorServerUITests/PatrolServer.swift
```diff
@@ -34,6 +34,10 @@ struct Selector: Codable {
   var text: String
 }
 
+struct DarkModeCommand: Codable {
+  var appId: String
```
iasiu, line 38: Since in dart part you call this `bundleId` I'd prefer to keep the naming consistent

iasiu (approved):

AutomatorServer/ios/AutomatorServerUITests/PatrolServer.swift
bartekpacia: This command also exists on Android and there we'd have to name it `packageName`, but this would require us to make different requests for every platform with no real benefit. IGL it's okay.

Anyway this won't be a problem once we migrate the communication to protobufs and gRPC in #338.

PR #383, opened 2022-09-28
Fix granting iOS permissions not covering all cases

### LINKED ISSUE #382: Fix granting permissions on iOS not working in all cases
First reported by @iasiu in the customer project, then reported independently [here](https://github.com/leancodepl/patrol/issues/354#issuecomment-1260475603).

bartekpacia:
fix #382 

> Take this and drink from it: this is the cup of my blood.

### DIFF
--- a/AutomatorServer/ios/AutomatorServerUITests/PatrolAutomation.swift
+++ b/AutomatorServer/ios/AutomatorServerUITests/PatrolAutomation.swift
@@ -160,14 +160,16 @@ class PatrolAutomation {
   func allowPermissionWhileUsingApp() {
     runAction("allowing while using app") {
       let systemAlerts = self.springboard.alerts
-      let okButton = systemAlerts.buttons["OK"]
-      if okButton.exists {
-        okButton.tap()
-      }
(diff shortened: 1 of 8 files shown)


AutomatorServer/ios/AutomatorServerUITests/PatrolAutomation.swift
```diff
@@ -160,24 +160,33 @@ class PatrolAutomation {
   func allowPermissionWhileUsingApp() {
     runAction("allowing while using app") {
       let systemAlerts = self.springboard.alerts
-      let okButton = systemAlerts.buttons["OK"]
-      if okButton.exists {
-        okButton.tap()
-      }
+      let labels = ["OK", "Allow", "Allow While Using App"]
```
iasiu, line 163: Shouldn't `"Allow While Using App"` be first?

iasiu (approved):

AutomatorServer/ios/AutomatorServerUITests/PatrolAutomation.swift
bartekpacia: It doesn't matter.

PR #397, opened 2022-09-30
Refactor the main `CommandRunner` and add more tests

### LINKED ISSUE #269: Improve test coverage of the CLI

bartekpacia:
This PR:
- fixes #269
- removes most of global dependencies, instead we're injecting them now through constructors*
- replaces most of dependencies on `dart:io` with `package:file` and `package:platform`

\*As the tool grows it might be worth taking a look at some more advanced DI solutions like [get_it](https://pub.dev/packages/get_it) and [injectable](https://pub.dev/packages/injectable), but I had only bad experiences with them and I hate code generation. Let's keep stuff simple.


packages/patrol_cli/lib/src/common/artifacts_repository.dart
```diff
@@ -3,30 +3,33 @@ import 'package:file/file.dart';
...
+        debug = false {
     _paths = _Paths(artifactPath);
   }
 
   static const artifactPathEnv = 'PATROL_CACHE';
 
   final FileSystem _fs;
-  final Platform _platform;
-  final TopLevelFlags _topLevelFlags;
+  Platform platform;
```
shilangyu, line 24: Not final?

packages/patrol_cli/lib/src/common/artifacts_repository.dart
```diff
@@ -3,30 +3,33 @@ import 'package:file/file.dart';
...
+  final http.Client _httpClient;
+  final ZipDecoder _zipDecoder;
+  bool debug;
```
shilangyu, line 27: Not final?

packages/patrol_cli/lib/src/common/extensions/process.dart
```diff
@@ -1,18 +1,18 @@
...
+import 'dart:io' show Process, ProcessResult, systemEncoding;
 
 import 'package:dispose_scope/dispose_scope.dart';
 
 extension ProcessResultX on ProcessResult {
   /// A shortcut to avoid typing `as String` every time.
   ///
   /// If [stdout] is not a String, this will crash.
-  String get stdOut => this.stderr as String;
+  String get stdOut => stderr as String;
```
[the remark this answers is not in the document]
bartekpacia: It should, lol. Thx!

packages/patrol_cli/lib/src/common/artifacts_repository.dart
bartekpacia: Nope, I set it after I know whether the user passed `--debug` (which is after the `ArtifactsRepository` is instantiated).

packages/patrol_cli/lib/src/common/artifacts_repository.dart
bartekpacia: Nope, this makes it easier to change platform in tests without creating a new object.

PR #413, opened 2022-10-05
Fix `denyPermission()` not working on iOS

bartekpacia:
Reported by @iasiu in the customer project.


shilangyu (approved):
Won't that break if the phone is in a different language?

iasiu (approved):

PR #433, opened 2022-10-10
Use `console` syntax and dollar sign for shell prompts

bartekpacia:
let' see how docs.page handles this

might have to revert


bartekpacia:
Using `console` syntax breaks local preview.

bartekpacia:
As I thought...

https://github.com/invertase/docs.page/issues/232

PR #448, opened 2022-10-13
Implement `isPermissionDialogVisible()`

### LINKED ISSUE #445: Add new `isPermissionDialogPresent()` RPC
Example use case: I want to `grantPermissionWhenInUse()` for notifications permission only when on Android API 33 and higher, because notification permission is granted implicitly for all apps on older Android versions.

bartekpacia:
This PR:
- fixes #445


contracts.proto
```diff
@@ -121,6 +122,14 @@ message SetLocationAccuracyRequest {
   }
 }
 
+message PermissionDialogVisibleRequest {
+  uint64 timeout = 1;
```
shilangyu, line 126: its unclear what this timeout's unit

shilangyu (approved):

contracts.proto
bartekpacia: Renamed to `timeoutMillis`. Thanks.

PR #458, opened 2022-10-15
Set correct exit code of Patrol CLI and test it

bartekpacia:
- set exit code 1 when any test fails
- write tests for DriveCommand exit code (failing)
- inject AndroidDriver and IOSDriver into DriveCommand
- inject FlutterDriver into DriverCommand, make all tests pass

### DIFF
--- a/packages/patrol_cli/lib/src/features/drive/drive_command.dart
+++ b/packages/patrol_cli/lib/src/features/drive/drive_command.dart
@@ -1,7 +1,6 @@
 import 'package:dispose_scope/dispose_scope.dart';
 import 'package:freezed_annotation/freezed_annotation.dart';
 import 'package:logging/logging.dart';
-import 'package:patrol_cli/src/common/artifacts_repository.dart';
 import 'package:patrol_cli/src/common/extensions/map.dart';
 import 'package:patrol_cli/src/common/staged_command.dart';
 import 'package:patrol_cli/src/features/devices/device_finder.dart';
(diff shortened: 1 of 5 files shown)


packages/patrol_cli/lib/src/features/drive/drive_command.dart
```diff
@@ -33,16 +32,20 @@ class DriveCommandConfig with _$DriveCommandConfig {
 class DriveCommand extends StagedCommand<DriveCommandConfig> {
   DriveCommand({
     required DisposeScope parentDisposeScope,
-    required ArtifactsRepository artifactsRepository,
     required DeviceFinder deviceFinder,
     required TestFinder testFinder,
     required TestRunner testRunner,
+    required AndroidDriver androidDriver,
+    required IOSDriver iosDriver,
```
shilangyu, line 39: Wouldnt it be cleaner if you were to always accept only a single driver (which has some common interface) and make this `TargetPlatform` check when creating this DriveCommand? It would make it easier to mock too, since you would not need to provide always two implementations when in the end only one is used anyways

packages/patrol_cli/lib/src/features/drive/flutter_driver.dart
```diff
@@ -49,22 +29,41 @@ class FlutterDriver {
     _disposeScope.disposedBy(parentDisposeScope);
   }
 
+  late String driver;
```
shilangyu, line 32: I dont like this `late` since there is no guarantee it was initialized (other than "I promise I checked all paths")

shilangyu (approved):

packages/patrol_cli/lib/src/features/drive/drive_command.dart
bartekpacia: Actually, we need two drivers, because the user can run the tests on more than a single device (e.g 1 device can be an iPhone and the other one can be an Android).

packages/patrol_cli/lib/src/features/drive/flutter_driver.dart
bartekpacia: I agree, but I didn't know how to do it better.

I have to be able to mock the `FlutterDriver` that `DriveCommand` uses. I can achieve this by passing an instance of it to `DriveCommand`.

`FlutterDriver` needs some information (`host`, `port`, `driver`, etc.) that is not known at the time of `DriveCommand` construction, so I have to pass that data at a later point. Previously this later point was `FlutterDriver.run()`, but I prefer that method accept only `String target` and `Device device`, that is: data unique to each invocation of it.

That's why I decided to have an `init()` method.

I'd appreciate if you looked at the code, maybe you'd have a better idea on how clean this up.

Issue #278, opened 2022-09-13
Research if we can distribute the prebuilt `.ipa`/`.app`


bartekpacia:
Currently, we're building the iOS AutomatorServer (`AutomatorServer/ios`) at every `patrol drive`, which adds some overhead (usually a few up to 10-20 seconds). In the ideal case, we should get rid of this and just download prebuilt `.ipa` or `.app` (I always forget which one is for what).

To keep in mind:
- simulator binaries need to be distributed as `x86_64` and `arm64e`
- physical iOS devices require the binary to be signed

bartekpacia:
Current problem: [GitHub Actions don't support running macOS VMs on Apple Silicon](https://github.com/actions/runner-images/issues/2187), so we can only build x86_64 artifact for a simulator, which when run on a simulator on Apple Silicon mac, only shows black screen and/or crashes (just random, unpredictable behavior).

bartekpacia:
**Idea**
We could just distribute unsigned `.app`, and then tell our users to codesign it manually. If we give good instructions, it should be a quite painless process.

Maybe with a tool like `codesign`?

bartekpacia:
Appium apparently doesn't codesign their binaries:
- https://github.com/appium/WebDriverAgent/blob/master/.github/workflows/publish.js.yml

PR #465, opened 2022-10-17
CLI - add support for prebuilt iOS artifacts

### LINKED ISSUE #462: Automatically download artifacts on `patrol update`
Currently `patrol update` only updates `patrol_cli` to the latest version.

We should also download artifacts after updating the CLI, because in practice the user always does this after updating the CLI.

bartekpacia:
This PR:
- adds support for [using prebuilt artifacts on iOS](https://github.com/leancodepl/patrol/issues/278), which improves the speed and reliability of running tests on iOS.
- fixes #462, small and plays nicely with changes done here

More about prebuilt artifacts:
- the CLI will use prebuilt artifacts when running on `iphonesimulator` (Simulator.app)
- the CLI will still rebuild artifacts on every test run when running on `iphoneos`(physical iOS devices), because of problems with code signing that will be resolved in a separate PR. 
- some refactoring was done as well


packages/patrol_cli/lib/src/common/artifacts_repository.dart
```diff
@@ -1,10 +1,125 @@
...
+
+  const factory Artifact.archive({
+    required String name,
+    String? version,
+    String? ext,
+  }) = _ArtifactArchive;
+
+  const Artifact._();
+
+  /// Nmae of the file once downloaded and extracted.
```
shilangyu, line 79: Typo

shilangyu (approved):

PR #488, opened 2022-10-20
Build iOS artifact for real iPhones

bartekpacia:
- make_artifacts: turn off codesigning only for iphonesimulator
- status: `make_artifacts` works after `fastlane match`


bartekpacia:
Useful links:
- https://discuss.bitrise.io/t/xcode-build-for-testing-step-issues/11134/3

bartekpacia:
I'm blocked because `fastlane match` needs access to App Store Connect to verify if the profiles and certificates are valid, and I currently don't have an App Store Connect API key to make it possible.

bartekpacia:
It's not possible.

If we want to distribute an iOS app (an our automation server is an iOS app), it has to be done through App Store. Otherwise we might use TestFlight or alternatives (AppCenter, Firebase App Distribution) but then we have to explicitly specify which devices the app can run on, which is a no-go for our case.

Appium has the same problem.

To sum up, the current status of running tests on iOS:
- on iOS simulators, we can use the prebuilt, unsigned app
- on iOS devices, the user has to manually build the app from source and sign it in Xcode. We need to update docs to state this.

cc: @mateuszwojtczak

PR #497, opened 2022-10-22
Investigate the issue with scroll spinning right round

### LINKED ISSUE #475: `scrollTo()` "spins" (tries to scroll in place and doesn't detect the widget that is hit testable)
`scrollTo()` sometimes doesn't detect the widget that is hit testable and spins in place. This slows tests down.

It's a strange and quite rare prurrently seen only in the custom project that @jBorkowska is working on.

bartekpacia:
First reported by @jBorkowska in the customer project.

Fixes #475

Demo (two tests, the first one is failing, the second one is passing because a fix is applied):

https://user-images.githubusercontent.com/40357511/197309642-a8027ec6-d3c4-4cba-9fe7-c0b54253ee73.mov


bartekpacia:
An issue on `flutter/flutter` that is loosely related and some comments in it are valuable:
- [[Docs]Tests to scroll a ListView and tap on a child #88762](https://github.com/flutter/flutter/issues/88762)

bartekpacia:
Looks like there's no perfect solution to this "scroll going crazy" problem. I think the best we can do is to educate our users on this matter.

bartekpacia:
**TODO**

PatrolTester should print a warning if more than 1 `Scrollable` exists but the user didn't explicitly specify which `Scrollable` to use.

bartekpacia:
@jBorkowska, what do you think about [this](https://github.com/leancodepl/patrol/pull/497#issuecomment-1289774934)?

bartekpacia:
Followed up in #531

jBorkowska:
I'm thinking about this idea to warn user if more `Scrollable`s found. Docs are cool, but not everybody reads them (eg. me :)). I remember situations, when I got so mad at some frameworks or tools bc I didn't read all of the docs (the info was hidden in a place I didn't expect). On the other hand it can be annoying to see those warnings in logs. Anyway we should try it in my opinion.

PR #514, opened 2022-10-25
Add example file picker screen and test

bartekpacia:
For @fylyppo with 💙

This test assumes that there's a file named `download.jpg` in the Downloads directory. 

https://user-images.githubusercontent.com/40357511/197803135-f1baeec1-466b-4ab8-a66f-d999cb49dd69.mov

### DIFF
--- a/packages/patrol/example/integration_test/features/file_picker_test.dart
+++ b/packages/patrol/example/integration_test/features/file_picker_test.dart
@@ -21,6 +21,7 @@ void main() {
 
       await $.native.tap(Selector(text: 'download.jpeg'));
       await $.pumpAndSettle();
+      await $.pumpAndSettle();
 
       expect($(#image_0), findsOneWidget);
     },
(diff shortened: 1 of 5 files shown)


jBorkowska:
I think it is developer's responsibility to have this in mind. There is nothing we can do.

pietraniktx:
I'm not sure about the "signaling" part of the question but I understand the problem is - "device needs to have an image in downloads folder, if not, the test will fail", right? So why can't we download the file before the test or put it there some other way programmatically?

jBorkowska:
"We" the framework or "we" the testers? :D

pietraniktx:
:D I belive in Bartek. And it would be nice for Patrol users to have the recipe for such scenario as it looks like quite common problem and preparing test data inside the test seems like a good practice. Cypress, for example, has a ton of recipes to help testers with cases like this, check out this one about [seeding database with node](https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/server-communication__seeding-database-in-node)

bartekpacia:
Closing so it doesn't clutter opened PRs.

neiljaywarner:
adb screenshot or similar?

PR #521, opened 2022-10-25
Re-add progress animations

bartekpacia:
The bug in `mason_logger` was fixed in https://github.com/felangel/mason/pull/549, so I'm bringing the animations back.

Animations were removed in #498.


bartekpacia:
Unfortunately it's still not ideal, but much better than before:

Screenshot 2022-10-26 at 1 21 13 AM

Gotta decide whether this is good enough.

PR #552, opened 2022-11-02
Add flake detection mode to `patrol drive`

### LINKED ISSUE #519: Introduce "flake detection" mode
Alternative title: **make it faster to run the same test multiple times**

### The problem

Flaky tests are bad – we know it. The standard way to detect them is to run the test many times. Currently it can be done in Patrol by simply running `patrol_cli` in the loop – we show our users how to do it in the [Tips & Tricks section in the docs](https://patrol.leancode.co/tips-and-tricks#running-a-test-many-times).

This works, but the performance could be much better. We rebuild the app on every test run, even though no changes are made.

### Suggested solution

We could build the app once, before the first test run, and then reuse that build in the subsequent test runs.

**Goals:**
- add the optional `-n <int>` argument to `patrol drive`, which specifies the number of times a test run should be repeated
  - the `-n` flag can be used only when a single `--target` is given. Otherwise, we print a friendly error message ("flake detection mode is not yet supported for multiple targets") and exit with code 1.

We could possibly allow for multiple `--targets`, but this will make the implementation more complex. We'd also have to decide on the order in which to build and run the tests. I think this can be postponed for later, especially since I feel that usually, our users will try to detect flakiness in a single test at once, not multiple. Also, I don't see any big benefits it'd have over calling (the hypothetical) `patrol drive -t <target> -n <int>` in a loop (just like we do currently).

**Non-goals:**
- detect if any files were changed during the test runs

  We don't care about this, it's the user's responsibility.
- report the results in a nice way

  This is important but I feel it fits in a separate issue/PR pair.

cc: @jBorkowska @jakubfijalkowski @mateuszwojtczak @shilangyu – I'd love to know what you think about this feature.

bartekpacia:
Fixes #519 

Checked
- example app (✅Android,✅iOS)
- customer: budowa (Android)

### DIFF
--- a/packages/patrol_cli/lib/src/features/drive/drive_command.dart
+++ b/packages/patrol_cli/lib/src/features/drive/drive_command.dart
@@ -183,6 +183,10 @@ class DriveCommand extends StagedCommand<DriveCommandConfig> {
       throw const FormatException('`repeat` argument is not an int');
     }
 
+    if (repeat < 1) {
+      throwToolExit('repeat count must not be smaller than 1');
+    }
+
     if (repeat != 1) {
       if (targets.length != 1) {
         throwToolExit('only single test target runs can be repeated');
--- a/packages/patrol_cli/lib/src/features/drive/drive_command.dart
+++ b/packages/patrol_cli/lib/src/features/drive/drive_command.dart
@@ -251,24 +251,25 @@ class DriveCommand extends StagedCommand<DriveCommandConfig> {
           return;
         }
 
-        try {
-          await _flutterDriver.build(target, device);
+        await _flutterDriver.build(target, device);
 
-          for (var i = 0; i < config.repeat; i++) {
-            if (_disposeScope.disposed) {
-              break;
-            }
-
-            await _flutterDriver.run(target, device);
+        for (var i = 0; i < config.repeat; i++) {
+          if (_disposeScope.disposed) {
+            _logger.fine('Skipping running repeated $target ($i)...');
+            break;
           }
-        } on FlutterDriverFailedException catch (err) {
-          exitCode = 1;
-          _logger
-            ..severe(err)
-            ..severe(
-              "See the logs above to learn what happened. If the logs above aren't "
-              "useful then it's a bug – please report it.",
-            );
+
+          try {
+            await _flutterDriver.run(target, device);
+          } on FlutterDriverFailedException catch (err) {
+            exitCode = 1;
+            _logger
+              ..severe(err)
+              ..severe(
+                "See the logs above to learn what happened. If the logs above aren't "
+                "useful then it's a bug – please report it.",
+              );
+          }
         }
       });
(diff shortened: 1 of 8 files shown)


packages/patrol_cli/lib/src/features/drive/flutter_driver.dart
```diff
@@ -161,20 +243,31 @@ class FlutterDriver {
       }
     }
 
+    String getApplicationBinaryPath() {
+      if (platform == TargetPlatform.android) {
+        const prefix = 'build/app/outputs/flutter-apk';
+        if (flavor != null) {
+          return join(prefix, 'app-${flavor.toLowerCase()}-debug.apk');
```
bartekpacia, line 250: hacky hacky

but I don't know how to do it better

packages/patrol_cli/lib/src/features/drive/drive_command.dart
```diff
@@ -166,6 +175,22 @@ class DriveCommand extends StagedCommand<DriveCommandConfig> {
...
 
+    var repeat = 1;
+    try {
+      final repeatStr = argResults?['repeat'] as String? ?? '1';
+      repeat = int.parse(repeatStr);
+    } on FormatException {
+      throw const FormatException('`repeat` argument is not an int');
+    }
+
+    if (repeat != 1) {
```
shilangyu, line 186: Missing check for negative repeats

packages/patrol_cli/lib/src/features/drive/flutter_tool.dart
```diff
@@ -50,22 +72,84 @@ class FlutterDriver {
...
+    );
+
+    int? exitCode;
+    final process = await Process.start(
+      'flutter',
+      [
+        '--no-version-check',
+        ...['build', platform.command],
+        '--debug',
+        ...['--target', target],
```
shilangyu, line 98: I like this way of organizing

packages/patrol_cli/lib/src/features/drive/flutter_tool.dart
```diff
@@ -50,22 +72,84 @@ class FlutterDriver {
...
+        if (flavor != null) ...['--flavor', flavor!],
+        if (platform == TargetPlatform.iOS)
+          if (device.real) '--no-simulator' else '--simulator',
+        for (final dartDefine in {...env, ...dartDefines}.entries) ...[
+          '--dart-define',
+          '${dartDefine.key}=${dartDefine.value}',
+        ]
+      ],
+    );
+    String kill() {
```
shilangyu, line 108: newline

shilangyu (approved):

packages/patrol_cli/lib/src/features/drive/drive_command.dart
bartekpacia: Right, fixed.

Issue #534, opened 2022-10-28
Take screenshot on failure option


mateuszwojtczak:

bartekpacia:
See also:
- #8

bartekpacia:
**Status**
- taking screenshots on-demand implemented in #593

bartekpacia:
Shit, we've got a serious problem – after an hour of searching, I can't find any way to do reverse port forwarding on physical iOS devices.

I've assumed it's not gonna be a problem because we have the `iproxy` program, which can forward port 2000 on my mac to port 3000 on my iPhone by doing:

```console
$ iproxy 2000 3000
```

but apparently, `iproxy` can't forward port 3000 on my iPhone to port 2000 on my mac.

This makes it impossible to use any _host_ features (such as taking screenshots, executing scripts, etc.) from withing a test running on a physical iOS device.

bartekpacia:
todo: check if we can take screenshot only test failed

bartekpacia:
I couldn't find a way to execute code only if a particular test case failed.

bartekpacia:
We tried to refine this today, but realized we don't know what the requirements are.

thangmoxielabs:
Hi, is there a solution on this? I've learned that you need `flutter drive` to use the native screenshot feature, but how can you run that with `patrol test` or `patrol develop`?

vishna:
Unsure if there's easy way of solving this for Android/iOS but for web you can use playwright's `page.screenshot` method https://playwright.dev/docs/screenshots - made a PR with this: https://github.com/leancodepl/patrol/pull/3047 .

PR #562, opened 2022-11-03
Add a simple `takeScreenshot()` method

bartekpacia:
Part 1 of #534


bartekpacia:
Superseded by #593.

We'll use `flutter screenshot` instead of platform-specific screenshot taking APIs. It'll be simpler to implement and more robust.

PR #569, opened 2022-11-04
write smoke test for async alert dialog

bartekpacia:


bartekpacia:
Closing because of inability to reproduce.

PR #585, opened 2022-11-09
Add support for optional file with `--dart-define`s

### LINKED ISSUE #578: Create .patrol.env file from which `--dart-define`s will be read
This was requested by our testers guild. They don't like entering `--dart-define` (with e.g credentials) every time they run the test, it's bad DX.

We want to provide a `.patrol.env` file that will follow the simple env-format:

```
USERNAME=bartek.pacia@leancode.pl
PASSWORD=ny4ncat!=xyz
```

bartekpacia:
This PR:
- fixes #578

### Usage

The `.patrol.env` file should not be checked into git.

```console
$ pwd
/Users/bartek/dev/leancode/budowa/mobile
$ cat .patrol.env
EMAIL=awesome_tester@leancode.pl
PASSWORD=doge=kool
EXAMPLE_KEY='Al8=//\\'
$ patrol drive
Got --dart-define: EMAIL=awesome_tester@leancode.pl
Got --dart-define: PASSWORD=doge=kool
Got --dart-define: EXAMPLE_KEY='Al8=//\\'
<...>
```

### DIFF
--- a/packages/patrol_cli/lib/src/common/extensions/core.dart
+++ b/packages/patrol_cli/lib/src/common/extensions/core.dart
@@ -1,0 +1,18 @@
+extension MapX<K, V> on Map<K, V?> {
+  Map<K, V> withNullsRemoved() {
+    return {
+      for (final entry in entries)
+        if (entry.value != null) entry.key: entry.value as V,
+    };
+  }
+}
+
+extension StringX on String {
+  /// Adapted from https://stackoverflow.com/a/60404614/7009800
+  List<String> splitFirst(String sep) {
+    final index = indexOf(sep);
+    final parts = [substring(0, index).trim(), substring(index + 1).trim()];
+
+    return parts;
+  }
+}
--- a/docs/tips-and-tricks.mdx
+++ b/docs/tips-and-tricks.mdx
@@ -74,6 +74,16 @@ To set `USERNAME` and `PASSWORD`, use `--dart-define`:
 $ patrol drive --dart-define 'USERNAME=Bartek' --dart-define 'PASSWORD=ny4ncat'
 ```
 
+Alternatively you can create a `.patrol.env` file in your project's root. Here's
+an example:
+
+```
+$ cat .patrol.env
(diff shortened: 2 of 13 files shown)


bartekpacia:
Ignore failing tests, it's because we're using using an overridden version of `flutter_driver` by Fiołek. What matters is that it passes on `stable`.

Albert221:
```
Got --dart-define: EMAIL=awesome_tester@leancode.pl
Got --dart-define: PASSWORD=doge=kool
Got --dart-define: EXAMPLE_KEY='Al8=//\\'
```

This will be printed out, but I don't think we want this in CI? Various CIs replace secrets/envs with asterisks, but here they won't know it's secret.

Maybe simply announce that `EMAIL` was provided without disclosing its value? Or only when a flag is present?

bartekpacia:
@Albert221 Yeah, but I can imagine it might be useful for the user to see what values were picked up. @testers

An easy solution would be to announce that a `KEY` was passed, and if `--print-secrets` flag was passed, we'd also print keys. But maybe there's even simpler solution?

jBorkowska:
If there is any other option to view this file?

packages/patrol_cli/lib/src/common/extensions/core.dart
```diff
@@ -0,0 +1,18 @@
+extension MapX<K, V> on Map<K, V?> {
+  Map<K, V> withNullsRemoved() {
+    return {
+      for (final entry in entries)
+        if (entry.value != null) entry.key: entry.value as V,
```
shilangyu, line 5: This is valid only if `V extends Object`. Otherwise this is bugged (like the flutter std)

shilangyu: We do have a lint for that but it seems to not flag this: https://dart-lang.github.io/linter/lints/unrelated_type_equality_checks.html

Technically doing `entry.value != null` should raise an error, as it is not known whether `null` is an instance of `V`

Issue #584, opened 2022-11-09
Enable tests to perform actions on behalf of the host (using service extensions)


bartekpacia:
I want to embed an HTTP server in `patrol_cli`. Requests to that server would run shell commands (e.g `adb`, `flutter screenshot`) from within Dart test code running on a device.

This will enable us to solve the following:
- #534
- #73
- #66
- #556
- #606

I'd like to use gRPC for this (we already use it for `Flutter <--> native` communication within a device).

See also:
- [this comment](https://github.com/leancodepl/patrol/issues/66#issuecomment-1155687176) where the initial idea for this proposal was born

cc: @mateuszwojtczak, @jakubfijalkowski, @shilangyu

bartekpacia:
Could you elaborate?

jakubfijalkowski:
gRPC is unidirectional with a single exception - client streaming. You then don't need to open additional connection to send limited commands to the client.

I think this comment is somewhat irrelevant - I missed what parties you want to be connected. Sorry!

When it comes to CLI <-> Dart's part of the app comms, I personally would go with Dart VM's observatory service extension - there already is CLI <-> Dart communication (although somewhat hidden - it is maintained by the Flutter Driver but, if FD is used directly, you can attach to it) and then you can use events to send requests from tests to the CLI. This will be, IMO, much more Dart-native.

bartekpacia:
@jakubfijalkowski Whoa, this is exactly the big-brain kind of idea I'd expect from you. :D

This approach sounds very interesting, and I think I like it. 

It'll require more research from me than wrote I wrote in the first post (embedding a gRPC in the CLI), and I'm not sure what benefits we'll gain aside from being more Dart-native.

jakubfijalkowski:
The main big advantage here is that you will need to maintain only one connection. With separate gRPC server you will have to have two of them, which gets _funny_ when only one of them breaks in a silent manner ;P

bartekpacia:
Another use case for the HTTP server:
- https://github.com/leancodepl/patrol/issues/556

bartekpacia:
Patrol 1.0 works only inside the host app, so this is not longer relevant. In the future, we'll research how we can add host feautres back, but they will only work when the host is available (so, for example, not on Firebase Test Lab).

See also #661.

PR #593, opened 2022-11-09
Create a custom VM service extension and use it to implement the screenshot taking feature 

bartekpacia:
Idea was sparked by @jakubfijalkowski in [this comment](https://github.com/leancodepl/patrol/issues/584#issuecomment-1308597099).


bartekpacia:
Currently, I'm registering the service extension in the test and calling from the driver, but I want it the reverse.

Will poke around with this tomorrow.

bartekpacia:
Works. Wow.

Caveats:
- gotta find a way to find test. Should take a look at how Flutter does it
- `HostAutomator.takeScreenshot()` will crash when more than single device is attached

packages/patrol/lib/src/custom_finders/patrol_tester.dart
```diff
@@ -63,15 +65,23 @@ class PatrolTester {
...
+  final HostAutomator? hostAutomator;
+
   /// Shorthand for [nativeAutomator]. Throws if [nativeAutomator] is null.
   NativeAutomator get native {
     assert(nativeAutomator != null, 'native automator is null');
     return nativeAutomator!;
   }
 
+  HostAutomator get host {
+    assert(hostAutomator != null, 'host automator is null');
```
shilangyu, line 81: Usually assert messages say what was expected, explain the needed invariants (for instance `'host getter can be used only if hostAutomator was initialized before'` or something). Because currently the error message is not helpful.

shilangyu (approved):

packages/patrol/lib/src/custom_finders/patrol_tester.dart
bartekpacia: Thanks, I agree.

bartekpacia: In the end, I improved the doc comment above `native` and `host` fields instead of improving the assertion message. 

Albert told me some time ago that assertions should be in the form of:

`X is null`, `Y is not in the inclusive 1..10 range`

instead of:

`X must not be null`, `Y must be in the inclusive 1..10 range`

shilangyu: > Albert told me some time ago that assertions should be in the form of:
> 
> `X is null`, `Y is not in the inclusive 1..10 range`
> 
> instead of:
> 
> `X must not be null`, `Y must be in the inclusive 1..10 range`

I cannot say I agree... But even then, I don't think such error messages are helpful at all. They don't give any meaningful information. I know what the value should or should not be, I can see what you are asserting (dart's `assert` is not a function, it is a language feature, they capture the expression and print the exact expression that was used to perform the assert, see attached screenshot). Asserts are for _enforcing invariants_ so the error message should explain what kind of invariant was not fulfilled. Moving this to docs is great, though.

image

Coming back to this specific case, as an outside person who read the docs above this getter I still have no idea what should I do if I encounter a null here. Why would it be not initialized? Is it allowed to be not initialized? It is a `final` value, so I don't think I can initialize it myself.

PR #615, opened 2022-11-17
fix trying to run tests which failed to build

### LINKED ISSUE #610: Don't run the test if it failed to build
Running a test has 2 parts: building and running. Currently, when building fails, patrol_cli still proceeds with running, which is wrong behavior.

This should be fixe.d

bartekpacia:
Fixes #610

### DIFF
--- a/packages/patrol_cli/test/features/drive/test_runner_test.dart
+++ b/packages/patrol_cli/test/features/drive/test_runner_test.dart
@@ -31,12 +31,23 @@ void main() {
   });
 
   group('TestRunner', () {
-    group('addDevice()', () {
-      setUp(() {
-        testRunner.builder = (_, __) => delay();
-        testRunner.executor = (_, __) => delay();
-      });
+    setUp(() {
+      testRunner.builder = (_, __) => delay();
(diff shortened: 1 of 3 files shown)


packages/patrol_cli/test/features/drive/test_runner_test.dart
```diff
@@ -31,12 +31,22 @@ void main() {
...
   group('TestRunner', () {
-    group('addDevice()', () {
-      setUp(() {
-        testRunner.builder = (_, __) => delay();
-        testRunner.executor = (_, __) => delay();
+    setUp(() {
+      testRunner.builder = (_, __) => delay();
+      testRunner.executor = (_, __) => delay();
+    });
+    group('set repeats', () {
```
shilangyu, line 38: newline?

shilangyu (approved):

packages/patrol_cli/test/features/drive/test_runner_test.dart
bartekpacia: yup

thanks

Issue #561, opened 2022-11-03
`TestRunner` could be improved to report more results, better


bartekpacia:
Some code that would go in #560 but I decided it doesn't fit there.

```dart
// TODO: Use this
class SingleTargetRunResult {
  SingleTargetRunResult({
    required this.target,
    required this.device,
    required this.plannedRuns,
    this.buildPassed = false,
    required this.runs,
  });

  final String target;
  final Device device;
  bool buildPassed;

  final int plannedRuns;
  final List<RunStatus> runs;
}
```

And then `TestRunner.run()`'s signature would be:

```dart
Future<List<SingleTargetRunResult>> run() {}
```

Benefits:
- This would make code in `DriveCommand` cleaner (it'd be moved to `TestRunner`)
- All the complexity would be contained inside `TestRunner` and properly tested
- We could print a nice test result summary, similar to what `spec` does:

  ![Screenshot 2022-11-03 at 6 32 54 PM](https://user-images.githubusercontent.com/40357511/199793490-b73fbe81-faf6-48a9-9e5c-98ed5c921d32.png)

  Without colors it looks like:
  
  ```
   PASS  test/command_runner_test.dart
   PASS  test/features/drive/test_runner_test.dart
   PASS  test/features/drive/drive_command_test.dart
   PASS  test/features/drive/device_finder_test.dart
   PASS  test/features/drive/test_finder_test.dart
   PASS  test/common/artifacts_repository_test.dart
  ```

  I'd like to print something like:

    ```
   PASS  integration_test/core_cutter_test.dart.dart (build passed, 10/10 runs passed)
   FAIL  integration_test/dual_cbr_test.dart  (build failed)
   FAIL  integration_test/hsv_test.dart (build passed, 9/10 runs passed)
   CANC  integration_test/srd_test.dart  (build passed, 3/10 runs passed)
  ```

jBorkowska:
Is this a place to add the error messages for why a test failed? Or would it be kept separate?

bartekpacia:
No. To find you why a test failed, you'd have to still look in the logs.

jBorkowska:
From my perspective: I need a place, where I can easily find info about which test failed and why. Logs are quite hard to read (at least now), that's why I'm asking.

jakubfijalkowski:
[...]

PR #624, opened 2022-11-18
Add basic reporting to the test runner

### LINKED ISSUE #598: Brief test run results summary

bartekpacia:
Fixes #598

Also mostly fixes _ #561, but #561 is a bit broader, so I'm not closing #561

Screenshot 2022-11-23 at 1 46 30 PM
Screenshot 2022-11-23 at 1 58 52 PM
Screenshot 2022-11-23 at 2 17 42 PM

### DIFF
--- a/packages/patrol_cli/lib/src/features/drive/drive_command.dart
+++ b/packages/patrol_cli/lib/src/features/drive/drive_command.dart
@@ -178,7 +178,7 @@ class DriveCommand extends StagedCommand<DriveCommandConfig> {
       throw const FormatException('`wait` argument is not an int');
     }
 
-    int repeat;
+    final int repeat;
     try {
       final repeatStr = argResults?['repeat'] as String? ?? '$_defaultRepeats';
       repeat = int.parse(repeatStr);
--- a/packages/patrol_cli/lib/src/features/drive/drive_command.dart
+++ b/packages/patrol_cli/lib/src/features/drive/drive_command.dart
@@ -222,15 +222,12 @@ class DriveCommand extends StagedCommand<DriveCommandConfig> {
       dartDefines: config.dartDefines,
     );
 
-    var exitCode = 0;
-
     _testRunner
       ..repeats = config.repeat
       ..builder = (target, device) async {
         try {
           await _flutterTool.build(target, device);
         } catch (err) {
-          exitCode = 1;
           _logger
             ..err('$err')
             ..err(
@@ -242,8 +239,7 @@ class DriveCommand extends StagedCommand<DriveCommandConfig> {
       ..executor = (target, device) async {
         try {
           await _flutterTool.drive(target, device);
-        } on FlutterDriverFailedException catch (err) {
-          exitCode = 1;
+        } catch (err) {
           _logger
             ..err('$err')
             ..err(
@@ -276,8 +272,10 @@ class DriveCommand extends StagedCommand<DriveCommandConfig> {
       }
     }
 
-    await _testRunner.run();
+    final results = await _testRunner.run();
+    // TODO: present test results
 
+    final exitCode = results.allSuccessful ? 0 : 1;
     return exitCode;
   }
 }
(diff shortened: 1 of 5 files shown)


packages/patrol_cli/lib/src/features/drive/drive_command.dart
```diff
@@ -175,9 +178,9 @@ class DriveCommand extends StagedCommand<DriveCommandConfig> {
       throw const FormatException('`wait` argument is not an int');
     }
 
-    var repeat = 1;
+    int repeat;
```
PiotrRogulski, line 181: It can be final now

packages/patrol_cli/test/features/drive/test_runner_test.dart
```diff
@@ -230,7 +230,10 @@ void main() {
 
           final expectedLog = <String>[];
           expect(actualLog, equals(<String>[]));
-          unawaited(testRunner.run());
+          late TestRunnerResult result;
+          unawaited(() async {
+            result = await testRunner.run();
+          }());
```
PiotrRogulski, line 236: The other tests below use regular `await`s to get the results – much better since you won't run into the risk of the log output not catching up with the intermediate delays

packages/patrol_cli/lib/src/features/drive/test_runner.dart
```diff
@@ -101,30 +158,45 @@ class TestRunner extends Disposable {
...
+    final targetRunResults = <TargetRunResult>[];
+
     final testRunsOnAllDevices = <Future<void>>[];
     for (final device in _devices.values) {
       Future<void> runTestsOnDevice() async {
         for (final target in _targets) {
+          final targetRuns = <TargetRunStatus>[];
+          targetRunResults.add(
+            TargetRunResult(target: target, device: device, runs: targetRuns),
+          );
```
PiotrRogulski, line 170: I know it's probably not that much of an issue here, but having `TargetRunResult` as an `Equatable` and then mutating its fields seems like a risky maneuver

packages/patrol_cli/lib/src/features/drive/drive_command.dart
bartekpacia: Thanks

Xezolpl (approved):

packages/patrol_cli/test/features/drive/test_runner_test.dart
bartekpacia: I'm doing it this way because of https://github.com/dart-lang/fake_async/issues/51.

BTW the delays are entirely controlled by the fake async, so if the log output "won't catch up", then it's a bug and it's good that tests will fail in that case.

packages/patrol_cli/lib/src/features/drive/test_runner.dart
bartekpacia: Why? I don't think it breaks `Equatable`'s contract.

packages/patrol_cli/lib/src/features/drive/test_runner.dart
```diff
@@ -1,8 +1,65 @@
...
 
 typedef _Callback = Future<void> Function(String target, Device device);
 
+class TestRunnerResult with EquatableMixin {
+  const TestRunnerResult({required this.targetRunResults});
+
+  final List<TargetRunResult> targetRunResults;
+
+  bool get allSuccessful => targetRunResults
+      .every((e) => e.runs.every((run) => run == TargetRunStatus.passed));
```
shilangyu, line 14: Use allRunPassed from this run class

packages/patrol_cli/lib/src/features/drive/test_runner.dart
shilangyu: I think it's ok. Besides the fact that i do not like mutability at all, but whether it is Equatable or not doesn't change anything here.

packages/patrol_cli/test/features/drive/test_runner_test.dart
shilangyu: Something like expectLater won't work the way you want?

bartekpacia: If you have an idea how to make this work, I'll be happy to hear it.

Issue #65, opened 2022-06-14
Add support for using existing apk


bartekpacia:
This will free us from having to do a Gradle build on each `maestro drive`.

`flutter drive` has `--use-application-binary` option, which takes path to APK as an argument.

PR #667, opened 2022-12-07
Add support for using --use-application-binary in drive command.

nonvachara-paloit:
### Add support for using --use-application-binary in drive command.

As `flutter drive` already supports this, so this commit added a new command line argument `--use-application-binary` in `patrol drive` and pass that to `flutter drive` command.

Directly relates to #65 .

### DIFF
--- a/packages/patrol_cli/lib/src/features/drive/drive_command.dart
+++ b/packages/patrol_cli/lib/src/features/drive/drive_command.dart
@@ -124,9 +124,10 @@ class DriveCommand extends StagedCommand<DriveCommandConfig> {
         'use-application-binary',
         help:
             'Specify a pre-built application binary to use when running. For Android applications, '
-            'this must be the path to an APK. For iOS applications, the path to an IPA. Other device types '
-            'do not yet support prebuilt application binaries.',
-        valueHelp: 'path/to/app.apk',
+            'this must be the path to an APK. For iOS applications, the path to a .app or an IPA. '
+            'Other device types do not yet support prebuilt application binaries. '
+            'See `flutter drive --help` on for more information.',
+        valueHelp: 'path/to/app.apk|path/to/iphonesimulator/app.app',
       );
   }
 
--- a/packages/patrol_cli/lib/src/features/drive/drive_command.dart
+++ b/packages/patrol_cli/lib/src/features/drive/drive_command.dart
@@ -30,6 +30,7 @@ class DriveCommandConfig with _$DriveCommandConfig {
     required String? packageName,
     required String? bundleId,
     required int repeat,
+    required String? useApplicationBinary,
   }) = _DriveCommandConfig;
 }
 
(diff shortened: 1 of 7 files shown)


packages/patrol_cli/lib/src/features/drive/drive_command.dart
```diff
@@ -118,6 +119,14 @@ class DriveCommand extends StagedCommand<DriveCommandConfig> {
         abbr: 'n',
         help: 'Repeat the test n times.',
         defaultsTo: '$_defaultRepeats',
+      )
+      ..addOption(
+        'use-application-binary',
+        help:
+            'Specify a pre-built application binary to use when running. For Android applications, '
+            'this must be the path to an APK. For iOS applications, the path to an IPA. Other device types '
```
bartekpacia, line 127: Does APP (for iOS) also work?

packages/patrol_cli/lib/src/features/drive/test_runner.dart
```diff
@@ -83,6 +83,15 @@ class TestRunner extends Disposable {
     _repeats = newValue;
   }
 
+  String? _useApplicationBinary;
+  set useApplicationBinary(String? newValue) {
+    if (_running) {
+      throw StateError('repeats cannot be changed while running');
```
bartekpacia, line 89:
```suggestion
      throw StateError('application binary cannot be changed while running');
```

packages/patrol_cli/lib/src/features/drive/test_runner.dart
```diff
@@ -175,11 +184,13 @@ class TestRunner extends Disposable {
...
-          } catch (_) {
-            targetRuns.add(TargetRunStatus.failedToBuild);
-            continue;
+          if (_useApplicationBinary == null) {
+            try {
+              await builder(target, device);
+            } catch (_) {
+              targetRuns.add(TargetRunStatus.failedToBuild);
+              continue;
+            }
```
bartekpacia, line 193: make an early return ("_useApplicationBinary != null; then return")

packages/patrol_cli/lib/src/features/drive/drive_command.dart
nonvachara-paloit: I tried with `--use-application-binary=build/ios/Debug-iphonesimulator/Runner.app`, also work.

packages/patrol_cli/lib/src/features/drive/test_runner.dart
nonvachara-paloit: I'm not sure we're on the same page here. 🤔
We checked whether _useApplicationBinary (from parseInput() then config object).
If it was specified, then we skip the building part only.
The rest should continue to execute normally.

packages/patrol_cli/lib/src/features/drive/drive_command.dart
bartekpacia: OK 👍 

Please update the description to state this :)

nonvachara-paloit: A new commit was pushed to state iOS path to a .app in help text.

bartekpacia:
Thanks for your contribution! 💙

PR #709, opened 2022-12-23
Remove Automator code from Patrol plugin

bartekpacia:
Can't verify if works because [Jitpack is down](https://github.com/jitpack/jitpack.io/issues/5337).


bartekpacia:
We decided to stay with the Flutter plugin approach and introduce a less invasive solution.

PR #839, opened 2023-01-28
Run all tests on iOS simulator

bartekpacia:


.github/workflows/integration-test-ios.yaml
```diff
@@ -46,17 +46,34 @@ jobs:
...
+        if: ${{ false }} # Not on Simulator
+
+      - run: patrol test -t integration_test/service_dark_mode_test.dart
+        if: success() || failure()
+
+      - run: patrol test -t integration_test/service_wifi_test.dart
+        if: ${{ false }} # Not on Simulator
+
+      - run: patrol test -t integration_test/swipe_test.dart
+        if: success() || failure()
```
kopax-polyconseil, line 76: always()

bartekpacia: Why?

PR #849, opened 2023-01-29
Android: add support for `Selector.index` 

### LINKED ISSUE #663: Webview interaction - selecting correct element when Selector applies to multiple elements
Let's assume the following login form:
Zrzut ekranu 2022-12-5 o 13 02 44

My app opens it in a web view, and I want to put some login and password and tap the "Log in" button. I use the following code:

```dart
await $.native.enterTextByIndex('some_login', index: 0);
await $.native.enterTextByIndex('some_password', index: 1);
await $.native.tap(Selector(text: 'Log in'));
```

Entering the text works like a charm, but the `tap()` command selects the header above the fields instead of the button. I tried with `className: 'button'`, `'className: 'Button'` (inspired by the [article on Medium](https://medium.com/azimolabs/android-webview-automation-testing-6311df8fe42e)), but it couldn't find such element. I also tried playing with the `instance` field, but I only got `IllegalArgumentException` with a message saying, that `instance` isn't supported.

Is there a way to select the button in such case?

bartekpacia:
Fixes #663


bartekpacia:
To do:
- finish iOS implementation to match Android
- make sure everything works

Issue #883, opened 2023-02-05
Gradle-generated reports are overwritten when running more than single test


bartekpacia:
**Describe the bug**

**To Reproduce**
1. Run `patrol test` passing two test `--target`s (make both fail to make report path visible)
2. Wait until both tests they fail, and open the generated report

   ```
	• Executing tests of apk with entrypoint notifications_test.dart on emulator-5554...

		FAILURE: Build failed with an exception.

		* What went wrong:
		Execution failed for task ':app:connectedDebugAndroidTest'.
		> There were failing tests. See the report at: file:///Users/bartek/dev/leancode/patrol/packages/patrol/example/android/test-reports/connected/index.html

		* Try:
		> Run with --stacktrace option to get the stack trace.
		> Run with --info or --debug option to get more log output.
		> Run with --scan to get full insights.

		* Get more help at https://help.gradle.org

		BUILD FAILED in 20s
	✗ Failed to execute tests of apk with entrypoint notifications_test.dart on emulator-5554 (Gradle test execution failed with code 1) (21.2s)
   ```

   In this case, open `<mobile root>/android/test-reports/connected/index.html`.

3. Only the latest report is present, and all the previous ones are overwritten by it.

**Expected behavior**

At the very least, Gradle-generated reports for every test target shouldn't be overwritten. We could copy every test target to its own directory in (for example) in `<mobile root>/android/test-reports/connected`.

A better solution would be to merge all Gradle-generated reports into a single final report.

<details>
<summary>Additional information</summary>

```
Flutter 3.7.0 • channel stable • https://github.com/flutter/flutter.git
Framework • revision b06b8b2710 (13 days ago) • 2023-01-23 16:55:55 -0800
Engine • revision b24591ed32
Tools • Dart 2.19.0 • DevTools 2.20.1

```

```yaml
patrol: 0.10.9
```

```console
patrol_cli v0.9.0
```

</details>

cc: @mateuszwojtczak

mateuszwojtczak:
It looks like it can be specified in the Gradle: https://stackoverflow.com/questions/19326527/specify-gradle-report-directory

Maybe we could somehow reset this after every test run, so that we have multiple files - each per test run

bartekpacia:
Sounds good.

`patrol test` already does this internally:

```console
./gradlew :app:connectedDebugAndroidTest \
   -Ptarget="$(pwd)/../integration_test/example_test.dart" \
```

[...]

PR #899, opened 2023-02-07
Fix Gradle test results being overwritten in the example app

bartekpacia:
Addresses #https://github.com/leancodepl/patrol/issues/883

I'm planning to include this in the documentation soon.

### DIFF
--- a/packages/patrol/example/android/app/build.gradle
+++ b/packages/patrol/example/android/app/build.gradle
@@ -55,8 +55,15 @@ android {
     }
 
     testOptions {
-        reportDir "$rootDir/test-reports"
-        resultsDir = "$rootDir/test-results"
+        def targetPath = findProperty("target")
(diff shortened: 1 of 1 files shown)


packages/patrol/example/android/app/build.gradle
```diff
@@ -55,8 +55,15 @@ android {
     }
 
     testOptions {
-        reportDir "$rootDir/test-reports"
-        resultsDir = "$rootDir/test-results"
+        def targetPath = findProperty("target")
+        if (targetPath != null) {
```
Albert221, line 59: What about the `else`? Is the default sufficient?

bartekpacia: We're okay with `target` being `null` - in such case, the default (`lib/main.dart`) is used.

PR #908, opened 2023-02-08
Use a single instance of `LoggingLocalProcessManager`

bartekpacia:


bartekpacia:
No idea why it's saying that there are warnings. Merging nevertheless.

In 1.0, all the deprecated members will be removed, maybe that'll help.

PR #955, opened 2023-02-17
Fix Ci failing to create GitHub Release

bartekpacia:


packages/patrol_cli/lib/src/common/constants.dart
```diff
@@ -1,2 +1,2 @@
 /// Version of Patrol CLI. Must be kept in sync with pubspec.yaml.
```
Albert221, line 1: I'd add the same comment in pubspec.yaml. I'd say that pubspec.yaml is _the_ default place to look for version, not this file

Issue #66, opened 2022-06-14
Close keyboard if opened


bartekpacia:
There's no easy way to do this using UiAutomator.

There's a hack though. An ugly one, but ~it works~ it works only through ADB.

 Here's [the link](http://androidsdet.blogspot.com/2016/05/how-to-close-android-keyboard-in.html)

bartekpacia:
We could differentiate between _host services_ and _local services_ ([ADB does it](https://android.googlesource.com/platform/packages/modules/adb/+/refs/heads/master/OVERVIEW.TXT#58)).

bartekpacia:
### Important update

In #593 we've got a powerful new capability of acting as a host. This means we're able to use `adb` and programmatically hide keyboard.

[Here's how Maestro does this](https://github.com/mobile-dev-inc/maestro/blob/main/maestro-client/src/main/java/maestro/drivers/AndroidDriver.kt) – the `hideKeyboard()` method.

bartekpacia:
We lost [this feature](https://github.com/leancodepl/patrol/issues/66#issuecomment-1315670061) in :)

yacineblr:
Do you have any news about this please ?

andriiShutov:
Will it implemented?

PR #967, opened 2023-02-18
Disable showing software keyboard in enterText/enterTextByIndex()

bartekpacia:
> This only affect `$.native.enterText()` and `$.native.enterTextByIndex()` methods.

The keyboard sometimes obscures a view. We don't have any `scrollTo()` equivalent in the `$.native` world, and it's impossible/very hard to reliably hide software keyboard in an Android UI test)

Removing showing software keyboard when entering text into text fields:

- Makes certain scenarios possible.

- We stray further from the real world. A real user always sees the software keyboard when entering some text.

I'm leaning toward merging this PR. We can come back to it once we have a reliable way

After all, Patrol's native features exist to unblock tests in native scenarios, and the software keyboard sometimes makes this goal impossible to achieve.

See also:
- #155
- #66

cc: @jBorkowska, @mateuszwojtczak

### DIFF
--- a/packages/patrol/android/src/main/kotlin/pl/leancode/patrol/Automator.kt
+++ b/packages/patrol/android/src/main/kotlin/pl/leancode/patrol/Automator.kt
@@ -189,10 +189,7 @@ class Automator private constructor() {
 
         val uiSelector = UiSelector().className(EditText::class.java).instance(index)
         val uiObject = uiDevice.findObject(uiSelector)
-        uiObject.click()
         uiObject.text = text
-
-        pressBack() // Hide keyboard.
     }
 
     fun enterText(text: String, uiSelector: UiSelector, bySelector: BySelector, index: Int) {
@@ -203,10 +200,7 @@ class Automator private constructor() {
         }
 
         val uiObject = uiDevice.findObject(uiSelector).getFromParent(UiSelector().className(EditText::class.java))
-        uiObject.click()
         uiObject.text = text
-
-        pressBack() // Hide keyboard.
     }
 
     fun swipe(startX: Float, startY: Float, endX: Float, endY: Float, steps: Int) {


bartekpacia:
Here's a test that is made possible by this change:

https://user-images.githubusercontent.com/40357511/222420296-daec03bb-fdb2-4351-bdbf-01267ac22c29.mp4

Albert221:
I think the keyboard should always appear, as it does for a real user. So I'm against this PR.

What we can do though - what I thought of - is, given we know the selector for the element we want to input text into, we may check where it is on the screen. if it's somewhere in the top half (above the part of the screen obstructed by the keyboard) then we do nothing. if that's in the lower half (below the keyboard), we move the screen by keyboard height or the distance from top of the keyboard to bottom of the selected element, or something like that.

> Makes certain scenarios possible.

What are those scenarios? I think this PR lacks a specific rationale in the description

jBorkowska:
We should be able to scroll in such situations. As user would :). From what I saw, I think that there is automatic scroll after showing the keyboard, so the field that has focus on is shown. But sometimes that doesn't work.

Albert221:
> there is automatic scroll after showing the keyboard
This is Flutter's default behavior, doesn't work outside the app

bartekpacia:
> What are those scenarios? I think this PR lacks a specific rationale in the description

[Here's one](https://github.com/leancodepl/patrol/pull/967#issuecomment-1451741402).

If the keyboard was visible, and we were entering `Email`, the Selector to tap on it would tap on the keyboard instead, and the test would fail.

bartekpacia:
I'd like a more clever solution than "disable keyboard" as well. Not showing keyboard is just very easy to do 😅

Albert221:
> https://github.com/leancodepl/patrol/pull/967#issuecomment-1451741402.

Please show me how it doesn't work without hidden keyboard

ehabibov:
> I'd like a more clever solution than "disable keyboard" as well. Not showing keyboard is just very easy to do 😅

@bartekpacia what about tapping somewhere else to dismiss keyboard? at (0,0) for ex.? Like appium, which has some strategies [to dismiss keyboard](https://appium.io/docs/en/commands/device/keys/hide-keyboard/) (but I can't say if that solution even could be implemented as have no clue how patrol works with native).
On iOS it could tap Done button, which is always there (as far as i remember).
And pass additional optional parameter to enterText() if there is a need to use that solution or not if it becomes stable.

Agree that permanent keyboard disabling is unacceptable but could be used as a measure of last hope and set somewhere in configs or whatever.

walid-ashik:
This PR seems interesting. Yesterday I was struggling with this actually `if there's any way not to show the keyboard while entering text` But for a quick workaround, after entering text when I want my test code to tap on some other buttons or texts, I hide the keyboard manually. This could be nice to have a feature that can be controlled with some flags on the test.

bartekpacia:
[see this](https://github.com/leancodepl/patrol/issues/1453#issuecomment-1617962213) - looks like it's possible to hide the keyboard easily during tests (I haven't checked it though)

bartekpacia:
We decided to go with a temporary solution: let the developer specify whether to show the keyboard during `$.native.enterText()`.

Example:

```dart
await $.native.enterText(
  Selector(text: 'Enter your email'),
  text: 'charlie@root.me',
  keyboardBehavior: KeyboardBehavior.ignore,
);
```

This will:
-  not show the keyboard at all on Android
-  hide the keyboard immediately after entering text by appending `\n` to the text written into the textfield

Albert221:
I don't like the name. KeyboardBehavior.ignore doesn't tell me anything straight away, whether it ignores showing the keyboard (doesn't show it) or ignores something else?

bartekpacia:
Superseded by #1511.

PR #1000, opened 2023-02-27
patrol: drop podspec platform version to 11

### LINKED ISSUE #999: Extract finders to a Flutter-only `patrol_finders` package
**Describe the bug**

I've started using Patrol in my widget tests because of the nice syntax with finders etc. It's great!

But when I run my app on iOS, I'm getting issues such as:

`Specs satisfying the `patrol (from `.symlinks/plugins/patrol/ios`)` dependency were found, but they required a higher minimum deployment target.`

I don't want to bump my deployment target, and I don't actually need the Patrol pod as I'm not using it for instrumentation tests.

Is there a supported way to only use Patrol in widget tests and not compile the Patrol plugins in the app?

josh-burton:
Fixes #999.

There doesn't seem to be a specific reason that 13 is required, and 11 allows a wider range of apps to use Patrol.

### DIFF
--- a/packages/patrol/ios/patrol.podspec
+++ b/packages/patrol/ios/patrol.podspec
@@ -18,12 +18,12 @@ Runs tests that use flutter_test and patrol APIs as native iOS integration tests
   s.dependency 'Flutter'
   s.weak_framework = 'XCTest'
   s.ios.framework  = 'UIKit'
-  s.platform = :ios, '13.0'
+  s.platform = :ios, '11.0'
 
   # Flutter.framework does not contain a i386 slice.
   s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES', 'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'i386' }
   s.swift_version = '5.0'
-  
+
   s.dependency 'gRPC-Swift', '~> 1.8.0' # This is the last version published on CocoaPods.
                                         # Newer ones are only available on SPM.
 end


bartekpacia:
Congrats on #1000 🎉! :)

> There doesn't seem to be a specific reason that 13 is required,

Unfortunately, that's not true. Patrol uses Swifts `async/await` feature, which is only available on iOS 13 and newer. [Learn more here](https://www.hackingwithswift.com/quick-start/concurrency/where-is-swift-concurrency-supported).

From #999:
> I guess ideally the patrol package would be split into patrol and patrol-integration, with the latter containing the native android/ios integration pieces.

This sounds like a good idea. To make this change seamless, we'll probably extract custom finders to `patrol_finders` package, and make `patrol` depend on `patrol_finder`. So, developers like you, who only use finders, will depend on `patrol_finders` package :)

josh-burton:
> Congrats on #1000 🎉! :)
> 
> > There doesn't seem to be a specific reason that 13 is required,
> 
> Unfortunately, that's not true. Patrol uses Swift's `async/await` feature, which is only available on iOS 13 and newer. [Learn more here](https://www.hackingwithswift.com/quick-start/concurrency/where-is-swift-concurrency-supported).
> 
> From #999:
> 
> > I guess ideally the patrol package would be split into patrol and patrol-integration, with the latter containing the native android/ios integration pieces.
> 
> This sounds like a good idea. To make this change seamless, we'll probably extract custom finders to `patrol_finders` package, and make `patrol` depend on `patrol_finder`. So, developers like you, who only use finders, will depend on `patrol_finders` package :)

Ok great, that sounds like a good solution. I'll close this PR.

PR #1062, opened 2023-03-09
SwiftPatrolPlugin: improve log messages

### LINKED ISSUE #969: Fully migrate native reporting to use gRPC
Scope:
- stop using MethodChannels for test reporting on Android. We use gRPC on iOS (it's impossible to use the method channel there). It'll simplify things if we make things work similarly.

Also, let's make sure to pass dart-define `INTEGRATION_TEST_SHOULD_REPORT_RESULTS_TO_NATIVE=false` to every build, because otherwise the `integration_test` plugin will try to submit the native test results over the method channel, which will fail and might cause unexpected behavior when coupled with our own native test reporting.

ios arch

android arch

bartekpacia:
related to #969


bartekpacia:
We've put this for later.

Issue #1048, opened 2023-03-06
`tap()` sometimes taps on invisible widgets


jBorkowska:
Example: https://github.com/jBorkowska/flutter-wonderous-app/blob/patrol-bug/integration_test/test_test.dart, but remove `scrollTo()` in line 25.
Then `tap()` is called on finder, which matches 2 widgets, but only one of them is hittestable. At the same time, the one that is not hittestable is higher in widget tree that the hittestable one (more in for in this [issue](https://github.com/leancodepl/patrol/issues/1046)). But real tap is called on `finder.first`, not `finder.hitTestable().first` (https://github.com/leancodepl/patrol/blob/master/packages/patrol/lib/src/custom_finders/patrol_tester.dart#L230).

bartekpacia:
Thank you very much for this report. This was a bug in `PatrolTester.waitUntilVisible()` – it should be fixed by #1072.

bartekpacia:
Fixed in #1072.

PR #1072, opened 2023-03-09
Fix `tap()` not being able to tap on widget previously found by `scrollTo()`

### LINKED ISSUE #1046: Can't tap on widget previously found by `scrollTo()`
Code can be found here: https://github.com/jBorkowska/flutter-wonderous-app/blob/patrol-bug/integration_test/test_test.dart#L25
In this line, we scroll to a widget and then try to tap on it. Scrolling works fine, but tap throws an error:

Image

Possible reason: there are 2 widgets with the same key in the widget tree at this moment in test: a container in https://github.com/jBorkowska/flutter-wonderous-app/blob/patrol-bug/lib/ui/screens/home/wonders_home_screen.dart#L293 and a stack in https://github.com/jBorkowska/flutter-wonderous-app/blob/patrol-bug/lib/ui/wonder_illustrations/common/animated_clouds.dart#L100. The container is found by scrollTo, but in error message from tap only the stack is mentioned (as offscreen). Also stack is higher in the widget tree than container, which means it's analyzed before container when resolving the finder. That potentially means, that scrollTo analyzed stack, but ignored it as offstage, but tap took first one from the tree and then checked hitTestable(), which returned false, resulting in exception from screenshot above.

Fix (to be reviewed by @bartekpacia ):
tldr: `scrollTo()` has to pass original finder or `finder.hitTestable().first`.
`scrollTo()` returns `finder.first` instead of original finder. So in this case it returned the stack widget, which is not hittestable. This was passed to `tap()` method and that is why the error message printed only this stack widget (`tap()` called `hitTestable()` on finder passed from `scrollTo()`, that's why it threw an error).

Another issue found in the process: https://github.com/leancodepl/patrol/issues/1048.

bartekpacia:
Fixes #1046

Huge thank you to @jBorkowska for such an in-depth investigation of the issue!

### DIFF
--- a/packages/patrol/lib/src/custom_finders/patrol_tester.dart
+++ b/packages/patrol/lib/src/custom_finders/patrol_tester.dart
@@ -324,8 +324,8 @@ class PatrolTester {
     return TestAsyncUtils.guard(() async {
       final duration = timeout ?? config.visibleTimeout;
       final end = tester.binding.clock.now().add(duration);
-
-      while (finder.hitTestable().evaluate().isEmpty) {
+      final hitTestableFinder = finder.hitTestable();
+      while (hitTestableFinder.evaluate().isEmpty) {
         final now = tester.binding.clock.now();
         if (now.isAfter(end)) {
           throw WaitUntilVisibleTimeoutException(
@@ -337,7 +337,7 @@ class PatrolTester {
         await tester.pump(const Duration(milliseconds: 100));
       }
 
-      return PatrolFinder(finder: finder, tester: this);
+      return PatrolFinder(finder: hitTestableFinder, tester: this);
     });
   }
 
@@ -390,7 +390,7 @@ class PatrolTester {
         settleTimeout: config.settleTimeout,
       );
 
-      return PatrolFinder(finder: finder.first, tester: this);
+      return PatrolFinder(finder: finder, tester: this);
     });
   }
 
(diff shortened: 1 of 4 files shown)


bartekpacia:
This should also fix #1048.

jBorkowska (approved):

packages/patrol/lib/src/custom_finders/patrol_tester.dart
```diff
@@ -439,7 +439,7 @@ class PatrolTester {
         settleTimeout: config.settleTimeout,
       );
 
-      return PatrolFinder(finder: finder.first, tester: this);
+      return PatrolFinder(finder: finder, tester: this);
```
bartekpacia, line 442: research if `finder.hitTestable()` breaks any existing tests

bartekpacia: do the same for liine 435

bartekpacia:
Changing line 442 from:

```dart
return PatrolFinder(finder: finder, tester: this);
```

to:

```dart
return PatrolFinder(finder: finder.hitTestable(), tester: this);
```

doesn't break any tests. This is bad, because it means there's some behavior in our custom finders that is not covered by unit tests. I'll fix this.

bartekpacia:
>  I'll fix this (add unit tests) in this PR.

Done.

jBorkowska (approved):

PR #1073, opened 2023-03-09
Fix "error while disposing"  when waiting for app to start on physical iOS device

bartekpacia:
Fixes bug mentioned by @jaumard in https://github.com/leancodepl/patrol/issues/1051#issuecomment-1462162470

### DIFF
--- a/packages/patrol_cli/lib/src/ios/ios_deploy.dart
+++ b/packages/patrol_cli/lib/src/ios/ios_deploy.dart
@@ -64,6 +64,7 @@ class IOSDeploy {
         _logger.detail(
           'Stopped waiting for ios-deploy becuase dispose scope was closed',
         );
+        break;
       }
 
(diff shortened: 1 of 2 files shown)


packages/patrol_cli/lib/src/ios/ios_deploy.dart
```diff
@@ -60,14 +60,20 @@ class IOSDeploy {
 
     var totalTimeElapsed = Duration.zero;
     while (!launchSucceeded) {
+      if (_disposeScope.disposed) {
+        _logger.detail(
```
jaumard, line 64: Shouldn't you break here to leave the infinite loop? Or it will not crash but still run.

bartekpacia: I definitely should! Thanks!

PR #1076, opened 2023-03-10
Fix xctestrun filename when scheme is not `Runner`

vincent-hoodoo:
Hello,

Thanks for the great package!

This PR adds support for running tests on iOS with schemes other than `Runner`


JustinmClapperton:
I was just about to write this PR myself.  How can we get this merged!  nice work

JustinmClapperton (approved):
I have tested this as well and it resolves the issue

bartekpacia:
Thank you for the PR @vincent-hoodoo and thank you @JustinmClapperton for checking it out :) Will release this soon.

PR #1095, opened 2023-03-16
Enable Dart beta and Flutter beta in CI

bartekpacia:


bartekpacia:
`dart analyze` in Ci fails because of https://github.com/leancodepl/flutter_corelibrary/issues/96.

PR #1097, opened 2023-03-16
Add `patrol build ios`, move existing `patrol build` into `patrol build android`

### LINKED ISSUE #898: Introduce `patrol build` to make it easier to test on CI
### Problem

`patrol test` does a lot. What we're interested in this issue is that `patrol test` reads configuration from `patrol` block in `pubspec.yaml` and `--dart-define`s from `.patrol.env` file.

Meanwhile in [Firebase Test Lab docs](https://patrol.leancode.co/ci), we recommend our users build apps for it with manual `gradle` or `xcodebuild` invocations. The problem is that the configuration from  `pubspec.yaml` or `.patrol.env` is not build into the app in such case, and usually results in surprising failures on CI.

### Proposed solution

Introduce `patrol build` command, which would do basically what `patrol test` does currently, but would stop after the apps are built.

It would also print paths where apps are built so users don't waste time looking for them.

### Scope

- [x] https://github.com/leancodepl/patrol/issues/1100
- [x] https://github.com/leancodepl/patrol/issues/1101
- [ ] `patrol build ios`: automatically create a XCTest .zip

XCTest .zip is required by Firebase Test Lab. It's quite easy to generate it manually once we have app binaries (which `patrol build ios` does), but if we can improve our users' experience more, let's do it.

https://user-images.githubusercontent.com/40357511/225779051-5602f057-9dba-4e86-813f-3b1414ace7eb.png

See also:
- #853 

cc: @mateuszwojtczak

### LINKED ISSUE #1100: Add `patrol build android` and `patrol build ios` commands
Part of #898

bartekpacia:
Fixes #1100


packages/patrol_cli/lib/src/commands/build_android.dart
```diff
@@ -0,0 +1,116 @@
...
+    final packageName = stringArg('package-name') ?? config.android.packageName;
+
+    final displayLabel = boolArg('label');
+
+    final customDartDefines = {
+      ..._dartDefinesReader.fromFile(),
+      ..._dartDefinesReader.fromCli(args: stringsArg('dart-define')),
+    };
+    final internalDartDefines = {
+      'PATROL_WAIT': defaultWait.toString(),
```
Albert221, line 73: ~Where does the `defaultWait` come from?~

I see it's from `PatrolCommand`. But what's the wait? Wait for/before what? And in what unit is it specified? Please add dartdoc for that field

packages/patrol_cli/lib/src/commands/build_ios.dart
```diff
@@ -0,0 +1,142 @@
...
+    required DartDefinesReader dartDefinesReader,
+    required PubspecReader pubspecReader,
+    required IOSTestBackend iosTestBackend,
+    required Logger logger,
+  })  : _testFinder = testFinder,
+        _dartDefinesReader = dartDefinesReader,
+        _pubspecReader = pubspecReader,
+        _iosTestBackend = iosTestBackend,
+        _logger = logger {
+    usesTargetOption();
```
Albert221, line 24: I like this pattern. But I see it comes from `PatrolCommand`. IMO this should be separate from the parent-class of all commands, in something like `FlutterBuildingPatrolCommandMixin`...? This name may suck, but you get the idea. `PatrolCommand` could of course preserve the arg parsing capabilities, but leave the `usesXOption` to more usecase group-oriented mixins

packages/patrol_cli/lib/src/commands/build_ios.dart
```diff
@@ -0,0 +1,142 @@
...
+    final target = _testFinder.findTest(targetArg.single);
+    _logger.detail('Received test target: $target');
+
+    final config = _pubspecReader.read();
+    final flavor = stringArg('flavor') ?? config.ios.flavor;
+    if (flavor != null) {
+      _logger.detail('Received iOS flavor: $flavor');
+    }
+
+    final bundleId = stringArg('bundle-id') ?? config.ios.bundleId;
```
Albert221, line 68: From that point on, to after `flutterOpts` definition there is a common code for Android and iOS (and possible future platforms as I suppose). Can this be unified and not repeated?

I'm not talking about plain simple code repetition, but more about the context, the context of this code (looks to me) is the same.

packages/patrol_cli/lib/src/runner/patrol_command.dart
```diff
@@ -43,7 +43,6 @@ abstract class PatrolCommand extends Command<int> {
       ..addFlag(
         'debug',
         help: 'Build a debug version of your app (default mode)',
-        defaultsTo: true,
```
Albert221, line 46: Won't that remove info about this from help? How are you telling the user that debug is a default?

packages/patrol_cli/lib/src/runner/patrol_command_runner.dart
```diff
@@ -73,19 +73,35 @@ class PatrolCommandRunner extends CompletionCommandRunner<int> {
           'patrol',
           'Tool for running Flutter-native UI tests with superpowers',
         ) {
```
Albert221, line 75: I see that all the commands depend on `testFinder`, `dartDefinesReader`, `pubspecReader`, `androidTestBackend`, `iosTestBackend`, and `logger`. I'm thinking whether there's some smart way to bundle it all together and not pass it explicitly to each and every command, but without simply wrapping it in a class, but somehow that would make sense

packages/patrol_cli/lib/src/commands/build_ios.dart
bartekpacia: TBH, I took inspiration from `flutter_tools`. They also have `FlutterCommand` and `usesXOption` methods are defined in there as well.

I agree with you, but I think it's... kind of an overkill?

packages/patrol_cli/lib/src/runner/patrol_command.dart
bartekpacia: > Build a debug version of your app (default mode)

I know it's not perfect but works.

packages/patrol_cli/lib/src/runner/patrol_command_runner.dart
bartekpacia: I'm also wondering about it quite often, but so far I haven't found a solution that would (1) reduce duplication and (2) be flexible and testable.

`flutter_tools` has [AppContext](https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/base/context.dart) for this coupled with Zones, but I'm sure it's an overkill for `patrol_cli`.

packages/patrol_cli/lib/src/commands/build_ios.dart
bartekpacia: Yup, I see.

In general, I had a few approaches to introducing some more rigid abstractions in `patrol_cli`, but in the end, the benefit was minimal and it was only constraining.

I'm going to invoke `flutter_tools` as well - from my reading of their code, they prefer to duplicate code rather than to write a (almost always) leaky abstraction to "unify and not repeat". There are just too many subtle differences between Android and iOS build&test processes (and more in the future).

I would _love_ for the codebase to better, but I think it's just not a problem at our scale. And apparently, at `flutter_tools`'s scale as well.

PR #1131, opened 2023-03-28
Add analytics

bartekpacia:
This PR resolves https://github.com/leancodepl/patrol-internal/issues/22

[`package:usage` doesn't support GA4](https://github.com/dart-lang/usage/issues/185). It uses the older Universal Analytics instead, which is shutting down on June 1st, 2023.

`flutter_tools` is using `package:usage` as well, but if they don't manage to upgrade this package on time, I'm pretty sure they'll get 1 year extension, which we most probably won't.


bartekpacia:
Screenshot 2023-03-29 at 12 14 53 AM

PR #1157, opened 2023-03-30
Add analytics

bartekpacia:


bartekpacia:
Sending standalone events works.

Screenshot 2023-03-31 at 2 24 16 PM

PR #1173, opened 2023-04-03
Make `patrol update` also update `package:patrol` in pubspec

### LINKED ISSUE #1171: Make `patrol update` also update the `patrol` package in pubspec
### Description

When the developer runs `patrol update` and a `pubspec.yaml` file exists in the current working directory or above, we should also run `flutter pub upgrade patrol`. This will bring many more users to the latest version, and in turn, lower the number of issues resulting from using outdated versions of [patrol](https://pub.dev/packages/patrol).

Also, add a flag `--[no-]pub-upgrade`, to let users opt out of this behavior.

Related issue:
- #743

bartekpacia:
This PR resolves #1171


packages/patrol_cli/test/commands/update_test.dart
bartekpacia, line 1: TBH, it was completely useless.

PR #1182, opened 2023-04-05
Improve Google Analytics in CLI

bartekpacia:


packages/patrol_cli/pubspec.lock
```diff
@@ -129,6 +129,14 @@ packages:
       url: "https://pub.dev"
     source: hosted
     version: "2.0.2"
+  ci:
```
bartekpacia, line 132: really nice small package by Invertase

https://pub.dev/packages/ci

packages/patrol_cli/lib/src/analytics/analytics.dart
```diff
@@ -85,7 +86,13 @@ class Analytics {
     );
   }
 
-  bool get enabled => _config?.enabled ?? false;
+  bool get enabled {
+    if (ci.isCI) {
+      return false;
+    }
```
bartekpacia, line 92: Do we also want to track commands executed on CI (because we could by attaching `is_ci: bool` in event data.

PR #1205, opened 2023-04-14
POC Translate Dart-side Flutter integration tests to virtual JUnit tests on Android

bartekpacia:
This PR _paves the way_ for resolving #1026.

### It's a full-on POC ...

As of Mon Apr 24:
- it is not backward compatible
- missing tests (at least a few good smoke tests would mean a lot here)
- granularity in JUnit reports is per-Dart file, also the hierarchy is flat (non-existent)
- no handling of failed tests, nor any other kind of error reporting

Also, Dart's `test` library's API is also not extensibility-friendly. I've come to not like it.

### ... but the problem is solved

Take a look at this report.

Screenshot 2023-04-21 at 3 22 00 PM

You can see the `permissions_many_1_test.dart` and `permissions_many_2_test.dart` tests. These tests are they same (I copy-pasted them). They both grant _the same exact permissions_, and we'd expect the second one to fail, since the permissions are already granted and no permission request dialgo appears. Yet they both succeeded. That's because the app data is cleared between Dart entrypoints, and each test starts from a completely fresh slate (as if the app was just installed for the first time). This is achieved with [Android Test Orchestrator](https://developer.android.com/training/testing/instrumented-tests/androidx-test-libraries/runner#use-android).

In a nutshell, that was the purpose of this PR. And it's done.

### Next steps

1. Add a basic test failure reporting
1. Clean up code, especially replace `print()`s with `package:logging`
1. Run the (currently manually) generated `bundled_test.dart` on our GitHub Actions
1. Write a code generator for `bundled_test.dart` in `patrol_cli` (see [#1025](https://github.com/leancodepl/patrol/issues/1025))
3. Publish a `dev` version on pub.dev, make a POC in one of customer projects

### Overview

<details><summary>Flowchart</summary>
<p>

I tried.

```mermaid
---
title: Sequence diagram of initial Android Test Orchestrator run
---
    sequenceDiagram
    box Android
    participant MAT as MainActivityTest.java
    participant PJUR as PatrolJUnitRunner.java
    end
    box Flutter
    participant main()
    participant PatrolAppService
    end
    Note over PJUR: onCreate()
    PJUR->>main(): async launch MainActivity
    PJUR->>MAT: starts executing test
    Note over main(): test patrol_test_explorer
    Note over MAT: static testCases()
    MAT->>PJUR: asks for Dart tests
    Note over PJUR: static setUp()
    Note over PJUR: wait until Flutter tells us its ready
[...]

PR #981, opened 2023-02-20
Fix error reports containing "Multiple errors were thrown" without any helpful information

bartekpacia:
Attempt to fix #951

See also:
- [friendly tips on using `FlutterError.onError`](https://github.com/leancodepl/patrol/issues/922#issuecomment-1434813074)

PR #1220, opened 2023-04-24
Bring back basic test reporting for bundled tests

bartekpacia:
Follow-up of #1205

This PR merely adapts the (rudimentary) support for native reports we already had.

Exploration on how we could include more useful info in native reports was conducted in #981, but that was aborted. We should get back to it soon.


bartekpacia:
Done:

Screenshot 2023-04-25 at 12 14 10 AM

PR #1255, opened 2023-05-09
Fix `patrol` command crashing on Windows

bartekpacia:
The infamous `runInShell` strikes back. I once again forgot to include it.

Also added some smoke tests to make sure that the CLI at least doesn't crash when run without any commands.


bartekpacia:
There are test failures on Windows, but I don't have time to fix them now. I created [#1256](https://github.com/leancodepl/patrol/issues/1256) to track it.

Merging.

PR #1274, opened 2023-05-14
Fix test failures that occur only on Windows

### LINKED ISSUE #1256: Patrol CLI tests are failing on Windows because of path separator problems
See [this workflow run](https://github.com/leancodepl/patrol/actions/runs/4929622932/jobs/8809481942) and expand the `dart test` step to see tests failing.

### Cause

We use the global [`join()`](https://pub.dev/documentation/path/latest/path/join.html) function, which depends on the OS we're running in. This makes the test depend on the operating system's path separator, which is bad and causes issues like this one.

I verified this cause by changing `MemoryFileSystem.test()` to `MemoryFileSystem.test(style: FileSystemStyle.windows)` and tests started to fail, since I'm on macOS, so the path separator used by `join()` is `/`, but the mocked file system uses `/`.

In tests, we use [MemoryFileSystem](https://api.flutter.dev/flutter/package-file_memory/MemoryFileSystem-class.html), which by default always has the `/` as path separator. We should instantiate it twice, with `/` and `\` (and run all the tests twice).

### Solution idea

Don't use the global [`join()`](https://pub.dev/documentation/path/latest/path/join.html) function.
Use the [`Context.join()`](https://pub.dev/documentation/path/latest/path/Context/join.html) method instead.

bartekpacia:
This PR fixes #1256.

I'm not that happy with how less-readable the code has become.

### DIFF
--- a/packages/patrol_cli/lib/src/dart_defines_reader.dart
+++ b/packages/patrol_cli/lib/src/dart_defines_reader.dart
@@ -1,5 +1,4 @@
 import 'package:file/file.dart';
-import 'package:path/path.dart' show join;
 import 'package:patrol_cli/src/base/extensions/core.dart';
 
 class DartDefinesReader {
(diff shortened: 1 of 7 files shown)


bartekpacia:
The last file that's yet to be fixed is `test/ios/ios_test_backend_test.dart`.

Running tests on iOS is only supported on Mac, but I'd like to fix it anyway so tests are green when running on Windows.

shilangyu (approved):

Issue #1214, opened 2023-04-21
Clean up test bundling implementation


jBorkowska:

bartekpacia:
This will happen after #1205 is merged to `develop`.

bartekpacia:
See https://github.com/leancodepl/patrol/issues/1214

bartekpacia:
This is done.

Issue #1026, opened 2023-03-03
Replicate Dart tests on Android - crucial


mateuszwojtczak:
See https://github.com/flutter/flutter/issues/115751#issuecomment-1426753920

bartekpacia:
Prerequisite: #969

bartekpacia:
**WIP**

**TBD: Decide on granularity of remote test invocation (test file vs test method invocation)**
Seems like remote test file invocation is simpler, but will be less granular? Needs research.

PTR - PatrolTestRunner
AUT - app under test


Let's consider this example file with 3 integration tests. We'll refer to them as TST1, TST2, TST3:

```dart
void main() {
  patrol(
    'TST1 counter state is incremented when button is tapped',
    ($) async {
      await createApp($);
      await $(FloatingActionButton).tap();
      expect($(#counterText).text, '1');
      await $(FloatingActionButton).tap();
      expect($(#counterText).text, '2');
    },
  );

  patrol(
    'TST2 counter state is the same after reopening the app',
    ($) async {
      await createApp($);
      await $(FloatingActionButton).tap();
      expect($(#counterText).text, '1');
      await $.native.pressHome();
      await $.native.openApp();
      expect($(#counterText).text, '1');
      await $(FloatingActionButton).tap();
      expect($(#counterText).text, '2');
    },
  );

  patrol(
    'TST3 counter state is the after opening and closing notifications',
    ($) async {
      await createApp($);
      await $(FloatingActionButton).tap();
      expect($(#counterText).text, '1');
      await $.native.openNotifications();
      await $.native.closeNotifications();
      expect($(#counterText).text, '1');
      await $(FloatingActionButton).tap();
      expect($(#counterText).text, '2');
    },
  );
}

```

```mermaid
sequenceDiagram
    participant PTR
    participant AUT
    AUT->>PTR: Send Dart tests (TST1, TST2, TST3)
    AUT-->>PTR: Run TST1
    AUT-->>PTR: Run TST2
    AUT-->>PTR: Run TST3
```

bartekpacia:
Current progress: 
- All bundled (currently manually) Dart tests are collected by an artificial Dart test and sent to the native Android side using gRPC. The native side correctly translates the groups into native JUnit `Description`s (groups and tests). This is not completed yet, requires more work.

A problem appeared with dynamically creating JUnit test cases. We need to do this, so we can use Android Test Orchestrator. 

Currently I see 2 solutions to this problem, none of them perfect:
[...]

PR #1278, opened 2023-05-15
Simplify code and add more comments in native Android part

bartekpacia:
Follow up of #1026, related to #1214.

I'd like to make the code comments to explain as much as possible. Please be brutal.

### DIFF
--- a/packages/patrol/android/src/main/kotlin/pl/leancode/patrol/PatrolJUnitRunner.java
+++ b/packages/patrol/android/src/main/kotlin/pl/leancode/patrol/PatrolJUnitRunner.java
@@ -32,7 +32,7 @@ public class PatrolJUnitRunner extends AndroidJUnitRunner {
 
         // We need to know what Dart tests exist before we can request their execution.
         // To gather the Dart tests, we need to run the app during the Orchestrator's initial run.
-        // But by default, AndroidJUnitRunner doesn't run the app during the initial run, when it gathers the ests to execute.
+        // But by default, AndroidJUnitRunner doesn't run the app during the initial run, when it gathers the tests to execute.
         // That's why we override this onCreate().
 
         // This is only true when the Orchestrator requests a list of tests from the app during the initial run.
--- a/packages/patrol/android/src/main/kotlin/pl/leancode/patrol/PatrolJUnitRunner.java
+++ b/packages/patrol/android/src/main/kotlin/pl/leancode/patrol/PatrolJUnitRunner.java
@@ -1,3 +1,7 @@
+// Terminology note:
+// "Run a test" is used interchangeably with "execute a test".
+// "Run a Dart test" is used interchangeably with "request execution of a Dart test" and "execute Dart test".
+
 package pl.leancode.patrol;
 
 import android.app.Instrumentation;
@@ -28,7 +32,7 @@ public class PatrolJUnitRunner extends AndroidJUnitRunner {
 
         // We need to know what Dart tests exist before we can request their execution.
         // To gather the Dart tests, we need to run the app during the Orchestrator's initial run.
-        // But by default, AndroidJUnitRunner doesn't run the app during the initial run, when it gathers the test tests to execute.
+        // But by default, AndroidJUnitRunner doesn't run the app during the initial run, when it gathers the ests to execute.
         // That's why we override this onCreate().
 
         // This is only true when the Orchestrator requests a list of tests from the app during the initial run.
--- a/packages/patrol/android/src/main/kotlin/pl/leancode/patrol/PatrolJUnitRunner.java
+++ b/packages/patrol/android/src/main/kotlin/pl/leancode/patrol/PatrolJUnitRunner.java
@@ -8,6 +8,7 @@ import androidx.test.runner.AndroidJUnitRunner;
 import io.grpc.StatusRuntimeException;
 import pl.leancode.patrol.contracts.Contracts.DartTestGroup;
 
+import java.util.Arrays;
 import java.util.Objects;
 import java.util.concurrent.ExecutionException;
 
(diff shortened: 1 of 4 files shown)


packages/patrol/android/src/main/kotlin/pl/leancode/patrol/PatrolJUnitRunner.java
```diff
@@ -1,3 +1,7 @@
+// Terminology note:
+// "Run a test" is used interchangeably with "execute a test".
+// "Run a Dart test" is used interchangeably with "request execution of a Dart test" and "execute Dart test".
+
```
bartekpacia, line 4: This also applies to the iOS part. I wonder what could be the best place to put such things. Maybe some `IMPLEMENTATION.md` in the (currently non-existent) `docs` directory in repo root?

packages/patrol/android/src/main/kotlin/pl/leancode/patrol/PatrolJUnitRunner.java
```diff
@@ -1,3 +1,7 @@
+// Terminology note:
+// "Run a test" is used interchangeably with "execute a test".
+// "Run a Dart test" is used interchangeably with "request execution of a Dart test" and "execute Dart test".
```
shilangyu, line 3: Maybe instead of using interchangeable terms stick to one?

packages/patrol/example/android/app/src/androidTest/java/pl/leancode/patrol/example/MainActivityTest.java
bartekpacia, line 1: This is the file that developers wanting to use Patrol in their app must paste in the proper directory under `<flutter_root>`/android/app/src/androidTest/com/package/name`.

packages/patrol/android/src/main/kotlin/pl/leancode/patrol/PatrolJUnitRunner.java
```diff
@@ -25,9 +30,10 @@ protected boolean shouldWaitForActivitiesToComplete() {
     public void onCreate(Bundle arguments) {
         super.onCreate(arguments);
 
-        // We override onCreate(), because we need to gather the Dart tests before the tests are run.
-        // By default, AndroidJUnitRunner doesn't run the app during the initial run (when it gathers the test list).
-        // But in our case, the app must be run to gather the list of tests (because the tests (the Dart ones) live in the app itself).
+        // We need to know what Dart tests exist before we can request their execution.
+        // To gather the Dart tests, we need to run the app during the Orchestrator's initial run.
+        // But by default, AndroidJUnitRunner doesn't run the app during the initial run, when it gathers the ests to execute.
```
shilangyu, line 35:
```suggestion
        // But by default, AndroidJUnitRunner doesn't run the app during the initial run, when it gathers the tests to execute.
```

shilangyu (approved):

packages/patrol/android/src/main/kotlin/pl/leancode/patrol/PatrolJUnitRunner.java
bartekpacia: I was thinking about this, but it'd make the docs not nice to read.

Also, and it may not be the greatest argument, [but Flutter does it](https://github.com/flutter/flutter/blob/76a9f683259ea4ad1844250c9772b5437e04c200/packages/flutter/lib/src/foundation/change_notifier.dart#L37-L38) :D

shilangyu: Generally I am not a fan of using many terms when referring to the same thing. It is just a source of confusion.

packages/patrol_cli/lib/src/test_bundler.dart
```diff
@@ -30,10 +30,16 @@ ${generateImports(testFilePaths)}
 // END: GENERATED CODE
 
 Future<void> main() async {
-  // TODO: Create and use PatrolNativeTestService instead of NativeAutomator
   final nativeAutomator = NativeAutomator(config: NativeAutomatorConfig());
   final binding = PatrolBinding.ensureInitialized();
-  // Create a PatrolAppService.
+  
+  // This is the entrypoint of the bundled Dart test.
```
bartekpacia, line 36: @shilangyu I'd be happy if you reviewed this comment.

bartekpacia:
Area for improvement:

Write a `DartTestRunner` that abstracts the workaround around https://github.com/dart-lang/test/issues/1998.

bartekpacia:
I like the advice about the terminology above. I'll try to do something about it.

packages/patrol_cli/lib/src/test_bundler.dart
shilangyu: Is this generated `test_bundle` expected to be committed by users? 

These docs are nice, but I wonder if its a good place for them. Will anyone care what this file does exactly? Seems like an implementation detail.

One thing I dont exactly understand are the `// START: GENERATED CODE` and `// END: ...`. The whole file is generated. Maybe add a standard header of `// GENERATED CODE - DO NOT MODIFY BY HAND`? 

Also, instead of disabling specific lint rules (which assumes this code wont generate warnings for all linter configurations) you can disable them all with `// ignore_for_file: type=lint`.

bartekpacia: > Is this generated `test_bundle` expected to be committed by users?

No, it's not. I'll have to make it more clear by putting something like `// GENERATED CODE - DO NOT COMMIT INTO VCS`.

Actually, at first, I tried generating it in `.dart_tool/patrol_build`, but _very strange_ things happened (tests that were not referenced by `test_bundle` were being executed). I should probably try doing that again. 

> I wonder if it's a good place for them. Will anyone care what this file does exactly? Seems like an implementation detail.

I share these doubts, but I think that if these docs were in a different place than the code, it'd be easier for them to drift away over time.

Also, I expect almost nobody to bother looking at that file. And if somebody does, in the worst case they won't understand anything and close it, and in the best, they'll learn about how Patrol works :)

> One thing I dont exactly understand are the `// START: GENERATED CODE` and `// END: ...`. The whole file is generated. Maybe add a standard header of `// GENERATED CODE - DO NOT MODIFY BY HAND`?

The whole file is generated, but it's mostly static. The 2 parts surrounded with `// START: GENERATED CODE` are actually generated by walking the hierarchy of the `integration_test` directory.

> you can disable them all with `// ignore_for_file: type=lint`

Thanks - I unknowingly assumed the only ruleset in the world is 🖤leancode_lint💛.


I created PR #1283 to apply these suggestions.

PR #1287, opened 2023-05-18
Fix artifacts from test ios workflow

fylyppo:
This PR fixes:

- broken screen recordings
- blank simulator logs
- wrong format of .xcresult after unzipping

Additionaly I have changed logic of setting run status from tests to be more precise - if tests don't run, there'll be error status. If tests fail, there'll be failure status. It depends on tests exit code now.

### DIFF
--- a/.github/workflows/test-ios-simulator.yaml
+++ b/.github/workflows/test-ios-simulator.yaml
@@ -88,8 +88,8 @@ jobs:
         if: success() || failure()
         uses: kishikawakatsumi/xcresulttool@v1
         with:
-          title: "Patrol tests on ${{ matrix.device }}"
-          upload-bundles: "never"
+          title: Patrol tests on ${{ matrix.device }}
+          upload-bundles: never
           path: |
             ${{ env.XCRESULT_PATH }}
      
--- a/.github/workflows/test-ios-simulator.yaml
+++ b/.github/workflows/test-ios-simulator.yaml
@@ -59,7 +59,8 @@ jobs:
       - name: Run tests
         id: tests_step
         run: |
-          xcrun simctl io booted recordVideo "${{ matrix.device }}.mov" &
+          xcrun simctl io booted recordVideo --codec=h264 "${{ matrix.device }}.mp4" &
+          recordingpid="$!"
 
           xcrun simctl spawn booted log stream --type log --color none > all_simulator_logs.txt &
           xcrun simctl spawn booted log stream --type activity --color none > activity_simulator_logs.txt &
@@ -78,6 +79,8 @@ jobs:
             --exclude integration_test/webview_login_test.dart \
             --exclude integration_test/webview_stackoverflow_test.dart || TESTS_EXIT_CODE=$?
 
+          pkill -SIGINT $recordingpid
+          echo "TESTS_EXIT_CODE"=$TESTS_EXIT_CODE >> $GITHUB_ENV
           exit $TESTS_EXIT_CODE
 
       - name: Find xcresult path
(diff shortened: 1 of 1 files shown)


.github/workflows/test-ios-simulator.yaml
```diff
@@ -70,62 +72,57 @@ jobs:
...
-          path: ~/Library/Logs/CoreSimulator/${{ steps.start_simulator.outputs.udid }}/system.log
+          path: |
+            ${{ github.workspace }}/packages/patrol/example/all_simulator_logs.txt
 
       - name: Upload captured video to artifacts
         if: success() || failure()
         uses: actions/upload-artifact@v3
         with:
-          name: Captured video from ${{ matrix.device }}
+          name: Captured video from ${{ matrix.device }}.mp4
```
bartekpacia, line 115: Is this why the video was broken?xD

.github/workflows/test-ios-simulator.yaml
```diff
@@ -70,62 +72,57 @@ jobs:
...
+        if: success() || failure()
+        run: echo "XCRESULT_PATH=$(find ~/Library/Developer/Xcode/DerivedData -name "*.xcresult" -type d)" >> $GITHUB_ENV
+      
       - name: Publish test report to summary
         if: success() || failure()
         uses: kishikawakatsumi/xcresulttool@v1
         with:
-          title: "Patrol tests"
+          title: "Patrol tests on ${{ matrix.device }}"
           upload-bundles: "never"
```
bartekpacia, line 92: It's okay to not put quotes here.

.github/workflows/test-ios-simulator.yaml
```diff
@@ -11,8 +11,8 @@ jobs:
     runs-on: macos-latest
     timeout-minutes: 60
     outputs:
-      failure_status: ${{ steps.set_failure.outputs.failure_status }}
-      error_status: ${{ steps.set_error.outputs.error_status }}
+      failure_status: ${{ steps.set_status.outputs.failure_status }}
+      error_status: ${{ steps.set_status.outputs.error_status }}
```
bartekpacia, line 15: Why are you using 2 outputs, instead of just one, let's say it could be named `steps.set_status.outputs.status`?

And you'd set the `status` to a string value whatever you like, and compare it later.

fylyppo: There's a matrix and it's better to know that one job is broken (has error output) while other job from matrix has failed test (failure output). Simply, it can give more info.

fylyppo: Now if there's error from at least one job (doesn't matter if there's failed test in other job), the slack message will be "there's error". There won't be info that one job has failed test. So maybe I will add more precise message text which will include these two outputs?

.github/workflows/test-ios-simulator.yaml
fylyppo: XD, noo. I was too brutal while killing recording. The solution is to send SIGINT.

.github/workflows/test-ios-simulator.yaml
```diff
@@ -72,6 +72,7 @@ jobs:
             --exclude integration_test/service_cellular_test.dart \
             --exclude integration_test/service_wifi_test.dart \
             --exclude integration_test/swipe_test.dart \
+            --exclude integration_test/webview_leancode_test.dart \
```
bartekpacia, line 75: what happens with it? flaky?

fylyppo: I've removed this test from `exclude` because I wanted to see failed test for workflow testing purposes. Webview tests don't work on simulators

bartekpacia: Ah, okay.

.github/workflows/test-ios-simulator.yaml
bartekpacia: XD

bartekpacia (approved):

PR #1289, opened 2023-05-19
Set mason_logger to fixed version 0.2.4

rhalff:


bartekpacia:
What a stupid error, lol.

rhalff:
@bartekpacia I've lowered the version to 0.2.4 which should work.

bartekpacia:
Looks like the best solution will be to just fix the compilation errors.

bartekpacia:
See #1290.

Thanks for contributing anyway!

Issue #1265, opened 2023-05-10
Tests fail after finish when using Flutter 3.10


bartekpacia:
When running Patrol tests on Flutter 3.10, the tests start and run just fine, but when they finish, they fail with

```
java.lang.Exception: ══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞═════════════════
The following assertion was thrown running a test:
A SemanticsHandle was active at the end of the test.
All SemanticsHandle instances must be disposed by calling
dispose() on the SemanticsHandle.
When the exception was thrown, this was the stack:
#0 WidgetTester._verifySemanticsHandlesWereDisposed (package:flutter_test/src/widget_tester.dart:1051:7)
#1 WidgetTester._endOfTestVerifications (package:flutter_test/src/widget_tester.dart:1042:5)
#2 TestWidgetsFlutterBinding._runTestBody (package:flutter_test/src/binding.dart:998:22)
<asynchronous suspension>
The test description was:
counter state is the same after going to Home and switching
apps
═════════════════════════════════════════════════════════════════
at pl.leancode.patrol.PatrolTestRunner.run(PatrolTestRunner.kt:76)
at org.junit.runners.Suite.runChild(Suite.java:128)
at org.junit.runners.Suite.runChild(Suite.java:27)
at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
at androidx.test.internal.runner.TestExecutor.execute(TestExecutor.java:67)
at androidx.test.internal.runner.TestExecutor.execute(TestExecutor.java:58)
at androidx.test.runner.AndroidJUnitRunner.onStart(AndroidJUnitRunner.java:446)
at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:2361)
```

[...]

PR #1302, opened 2023-05-22
Temporary downgrade Flutter version to 3.7.12 for Android CI workflows

fylyppo:
We're downgrading from stable latest to stable 3.7.12 because of issue mentioned in https://github.com/leancodepl/patrol/issues/1265

### DIFF
--- a/.github/workflows/test-android-device.yaml
+++ b/.github/workflows/test-android-device.yaml
@@ -43,8 +43,7 @@ jobs:
       - name: Set up Flutter
         uses: subosito/flutter-action@v2
         with:
-          channel: 'stable'
-          flutter-version: '3.7.x'
+          flutter-version: 3.7.x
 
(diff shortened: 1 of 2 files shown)


.github/workflows/test-android-device.yaml
```diff
@@ -43,7 +43,8 @@ jobs:
       - name: Set up Flutter
         uses: subosito/flutter-action@v2
         with:
-          channel: ${{ matrix.flutter-channel }}
+          channel: 'stable'
```
bartekpacia, line 46: channel defaults to `stable`
```suggestion
```

fylyppo: It isn't said in action's readme. But let's remove it.

bartekpacia (approved):

bartekpacia:
let's go

PR #1303, opened 2023-05-22
Fix test bundling on physical iOS devices

### LINKED ISSUE #1059: Hot restart on physical iOS is not working
Described here: https://github.com/leancodepl/patrol/pull/1047#issuecomment-1461862167

### LINKED ISSUE #1297: Test bundling doesn't work on physical iOS devices
Since iOS 14, Flutter apps built in debug/profile mode must be run through a debugger.

Before test bundling, we've been accomplishing this by doing just this – starting the app under test in debug mode (with the lldb debugger attached using `ios-deploy`).

This approach breaks with test bundling because we're running every Dart test in a separate app process – which doesn't work, because there's no debugger attached.

In other words, this is prohibited:

```objective-c
[[[XCUIApplication alloc] init] launch];
```

We depend on the above heavily.

The only workaround I see around this is to build the app in release mode.

bartekpacia:
This PR fixes #1297.

### Caveats

- Running tests on iOS works only in release mode (`patrol test --release`).

- Hot Restart doesn't work.
    That's because, in a Flutter app built in release mode, there's no Dart VM that can JIT compile code.
    In other words, for Hot Restart to work, the app must be started in debug mode on iOS. 
    Starting the app in debug mode on iOS is mutually exclusive with test bundling.

    A workaround would be to not use test bundling at all when using Hot Restart.
    This would require re-adding the code that was present before test bundling.
   
    But even if we fixed this and re-added that, Hot Restart on iOS wouldn't work right away. We'd also have to fix #1059.

- Enabling or disabling airplane mode crashes the test suite.

cc: @mateuszwojtczak, @jBorkowska


bartekpacia:
Tested locally on my private iPhone and iPad and it works. Here's the xcresult:


Screenshot 2023-05-23 at 12 01 28 AM

packages/patrol_cli/lib/src/ios/ios_test_backend.dart
```diff
@@ -140,12 +136,6 @@ class IOSTestBackend {
       final subject = '${options.description} on ${device.description}';
       final task = _logger.task('Running $subject');
 
-      Process? iosDeployProcess;
```
bartekpacia, line 143: Calling [XCUIApplication.launch()](https://developer.apple.com/documentation/xctest/xcuiapp
lication/1500467-launch) with the debugger attached causes really nasty crashes.

bartekpacia:
TODO:

Print error message when trying to `patrol test` on physical iOS device.

> Running tests on a physical iOS devices requires the --release flag.

shilangyu:
Are these limitations documented anywhere (do you have any changelog?)?

>     * Running tests on iOS works only in release mode (`patrol test --release`).

What is the predicted impact of this limitation? Is there any benefit of debug mode testing?

>     * Enabling or disabling airplane mode crashes the test suite.

Because of some connection issue?

shilangyu (approved):

mateuszwojtczak (approved):
LGTM! Do we plan using [go-ios](https://github.com/danielpaulus/go-ios)?

bartekpacia:
> Do we plan using [go-ios](https://github.com/danielpaulus/go-ios)?

`go-ios` is a replacement for `ios-deploy`. And since in this PR I've removed our usage of `ios-deploy`, using `go-ios` doesn't make much sense anymore.

bartekpacia:
@shilangyu 

> Are these limitations documented anywhere (do you have any changelog?)?

Yes, we do, although it's pretty basic (https://pub.dev/packages/patrol/changelog).

We'll be rewriting the docs for 2.0 + writing a migration guide.

> * Enabling or disabling airplane mode crashes the test suite.

I don't know, but I suspect it's because Patrol (running in the app under tests) makes HTTP requests to the Patrol server (running in the instrumentation app). More specifically, all communication between these 2 apps happen over gRPC (which builds on HTTP/2’s long-lived connections).

And toggling (yes, toggling - everything works when airplane mode is disabled, and enabling it results in tests crashing) airplane mode probably somehow interferes with this.

PR #1315, opened 2023-05-23
Use correct, platform-native path separator everywhere

bartekpacia:
This PR is a follow-up of #1274. Most of the work to migrate to platform-native path separator was done there.

In this PR, I'm only refactoring the iOS-specific classes. This is low-prio because functionality provided by these iOS-specific classes can only be used when `patrol_cli` is running on a macOS host, which always has the Unix-style path separator.

### DIFF
--- a/packages/patrol_cli/lib/src/crossplatform/app_options.dart
+++ b/packages/patrol_cli/lib/src/crossplatform/app_options.dart
@@ -1,7 +1,7 @@
 import 'dart:convert' show base64Encode, utf8;
 
 import 'package:meta/meta.dart';
-import 'package:path/path.dart' show basename;
+import 'package:path/path.dart' as path;
 import 'package:patrol_cli/src/devices.dart';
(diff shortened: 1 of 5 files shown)


bartekpacia:
Current status: tests with `Platform.windows` are failing. The cause is most likely https://github.com/dart-lang/glob/issues/57.


Some logs:

```
created file at C:\projects\awesome_app\build\ios_integ\Build\Products\dev_iphonesimulator16.2-arm64.xctestrun
Looking for .xctestrun matching dev_iphonesimulator16.2*.xctestrun at C:\projects\awesome_app\build\ios_integ\Build\Products

    Error: No .xctestrun file was found at C:\projects\awesome_app\build\ios_integ\Build\Products
    package:patrol_cli/src/base/exceptions.dart 7:3         throwToolExit
    package:patrol_cli/src/ios/ios_test_backend.dart 238:7  IOSTestBackend.xcTestRunPat
```

PR #1325, opened 2023-05-30
Fix/scrolling rework

jBorkowska:
This PR resolves #1323

PR #1285, opened 2023-05-18
Feature/maybe settle

jBorkowska:
Resolves #1031 .

PR #1345, opened 2023-06-12
Add migration guide for v2 + fix outdated docs

### LINKED ISSUE #948: Improve native automation iOS setup docs
From #937 by @markwigzell:

> I recommend changing the documentation to just having the one "Getting Started" section incorporate the Setup into it, and have it handle both IOS and Android in the same "Getting Started" section. Also, delay introducing the actual test code until all necessary steps have been sequentially outlined. Isolate the optional parts clearly so the flow is not broken. Because of the back and forth nature of how the set up is documented, I actually ended up missing some essential aspect of Setup.

I agree.

bartekpacia:
This PR adds a migration guide from Patrol v1 to Patrol v2.

It also fixes #948.

It doesn't cover [scrolling changes](https://github.com/leancodepl/patrol/pull/1325) and [pump-and-settle behavior changes](#1285).


bartekpacia:
2 problems:
- provisioning of `packageName` should be simplified. Putting this in `app/build.gradle` in `testInstrumentationRunnerArguments` works only when running tests with Gradle. This breaks on e.g. cloud device farms:
  
  ```
  testInstrumentationRunnerArguments clearPackageData: "true", "packageName": "com.example.myapp"
  ```

  Specifying `packageName` in `MainActivityTest.java` is probably a better idea:

  ```java
  @RunWith(Parameterized.class)
  public class MainActivityTest {
      @Parameters(name = "{0}")
      public static Object[] testCases() {
          PatrolJUnitRunner.setPackageName("com.example.myapp");
          PatrolJUnitRunner.waitForPatrolAppService();
          return PatrolJUnitRunner.listDartTests();
      }
  
      // ...
  }
  ```

   We could also try getting it from `BuildConfig` – [see here](https://stackoverflow.com/questions/42205767/context-getpackagename-vs-buildconfig-application-id).

- setting `clearPackageData: "true"` (see above) causes behavior to be differ between Android and iOS. There's no equivalent of clearing all app's data on iOS.

PR #1360, opened 2023-06-15
Update Feature Parity table in the docs

bartekpacia:
cc @jBorkowska @fylyppo @mateuszwojtczak


bartekpacia:
This won't be reliable, the swipe-back gesture is not universal.

What we can do is add support for `$.native.swipe` and you can implement your poor man's "pressBack` on iOS.

PR #1384, opened 2023-06-20
New scrolling bug repro

bartekpacia:


bartekpacia:
Closed so it doesn't clutters open PRs.

This problem still persists.

PR #1396, opened 2023-06-24
Fix Slack messages from matrix

fylyppo:


.github/workflows/test-ios-simulator.yaml
```diff
@@ -58,6 +59,21 @@ jobs:
       - name: Set simulator location
         run: xcrun simctl location booted set 52.17469,21.03193
 
+      - name: Set tests to exclude
+        id: set_excluded_tests
+        run: |
+          TESTS_TO_EXCLUDE="
```
bartekpacia, line 65: i'm sure this is WIP but I'll just say that I'd like to see a slack message only with test name (e.g. "android_app_test") instead the full "--exclude integration_test/android_app_test.dart"

fylyppo: I'll parse this somehow :)

bartekpacia (approved):

PR #1405, opened 2023-06-27
Use flutter 3.10.x

fylyppo:


.cirrus.yml
```diff
@@ -15,8 +16,11 @@ test_linux_task:
   setup_flutter_script: |
     curl -LO https://github.com/fluttertools/fvm/releases/download/2.4.1/fvm-2.4.1-linux-x64.tar.gz
     tar -xf fvm-2.4.1-linux-x64.tar.gz
-    fvm install 3.7.12
-    fvm global 3.7.12
+    flutter_versions=$(fvm releases)
+    flutter_matching_versions=$(echo "$flutter_versions" | grep -v "pre" | grep -o " ${FLUTTER_VERSION}.[0-9]*")
+    latest_flutter_matching_version=$(echo "$flutter_matching_versions" | sort -rV | head -n 1)
+
+    echo "y" | fvm global $latest_flutter_matching_version
```
bartekpacia, line 23: that's great, but could we have this in a separate script?

Oh, and that script could be in https://github.com/leancodepl/mobile-tools :) named `install_flutter` or something like that

.cirrus.yml
```diff
@@ -42,10 +42,10 @@ test_linux_task:
   create_emulator_script: |
     avdmanager -s create avd -n MyAVD -k "$EMULATOR_IMAGE"
     cat << EOF >> ~/.android/avd/MyAVD.avd/config.ini
-    hw.cpu.ncore=4
+    hw.cpu.ncore=2
```
bartekpacia, line 45: this script should also be in mobile-tools

PR #1410, opened 2023-06-27
update promo graphics

bartekpacia:


bartekpacia:
Higher-ups decided to leave this as is.

PR #1419, opened 2023-06-28
Enable testing on BrowserStack on iOS

bartekpacia:
Sibling of #1402.


bartekpacia:
Closing because this PR uses Test Plans (aka BS XCUITest v2 API), which is not production ready yet.

New PR will be created.

PR #1443, opened 2023-06-30
Minor fixes to `test android emulator` workflow

bartekpacia:


bartekpacia:
@fylyppo From now on, you can use `ew-cli emulatorwtf.yaml:main` to run tests locally. Much shorter:)

Issue #1457, opened 2023-07-04
Custom Finder: finder.enterText won't work on Web because of Platform check


desmond206x:
### Steps to reproduce

Run this test

integration_test.dart
```
patrolTest('login with patrol', (PatrolTester $) async {
  await $.pumpWidgetAndSettle(AwesomeApp());

  await $(#emailTextField).enterText('charlie@root.me');
});
```

with 
`flutter drive --driver=test_driver/integration_test.dart --target=integration_test/integration_test.dart -d chrome`

### Actual results

```
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following UnsupportedError was thrown running a test:
Unsupported operation: Platform._operatingSystem

When the exception was thrown, this was the stack:
C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 266:49     throw_
C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/_internal/js_dev_runtime/patch/io_patch.dart 244:5                  _operatingSystem
C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/io/platform_impl.dart 56:40                                         get operatingSystem
C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/io/platform.dart 64:45                                              get _operatingSystem
C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 839:8  get
C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/io/platform.dart 153:47                                             get isIOS
C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 839:8  get
packages/patrol/src/custom_finders/patrol_tester.dart 317:18                                                                  enterText
packages/patrol/src/custom_finders/patrol_finder.dart 250:18                                                                  enterText
C:/b/s/w/ir/cache/builder/src/out/host_debug/dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 84:54               runBody
[...]

PR #1459, opened 2023-07-04
Added check of kIsWeb before usage of Platform / dart:io package

desmond206x:
Fix for https://github.com/leancodepl/patrol/issues/1457

Prevent usage of Platform package which uses dart:io, because this doesn't work on Flutter Web.


bartekpacia:
Thanks for contributing! I'll release this shortly as v2.0.2.

bartekpacia:
Thanks for contributing! I'll release this shortly as v2.0.2.

PR #1477, opened 2023-07-06
Add dart defines from file

anfernandezz:
In order to receive a JSON file for the configuration, add a `--dart-define-from-file` flag and use the same approach as flutter does [to create the gradle config with the JSON keys](https://github.com/blendthink/flutter/blob/2e8c3c0950e8fc64d9cdc901e5aa4c635e57528f/packages/flutter_tools/lib/src/build_info.dart#L310). For iOS, passing the JSON map to the environment in the `_processManager.start `method.

### DIFF
--- a/packages/patrol_cli/lib/src/crossplatform/app_options.dart
+++ b/packages/patrol_cli/lib/src/crossplatform/app_options.dart
@@ -11,12 +11,14 @@ class FlutterAppOptions {
     required this.flavor,
     required this.buildMode,
     required this.dartDefines,
+    this.dartDefinesPath = '',
   });
 
   final String target;
   final String? flavor;
   final BuildMode buildMode;
   final Map<String, String> dartDefines;
+  final String dartDefinesPath;
 
   /// Translates these options into a proper `flutter attach`.
   @nonVirtual
@@ -29,6 +31,10 @@ class FlutterAppOptions {
       for (final dartDefine in dartDefines.entries) ...[
         '--dart-define',
         '${dartDefine.key}=${dartDefine.value}',
+      ],
+      if (dartDefinesPath.isNotEmpty) ...[
+        '--dart-define-from-file',
+        dartDefinesPath
       ],
     ];
 
@@ -108,6 +114,11 @@ class AndroidAppOptions {
       }
 
       cmd.add('-Pdart-defines=$dartDefinesString');
+    }
+    if (flutter.dartDefinesPath.isNotEmpty) {
+      final dartDefinesFromFile =
+          '-Pdart-define-from-file=${flutter.dartDefinesPath}';
+      cmd.add(dartDefinesFromFile);
     }
 
     return cmd;
--- a/packages/patrol_cli/lib/src/crossplatform/app_options.dart
+++ b/packages/patrol_cli/lib/src/crossplatform/app_options.dart
@@ -12,6 +12,7 @@ class FlutterAppOptions {
     required this.buildMode,
     required this.dartDefines,
     required this.dartDefineFromFile,
+    required this.dartDefineFromFilePath,
   });
 
   final String target;
@@ -19,6 +20,7 @@ class FlutterAppOptions {
   final BuildMode buildMode;
   final Map<String, String> dartDefines;
   final Map<String, dynamic> dartDefineFromFile;
+  final String dartDefineFromFilePath;
 
   /// Translates these options into a proper `flutter attach`.
   @nonVirtual
@@ -32,6 +34,10 @@ class FlutterAppOptions {
         '--dart-define',
         '${dartDefine.key}=${dartDefine.value}',
       ],
+      if (dartDefineFromFilePath.isNotEmpty) ...[
+        '--dart-define-from-file',
+        dartDefineFromFilePath,
+      ]
     ];
 
     return cmd;
@@ -166,6 +172,10 @@ class IOSAppOptions {
         '--dart-define',
         '${dartDefine.key}=${dartDefine.value}',
       ],
+      if (flutter.dartDefineFromFilePath.isNotEmpty) ...[
+        '--dart-define-from-file',
+        flutter.dartDefineFromFilePath,
+      ]
     ];
 
     return cmd;
--- a/packages/patrol_cli/lib/src/crossplatform/app_options.dart
+++ b/packages/patrol_cli/lib/src/crossplatform/app_options.dart
@@ -118,7 +118,11 @@ class AndroidAppOptions {
       cmd.add('-Pdart-defines=$dartDefinesString');
     }
     if (flutter.dartDefineFromFile.isNotEmpty) {
-      cmd.addAll(_parseJsonToGradleConfig(flutter.dartDefineFromFile));
+      final dartDefinesFromFile = _overrideKeys(
+        json: flutter.dartDefineFromFile,
+        dartDefines: flutter.dartDefines,
+      );
+      cmd.addAll(_parseJsonToGradleConfig(dartDefinesFromFile));
     }
 
     return cmd;
@@ -131,6 +135,21 @@ class AndroidAppOptions {
     });
 
     return result;
+  }
+
+  Map<String, dynamic> _overrideKeys({
+    required Map<String, dynamic> json,
+    required Map<String, String> dartDefines,
+  }) {
+    final modified = Map<String, dynamic>.from(json);
+    if (dartDefines.isNotEmpty) {
+      modified.forEach((key, value) {
+        if (dartDefines.containsKey(key)) {
+          modified[key] = dartDefines[key];
+        }
+      });
+    }
+    return modified;
   }
 }
 
--- a/packages/patrol_cli/lib/src/crossplatform/app_options.dart
+++ b/packages/patrol_cli/lib/src/crossplatform/app_options.dart
@@ -20,7 +20,7 @@ class FlutterAppOptions {
   final BuildMode buildMode;
   final Map<String, String> dartDefines;
   final Map<String, dynamic> dartDefineFromFile;
-  final String dartDefineFromFilePath;
+  final String? dartDefineFromFilePath;
 
   /// Translates these options into a proper `flutter attach`.
   @nonVirtual
@@ -34,9 +34,9 @@ class FlutterAppOptions {
         '--dart-define',
         '${dartDefine.key}=${dartDefine.value}',
       ],
-      if (dartDefineFromFilePath.isNotEmpty) ...[
+      if (dartDefineFromFilePath != null) ...[
         '--dart-define-from-file',
-        dartDefineFromFilePath,
+        dartDefineFromFilePath!,
       ]
     ];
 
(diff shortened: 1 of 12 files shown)


packages/patrol_cli/lib/src/crossplatform/app_options.dart
```diff
@@ -109,9 +111,21 @@ class AndroidAppOptions {
 
       cmd.add('-Pdart-defines=$dartDefinesString');
     }
+    if (flutter.dartDefineFromFile.isNotEmpty) {
+      cmd.addAll(_parseJsonToGradleConfig(flutter.dartDefineFromFile));
+    }
```
bartekpacia, line 116: I think that `dartDefines` should take precedence over `dartDefinesFromFile` here.

It'd be useful to write a test for this.

valentindrolet:
I was about to create a PR that I started writing a few days ago for the same need, great job, thanks!

My take was to add the json key/value pairs to the `customDartDefines` but the result remains the same.

I think that it would be useful to pass this argument to the `flutter attach` and `flutter build` invocation to be able to use the variables on the Flutter side.

bartekpacia:
@fylyppo @jBorkowska Once this feature lands, I'd drop support for `.patrol.env`. We don't want to have 2 ways of doing the same thing ("passing dart defines from file").

packages/patrol_cli/lib/src/crossplatform/app_options.dart
```diff
@@ -11,12 +11,14 @@ class FlutterAppOptions {
     required this.flavor,
     required this.buildMode,
     required this.dartDefines,
+    this.dartDefineFromFile = const {},
```
bartekpacia, line 14: Let's make it `required`. This way we'll see all the places where it should be passed but currently isn't (for example [here](https://github.com/leancodepl/patrol/blob/f0bae1d5b2cb19d451d9f83ef83789f28ab4e5f3/packages/patrol_cli/lib/src/commands/develop.dart#L234-L239)).

anfernandezz: got you! makes sense

packages/patrol_cli/lib/src/crossplatform/app_options.dart
anfernandezz: do you mean to verify if the key wasn't passed in dartDefines and if it was, to take it out from dartDefinesFromFile?

anfernandezz:
> I was about to create a PR that I started writing a few days ago for the same need, great job, thanks!
> 
> My take was to add the json key/value pairs to the `customDartDefines` but the result remains the same.
> 
> I think that it would be useful to pass this argument to the `flutter attach` and `flutter build` invocation to be able to use the variables on the Flutter side.

thanks @valentindrolet! Will add it to those methods :)!

packages/patrol_cli/lib/src/crossplatform/app_options.dart
bartekpacia: I mean something like this:

1. `EXAMPLE_KEY=original` is defined in the JSON file
2. `--dart-define EXAMPLE_KEY=overridden` is passed to a command

Expected result: `EXAMPLE_KEY=overridden` is present in the app.

valentindrolet:
> @fylyppo @jBorkowska Once this feature lands, I'd drop support for `.patrol.env`. We don't want to have 2 ways of doing the same thing ("passing dart defines from file").

I think it would be interesting to keep these 2 ways:
- the json file to store environment variables used for both running and testing the app
- the .patrol.env file to store tests specifics variables and not interfere with the global app config
What do you think ?

packages/patrol_cli/lib/src/crossplatform/app_options.dart
anfernandezz: makes sense!

anfernandezz: so if the key exists in dart-defines and in the JSON file, override the values from the JSON file

bartekpacia: yup

bartekpacia:
@valentindrolet We'll definitely leave them both for the time being, but what worries me is that people might define the same `dart-define` in the JSON file and in `.patrol.env`. Which one should be considered more important?

anfernandezz:
I have added the changes, I had to add a parameter with the path of the config file for the `flutter attach`  and `flutter build` methods as per my understanding in that case we pass the path and flutter handles it (correct me if I'm wrong!).

I added a test for the `AppOptions` to verify it overrides the` dart-defines-from-file` values with the ones from the `dart-defines`, I'm not sure how to test it for iOS tho, because I pass the JSON as the process environment. 
 
Please let me know if anything @bartekpacia

packages/patrol_cli/lib/src/ios/ios_test_backend.dart
```diff
@@ -110,6 +110,7 @@ class IOSTestBackend {
         options.buildForTestingInvocation(),
         runInShell: true,
         workingDirectory: _fs.currentDirectory.childDirectory('ios').path,
+        environment: Map.castFrom(options.flutter.dartDefineFromFile),
```
bartekpacia, line 113: I'm pretty sure specifying `environment` here (and below as well) is not needed.

The dart defines are passed to `flutter build ios --config-only` invocation above, and it handles base64-encoding them in `ios/Flutter/Generated.xcconfig`.

packages/patrol_cli/test/crossplatform/app_options_test.dart
```diff
@@ -103,6 +111,75 @@ void main() {
...
+        expect(
+          invocation,
+          equals([
+            r'.\gradlew.bat',
+            ':app:assembleDevDebugAndroidTest',
+            r'-Ptarget=C:\Users\john\app\integration_test\app_test.dart',
+            '-Pdart-defines=RU1BSUw9dXNlckBleGFtcGxlLmNvbQ==,UEFTU1dPUkQ9bnk0bmNhdA==,Zm9vPWJhcg==',
+            '-PEMAIL=user@example.com',
+            '-PPASSWORD=ny4ncat',
+            '-PFLAVOR=dev',
```
bartekpacia, line 151: That's a nice test (and below as well!), but I think it's slightly wrong.

First of all, it's not needed to pass these values twice: once in `Pdart-defines` (in base64-encoded form) and once again in `Pkey=value`. Specifying them in `-Pdart-defines` is enough.

And second possible problem: there are only 3 defines in `-Pdart-defines`: `EMAIL`, `PASSWORD`, and `foo`. But there's also `FLAVOR`, and I don't see it encoded in `-Pdart-defines`. I guess this probably stems from the first problem.

packages/patrol_cli/test/crossplatform/app_options_test.dart
```diff
@@ -103,6 +111,75 @@ void main() {
...
+        expect(
+          invocation,
+          equals([
+            './gradlew',
+            ':app:assembleDevDebugAndroidTest',
+            '-Ptarget=/Users/john/app/integration_test/app_test.dart',
+            '-Pdart-defines=RU1BSUw9dXNlckBleGFtcGxlLmNvbQ==,UEFTU1dPUkQ9bnk0bmNhdA==,Zm9vPWJhcg==',
+            '-PEMAIL=user@example.com',
+            '-PPASSWORD=ny4ncat',
+            '-PFLAVOR=dev',
```
bartekpacia, line 178: same situation as above

packages/patrol_cli/lib/src/ios/ios_test_backend.dart
```diff
@@ -156,6 +157,7 @@ class IOSTestBackend {
         ),
         runInShell: true,
         workingDirectory: _fs.currentDirectory.childDirectory('ios').path,
+        environment: Map.castFrom(options.flutter.dartDefineFromFile),
```
bartekpacia, line 160: same as above

bartekpacia (commented):
lgtm after the comments are resolved 👍🏻 

thank you!

packages/patrol_cli/test/crossplatform/app_options_test.dart
anfernandezz: oooh okay, got you! I'm worried tho about not passing both (I don't know if you remember my issue) but since my app was launched using `--dart-defines-from-file` it wouldn't take the `--dart-defines` values as valid. So in that case, wouldn't it be better to send them in the `PKEY=VALUE` format without the `Pdart-defines=KEY=VALUE`? hope that makes sense.

In the case of FLAVOR it was that it was in the config file but not in the dart-defines, so it shouldn't be taking it anyways? So it overrides all the values from the dart defines, but it also keeps the ones that aren't passed

anfernandezz: nvm, I realized that I usually use either the config file or the dart defines, not sure if the first point still applies.
The second one, I'm still curious about

valentindrolet:
> @valentindrolet We'll definitely leave them both for the time being, but what worries me is that people might define the same `dart-define` in the JSON file and in `.patrol.env`. Which one should be considered more important?

I'd say the variables defined in the .patrol.env should override those defined in the JSON file but you are right, it may be confusing to have both files handled the same way but with different formats.

bartekpacia:
`flutter` supports passing multiple `--dart-define-from-file` in a single invocation, for example:



```
$ cat defines.json
{
    "EXAMPLE_KEY": "This is from the JSON file"
}
$ cat defines2.json
{
    "EXAMPLE_KEY": "This is from the second JSON file"
}
$ flutter run --dart-define-from-file defines.json --dart-define-from-file defines2.json  --dart-define 'EXAMPLE_KEY=I override you!'
```

This will make it easy for users to have several dart-define JSON files with config for various environments, etc. I'm confident that we should deprecate `.patrol.env` and migrate to this.

(btw, the result of the above is `EXAMPLE_KEY=This is from the second JSON file` - a bit strange)

bartekpacia:
This is not as simple as it seems at first. There are some annoying edge cases.

### Code duplication

Ideally, this would be very simple – just forward the `--dart-define-from-file` args from `patrol_cli` to `flutter build`/`flutter run`. Unfortunately, it's not possible because of our how `patrol_cli` interacts with Gradle. We build the apps by invoking `gradlew` directly, which bypasses `flutter`.

This means that:
- we have to duplicate the dart-define-handling code that's already written in `flutter`
- when `flutter` adds a new feature (e.g. [`--dart-defines-from-file` will also work for .env files](https://github.com/flutter/flutter/pull/128668)), we won't get that "for free". We'll have to manually write code for that.

### Behavior

I think that `patrol_cli` should handle `--dart-define`s and `--dart-define-from-file`s the same way as `flutter` handles them. The problem is: I don't think `flutter` handles them well.

For example: when `--dart-define-from-file` is passed, all `--dart-define`s are ignored. This is complete

I want to report an issue to Flutter.

Thank you very much @anfernandezz for this PR. IMHO the best thing we can do right now is to wait.

bartekpacia:
I reported the issues:
- https://github.com/flutter/flutter/issues/130599
- https://github.com/flutter/flutter/issues/130604

anfernandezz:
Thanks for the fully detailed explanation @bartekpacia, makes total sense to wait! I'll be also checking updates on those issues.

bartekpacia:
Closing to keep our open PRs list clean. Also created an issue to track it: #1742.

Nevertheless this PR will surely come in useful once the bug in Flutter is fixed and we get to implement it.

mateuszwojtczak:
Thanks for your work @anfernandezz, FYI we're implementing this now in #2232

Issue #1474, opened 2023-07-06
Some of iOS tests are marked as failed but should be successful on physical iOS device


fylyppo:
### Steps to reproduce

1. Go to my branch `ci/small_fixes`
2. Run `patrol test --release --verbose -t integration_test/webview_leancode_test.dart` in patrol example app on physical iOS device

### Actual results

You can see:
```
PatrolServer: INFO: done tapping on view with text "Subscribe" in app pl.leancode.patrol.Example
 2023-07-06 15:46:27.633293+0200 RunnerUITests-Runner[1779:432270] PatrolServer: INFO: result: ()
/Users/fylyppo/Documents/GitHub/patrol/packages/patrol/example/ios/RunnerUITests/RunnerUITests.m:5: error: -[RunnerUITests webviewLeancodeTest] : ((response.passed) is true) failed
```
Tapping on "Subscribe" button is the last action in my test and it's done as you can see in logs so test should be successful.
I have the same problem also in other test but I can't find any connections between them. All failing tests:
* webview_leancode_test
* open_quick_settings_test
* service_airplane_mode_test
* service_bluetooth_test
* service_cellular_test
* webview_hackernews_test
* webview_stackoverflow_test

### Logs

<details>
<summary>Logs</summary>

```
patrol test --release --verbose -t integration_test/webview_leancode_test.dart
You're using Patrol CLI 2.0, which has breaking changes.
Read the migration guide at https://patrol.leancode.co/v2.
Disable this warning by setting the PATROL_MIGRATED environment variable.

Verbose mode enabled. More logs will be printed.
Received 1 test target(s)
Received test target: /Users/fylyppo/Documents/GitHub/patrol/packages/patrol/example/integration_test/webview_leancode_test.dart
Generated entrypoint /Users/fylyppo/Documents/GitHub/patrol/packages/patrol/example/integration_test/test_bundle.dart with 1 bundled test(s)
$ flutter --no-version-check devices --machine
No device specified, using the first one (Julia’s iPhone)
Received 1 device(s) to run on
Received device: Julia’s iPhone
Received 7 --dart-define(s) (0 custom, 7 internal)
Received internal --dart-define: PATROL_WAIT=0
Received internal --dart-define: PATROL_APP_PACKAGE_NAME=pl.leancode.patrol.example
Received internal --dart-define: PATROL_APP_BUNDLE_ID=pl.leancode.patrol.Example
[...]

PR #1491, opened 2023-07-10
Add workaround for `SemanticsHandle` on physical iOS devices

bartekpacia:
This PR implements a workaround for #1474.


nc-jeni:
Why is this only added for the iOS platform, and not for the Android and emulator platform?

@bartekpacia

PR #1509, opened 2023-07-11
Fix cryptic failures with Gradle builds

bartekpacia:


bartekpacia:
Looks like using the `plugins {}` API isn't a great idea right now. This is a warning I get after running `./gradlew :app:assembleDebug`:

```
The Kotlin Gradle plugin was loaded multiple times in different subprojects, which is not supported and may break the build.
This might happen in subprojects that apply the Kotlin plugins with the Gradle 'plugins { ... }' DSL if they specify explicit versions, even if the versions are equal.
Please add the Kotlin plugin to the common parent project or the root project, then remove the versions in the subprojects.
If the parent project does not need the plugin, add 'apply false' to the plugin line.
See: https://docs.gradle.org/current/userguide/plugins.html#sec:subprojects_plugins_dsl
The Kotlin plugin was loaded in the following projects: ':location', ':patrol'
```

This was not occurring when we were not using the `apply` plugin API.

PR #1520, opened 2023-07-12
Stop passing the flutter build mode to the gradle test tasks

ahmdaeyz:
Hello, 
Thanks for your efforts :)
This PR fixes the errors the users will get when they pass the --profile and --release build args:

` patrol test -t integration_test/app_test.dart --verbose --release`
```
FAILURE: Build failed with an exception.
        
        * What went wrong:
        Cannot locate tasks that match ':app:assembleReleaseAndroidTest' as task 'assembleReleaseAndroidTest' not found in project ':app'.
```

` patrol test -t integration_test/app_test.dart --verbose --profile`
```
FAILURE: Build failed with an exception.
        
        * What went wrong:
        Cannot locate tasks that match ':app:assembleProfileAndroidTest' as task 'assembleProfileAndroidTest' not found in project ':app'.
```

### DIFF
--- a/packages/patrol_cli/lib/src/crossplatform/app_options.dart
+++ b/packages/patrol_cli/lib/src/crossplatform/app_options.dart
@@ -59,7 +59,7 @@ class AndroidAppOptions {
     // for example: assembleDevDebugAndroidTest, assembleReleaseAndroidTest
     return _toGradleInvocation(
       isWindows: isWindows,
-      task: 'assemble$_effectiveFlavor${_buildMode}AndroidTest',
+      task: 'assemble${_effectiveFlavor}DebugAndroidTest',
     );
   }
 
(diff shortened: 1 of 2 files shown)


bartekpacia:
Hi, thanks for the PR! 

I'm just not sure this is the right behavior. If Gradle test tasks for `--profile` and `--release` build modes don't exist (we can check that easily), we should prompt the user to add them, not silently fallback to `debug`.

bartekpacia:
We should update documentation, and not mask the error from the user.

ahmdaeyz:
Sorry for my late reply. I'm actually interested in learning how to create a Gradle task that can do that. I've researched the topic and ended up adding some config to my app's build.gradle. That resulted in creating the `assembleProfileAndroidTest` task and patrol started but what happened was that It just opened and closed. Can't remember if It reported failure in executing the tests tho (I'll confirm this once I've access to my work macbook). I think the accompanying apk used for testing will be/is missing what is essential to execute the tests. 

I get your take on having to be explicit about this to the users tho.

bartekpacia:
> That resulted in creating the assembleProfileAndroidTest task and patrol started but what happened was that It just opened and closed.

This sounds concerning – it'd be great if you could share a logcat. There'll be probably something useful in there.

PR #1545, opened 2023-07-18
Fix/scrolling speed

### LINKED ISSUE #1383: New scrolling is too slow
Reproduction here: #1384, but basically, add the following line at the end of `example_test`

```dart
await $('EXAMPLE_KEY: at the bottom').scrollTo().tap();
```

---

Clone it and run:

```
patrol test -t integration_test/example_test.dart --dart-define EXAMPLE_KEY="at the bottom"
```

Observe slow scroll:

https://github.com/leancodepl/patrol/assets/40357511/1f83cb83-4385-4bb6-8f2e-87b7fbbcc29c

jBorkowska:
This PR resolves #1383

### DIFF
--- a/packages/patrol/lib/src/custom_finders/patrol_tester.dart
+++ b/packages/patrol/lib/src/custom_finders/patrol_tester.dart
@@ -591,10 +591,20 @@ class PatrolTester {
       finder: view,
       tester: this,
     ).waitUntilVisible();
+    AxisDirection direction;
+    if (scrollDirection == null) {
+      if (view.evaluate().first.widget is Scrollable) {
+        direction = tester.firstWidget<Scrollable>(view).axisDirection;
+      } else {
+        direction = AxisDirection.down;
+      }
+    } else {
+      direction = scrollDirection;
+    }
 
     return TestAsyncUtils.guard<PatrolFinder>(() async {
       Offset moveStep;
-      switch (tester.firstWidget<Scrollable>(view!).axisDirection) {
+      switch (direction) {
         case AxisDirection.up:
           moveStep = Offset(0, delta);
           break;
--- a/packages/patrol/lib/src/custom_finders/patrol_tester.dart
+++ b/packages/patrol/lib/src/custom_finders/patrol_tester.dart
@@ -15,7 +15,7 @@ class PatrolTesterConfig {
     @Deprecated('Use settleBehavior argument instead') this.andSettle = true,
     // TODO: change default to trySettle, see #1369 (https://github.com/leancodepl/patrol/issues/1369)
     this.settlePolicy = SettlePolicy.settle,
-    this.dragDuration = const Duration(milliseconds: 500),
+    this.dragDuration = const Duration(milliseconds: 100),
     this.settleBetweenScrollsTimeout = const Duration(seconds: 2),
   });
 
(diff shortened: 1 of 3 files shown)


packages/patrol/lib/src/custom_finders/patrol_tester.dart
```diff
@@ -435,35 +455,41 @@ class PatrolTester {
...
         );
         iterationsLeft -= 1;
       }
 
       if (iterationsLeft <= 0) {
         throw WaitUntilExistsTimeoutException(
           finder: finder,
-          // TODO: set reasonable duration
-          duration: pumpDuration,
+          // TODO: set reasonable duration or make new exception for this case
```
bartekpacia, line 491:
```suggestion
          // TODO: set reasonable duration or create new exception for this case
```

packages/patrol/test/patrol_tester_test.dart
```diff
@@ -868,10 +873,17 @@ void main() {
           expect(find.text('top text').hitTestable(), findsNothing);
           expect(find.text('bottom text').hitTestable(), findsNothing);
 
-          await tester.scrollUntilExists(finder: find.text('top text'));
+          await tester.scrollUntilExists(
+            finder: find.text('top text'),
+            maxScrolls: 100,
```
bartekpacia, line 878: Did we have to change these tests because we modified the default?

If we really want to change the tests, then I'd prefer to change the tests so they pass with the default value.

bartekpacia (commented):
a few small comments, in general: lgtm

bartekpacia (approved):

packages/patrol/lib/src/custom_finders/patrol_tester.dart
```diff
@@ -79,7 +92,7 @@ class PatrolTesterConfig {
 const defaultScrollDelta = 64.0;
 
 /// Default maximum number of drags during scrolling.
-const defaultScrollMaxIteration = 100;
+const defaultScrollMaxIteration = 15;
```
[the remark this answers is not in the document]
jBorkowska: yes, in most cases (as a human) you don't need to do more than a few scrolls, most apps are designed this way. It can be modified, so the 1% of users can change it. I decreased it drastically, bc in case there is no such element that we are scrolling to, the test fails faster.

packages/patrol/test/patrol_tester_test.dart
jBorkowska: In this line it is unnecessary, but in other places - yes

packages/patrol/test/patrol_tester_test.dart
```diff
@@ -330,7 +330,11 @@ void main() {
           await tester.dragUntilExists(
             finder: find.text('bottom text'),
             view: find.byType(Scrollable),
+            // defaultScrollDelta can't be used, because it not possible
+            // to achieve a state, in which 'bottom text' exists, but
+            //is not visible
```
bartekpacia, line 335:
```suggestion
            // is not visible
```

packages/patrol/test/patrol_tester_test.dart
```diff
@@ -868,10 +872,19 @@ void main() {
...
+            finder: find.text('top text'),
+          );
           expect(find.text('top text').hitTestable(), findsOneWidget);
           expect(find.text('bottom text'), findsNothing);
-          await tester.scrollUntilExists(finder: find.text('bottom text'));
+          await tester.scrollUntilExists(
+            finder: find.text('bottom text'),
+            // defaultScrollDelta can't be used, because it not possible
+            // to achieve a state, in which 'bottom text' exists, but
+            //is not visible
```
bartekpacia, line 884:
```suggestion
            // is not visible
```

bartekpacia (approved):

PR #1567, opened 2023-07-20
Update protobuf to 3.0.0

hchan10:
* Regenerate new protobuf files

* Resolve issues caused by the protobuf update

### DIFF
--- a/packages/patrol/example/integration_test/notifications_test.dart
+++ b/packages/patrol/example/integration_test/notifications_test.dart
@@ -19,7 +19,7 @@ void main() {
       await Future<void>.delayed(const Duration(seconds: 5));
 
       await $.native.tapOnNotificationBySelector(
-        Selector(textContains: 'Someone liked'),
+        Selector.create()..textContains = 'Someone liked',
       );
 
       await $('Tapped notification with ID: 1').waitUntilVisible();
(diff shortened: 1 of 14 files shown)


bartekpacia:
Thanks for the PR!

I'd like to do it but for now, we've gotta support Flutter 3.3 and Flutter 3.7.

Also, the fact that API changed so much (`Selector(text: 'hey')` -> `Selector..text = 'hey'`) will probably mean we'll have to add some compatibility code by hand. We don't want breaking changes because of a dependency update.

shilangyu:
After a backlash from this change in generation (making migration extremely tedious and result in worse code), they are working on adding a flag to restore the old behavior: https://github.com/google/protobuf.dart/pull/855. Probably worth to wait for it?

bartekpacia:
Thanks for info @shilangyu, I' happy they're bringing back the old syntax. Let's wait.

bartekpacia:
https://github.com/google/protobuf.dart/pull/855 is done, but we're blocked because its newewst version requires Dart 2.19, and our current min is Dart 2.18 (from Flutter 3.3).

bartekpacia:
Thank you for the contribution, but we decided to ditch protobufs and gRPC in #1567.

PR #1605, opened 2023-08-07
Add Allure runner support

bartekpacia:


packages/patrol/example/android/app/src/androidTest/java/pl/leancode/patrol/example/PatrolAllureJUnitRunner.kt
```diff
@@ -0,0 +1,39 @@
...
+            arguments.getCharSequence("listener"),
+            AllureJunit4::class.java.name,
+            ExternalStoragePermissionsListener::class.java.name.takeIf { useTestStorage }
+        ).joinToString(separator = ",")
+        arguments.putCharSequence("listener", listenerArg)
+        super.onCreate(arguments)
+    }
+
+    protected open fun createAllureAndroidLifecycle() : AllureAndroidLifecycle =
+        createDefaultAllureAndroidLifecycle()
```
bartekpacia, line 25: why is it `protected`?

bartekpacia (commented):
This is very nice, but I don't think we should modify our example app to use Allure. This might confuse many beginner users who try to set up Patrol and often look on our example app. IMHO example app should be as vanilla (in terms of setup) as possible.

Another option is to add `PatrolAllureJUnitRunner` to `patrol` (instead of our examplea app), but then we make `patrol` depend on `io.qameta.allure:allure-kotlin*` - not good as well.

I'd lean towards adding a page in our docs explaining how to create such a custom runner. Essentially copy-paste code from this PR + some explanation.

mateuszwojtczak:
To get the results after the run, execute:
```
adb exec-out sh -c 'cd  /sdcard/googletest/test_outputfiles && tar cf - allure-results' | tar xvf - -C build/reports
```

bartekpacia:
Closing in favor of #1621.

PR #1621, opened 2023-08-10
Add "integrations" docs section

bartekpacia:


fylyppo:
Maybe add also a section how to run an allure report locally in browser?

bartekpacia:
The thing is I never did that. Can you point me to some tutorial/docs?

fylyppo:
And if so maybe we should link also docs to how to setup Allure report? XD

fylyppo:
https://github.com/allure-framework/allure2

bartekpacia:
> And if so maybe we should link also docs to how to setup Allure Report? XD

Great idea. At first I just wanted this PR to be a replication of #1605 but in docs, but if there's more good we can do, let's do it.

fylyppo (approved):

Issue #1589, opened 2023-07-28
Add possibility to set derivedDataPath for ios test command


kkalisz:
Hello, 

Currently, when running tests for iOS, derivedData path of xcresult is generated based on the runner name and some generated id.
` /...Developer/Xcode/DerivedData/Runner-bxllheovvqkgdaenftcnnkrzxfib/Logs/Test/Test-Runner-2023.07.28_21-38-27-+0200.xcresult`

It's not easy to extract the correct path when we would like to upload test results.
One of the possible solutions would be to specify derivedData path for the `test` command. 
I have already made some code modifications in the code and it works fine, so I would like to work on PR for this feature.

But I need some advices in terms of the final solution and overall code of conduct.

1. I think that by default it should work as is now, to not break actual usage
2. Should we allow to specify the full path or rather use some hardcoded path? like for build command.
3. Should we just give the possibility to specify `--derivedDataPath` property in the test command? This may not be super pretty as it will have no effect on other platforms, or maybe we have a different option.

 Best Regards,
 Kamil

fylyppo:
Hi, we were thinking about https://github.com/leancodepl/patrol/issues/1190 in the past but it's not refined yet. As you can see we're collecting use cases, so feel free to chip in if you have some in mind. At the moment our main maintainer @bartekpacia is on vacations but when he's back, will share his ideas and respond you.

bartekpacia:
This is a reasonable request. Thanks for filing it.

We can either:
- solve only this issue by adding this single additional option to `patrol test`, `patrol build ios`, and `patrol develop`
- fix the problem #1190 altogether: add a `--xcodebuild-params` flag to `patrol test`, `patrol build ios`, and `patrol develop`

bartekpacia:
See #1623 which mitigates this.

Kendru98:
I think that is fixed already, and we're also printing paths

PR #1623, opened 2023-08-10
Change location of generated iOS test results into `build/ios_results_$timestamp`

bartekpacia:
Thanks to this, the location of iOS reports will be much easier to find.

See also: #1589

### DIFF
--- a/packages/patrol_cli/CHANGELOG.md
+++ b/packages/patrol_cli/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Unreleased
 
 - Change the location of iOS test reports to `build/ios_results` with timestamp
-  appended (#1623)
+  appended. The path to the report is printed after `patrol test` exits (#1623)
 
 ## 2.1.0
 
--- a/packages/patrol_cli/CHANGELOG.md
+++ b/packages/patrol_cli/CHANGELOG.md
@@ -1,4 +1,10 @@
+## Unreleased
+
+- Change the location of iOS test reports to `build/ios_results` with timestamp
+  appended (#1623)
+
 ## 2.1.0
+
 - Add `--no-generate-bundle` flag (#1565)
 - Don't check for `patrol_cli` updates on CI (#1557)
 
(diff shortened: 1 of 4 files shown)


bartekpacia:
@fylyppo Can this break our CI workflows?

bartekpacia:
We could also disable diagnostics to make `.xcresult`s much smaller in size: https://developer.apple.com/forums/thread/698054

kkalisz:
Hi @bartekpacia,

Really great job, now it should be easier to find result data.
Just a few points that we don't really need action on, but still may be valid to explore.

- the name is not fixed, but it should be easy to put some regexp rules to get the correct folder.
- depending on how people were searching for test results, it may be breaking change. But taking into consideration that searching for test results was not easy and there were not a lot of issues reported to this repository regarding that, I think it should not affect most of the users.
- with a few subsequent runs we will have a few result folders, this should not be a real problem as in most cases in CI environment we will have a completely new vm instance each run, even with some specific bare metal self-hosted runners that will keep the files between runs, we can clear the folder before the tests invocation.

Regards

bartekpacia:
Thanks for voicing your opinion @kkalisz.

I'm torn between overriding the previous test run results vs creating a new one. By default, iOS creates new directories for new test runs, and Android overrides the previous result.

At first I wanted to honor the OS default but I'm not so sure about it. Maybe it's better to always override and be consistent across platforms.

And yes, the size of iOS reports can get big :)

bartekpacia:
FTR: #1630 which migrates our own CI workflows to new location of iOS reports.

kkalisz:
Oh and with having multiple folders... I was mostly concerned about having a situation that we are not able to easily retrieve test results because again we have few folders. But as I said this may not be real use case and even if it is. we can just clear folders.

Having a static/predictable path is especially helpful in CI environments where you would like just to know the path (or set the path) instead of playing with regexp to retrieve the correct folder. But even with a current proposal with timestamp, it's easy enough to retrieve this path.

PR #1707, opened 2023-09-18
Update getting-started.mdx

ethnosdev:
Clarify where this relative path is located. (Otherwise some users might think you are talking about the ios folder.)


bartekpacia:
Thanks for clarifying that!

PR #1721, opened 2023-09-19
Add support for basic test lifecycle callbacks

### LINKED ISSUE #1503: setUp and tearDown functions (and setUpAll and tearDownAll, which won't be supported when sharding is used)

bartekpacia:
This PR resolves #1503.

### Possible concerns, defused

Problem: `PatrolBinding.ensureInitialized()` adds a `setUp` and a `tearDown`.
Answer: Yes, but PatrolBinding is a singleton (like all bindings), and its initialization happens in the generated `test_bundle.dart`.

Problem: Reporting results back to the native side depends on `tearDown`
Answer: Yes, but tearDown execute in the reverse order, so the first `tearDown` that `PatrolBinding` adds executes the last. This is what we want.

Problem: What if an exception in thrown inside `setUp` or `tearDown`?
Answer: The subsequent setUps and the test won't be executed. TearDowns will be executed. A failure will be present in the native report.


bartekpacia:
The playground in `test/callbacks_test.dart` [has been moved to a gist](https://gist.github.com/bartekpacia/36d09bd103aa298a076916350d735768).

PR #1722, opened 2023-09-19
Bump gradle build action version

fylyppo:


bartekpacia (commented):
Why this change? `@v2` is the same as `@v2.8.0`.

PR #1741, opened 2023-09-22
Lifecycle callbacks setupall investigation

bartekpacia:
Looks like it's not possible to for an instrumentation app to store cache/tmp files ([link](https://stackoverflow.com/questions/69236480/how-to-access-cache-in-android-for-instrumented-tests/70875214#70875214)).

### DIFF
--- a/packages/patrol/example/android/app/src/androidTest/java/pl/leancode/patrol/example/MainActivityTest.java
+++ b/packages/patrol/example/android/app/src/androidTest/java/pl/leancode/patrol/example/MainActivityTest.java
@@ -1,11 +1,15 @@
 package pl.leancode.patrol.example;
 
+import android.util.Log;
 import androidx.test.platform.app.InstrumentationRegistry;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
 import pl.leancode.patrol.PatrolJUnitRunner;
+
+import java.io.File;
+import java.io.IOException;
 
 @RunWith(Parameterized.class)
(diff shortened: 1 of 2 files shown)


bartekpacia:
Remembering `setUpAll`s that were executed across multiple _app under test_ runs requires some safe, persistent place.

On iOS, this can simply be _instrumentation_'s memory, because it lives for the whole duration of the test suite execution.

On Android though, the _instrumentation_ and _app under test_ live in the same process, and both are killed and their memory is cleared after a single test finishes running. It's Android Test Orchestrator who manages execution, and it doesn't expose any APIs for storing some data persistently.

---

Another idea is to store the status of `setUpAll` that were executed in the filesystem.

As an example, we can take a look at how Allure accomplishes that:
- [AllureAndroidJUnitRunner](https://github.com/allure-framework/allure-kotlin/blob/16cc20bd274ed7e19ba22d25758b9e70d2991662/allure-kotlin-android/src/main/kotlin/io/qameta/allure/android/runners/AllureAndroidJUnitRunners.kt#L77)
- [ExternalStoragePermissionsListener](https://github.com/allure-framework/allure-kotlin/blob/16cc20bd274ed7e19ba22d25758b9e70d2991662/allure-kotlin-android/src/main/kotlin/io/qameta/allure/android/listeners/ExternalStoragePermissionsListener.kt)

bartekpacia:
[This sounds good](https://github.com/android/android-test/blob/5c406a14c0393a393f624348ca05562523d78189/services/storage/java/androidx/test/services/storage/BUILD#L1-L2)

bartekpacia:
Superseded by #1751.

PR #1751, opened 2023-09-28
Lifecycle callbacks - support `setUpAll`

bartekpacia:
Targeted at #1503.

See investigation conducted in #1741.


bartekpacia:
Patrol's test bundling feature offloads scheduling&execution of tests to `package:test`. How it works is that we always execute all the tests, but most of them are skipped (i.e. their body is executed but exits early). This behavior implies being able to access `currentTest` from within `testWidgets`, `setUp` and `tearDown` body.

However, `currentTest` is not available inside `setUpAll` and `tearDownAll` body callbacks, so this exact approach won't work.

I think a workaround based on using `parentGroup` should work. See [gist](https://gist.github.com/bartekpacia/56316213fd800929d0e1e9842ec8951c).

---

`currentTest` = `Invoker.current!.liveTest.individualName`
`parentGroup` = `Invoker.current!.liveTest.groups.last.name`

bartekpacia:
When should the setUpAll callbacks be run? The obvious answer is "before its tests start running". This means that Dart side has to know which Dart setUpAll callbacks have already run, and which not.

This is a problem, because currently, the only synchronization point between instrumentation app and app under test is `markPatrolAppServiceReady()` and `listDartTest()`. Both of them happen too late, after all tests have been executed, after all setUpAlls have been called.

bartekpacia:
### Plan

To discover `setUpAll`s, all tests must run. This currently cannot be done during the inital run, because Dart tests only run after `runDartTest()` is called.

To fix that, behavior of Dart tests must differ between the initial run and normal runs. Currently this behavior is mostly the same:

1. App under test starts
2. `patrol_test_explorer` test runs, subsequent tests DO NOT RUN
3. Once `patrol_test_explorer` finishes, instrumentation calls `listDartTests()`
4. [Only in normal run] Once `listDartTests()` finishes, instrumentation calls `runDartTest()`

---

bartekpacia:
Wohoo!

Screenshot 2023-10-13 at 11 51 13 PM


I also ran a few example app tests, they still pass.

bartekpacia:
TODO:
- [ ] iOS implementation
- [ ] `patrol develop`
- [ ] `tearDownAll()`

Issue #1602, opened 2023-08-04
If test fails, it never completes


novas1r1:
### Steps to reproduce

1. Run `flutter build apk --debug` 
2. Write test file where all tests succeed:
```
void main() {
  patrolTest(
    'App is starting and login screen is shown',
    nativeAutomation: true,
    ($) async {
      final app = await AppConfig.initializeApp();

      await $.pumpWidgetAndSettle(app);

      expect($('Benutzername'), findsOneWidget);
      expect($('Passwort'), findsOneWidget);
      expect($('Login'), findsOneWidget);
    },
  );
}
```
3. Run `patrol test --target integration_test\example_test.dart --verbose`
4. Write test with a failing test:
```
void main() {
  patrolTest(
    'App is starting and login screen is shown',
    nativeAutomation: true,
    ($) async {
      final app = await AppConfig.initializeApp();

      await $.pumpWidgetAndSettle(app);

      expect($('NOT FOUND'), findsOneWidget);
      expect($('Passwort'), findsOneWidget);
      expect($('Login'), findsOneWidget);
    },
  );
}
```
5. Run `patrol test --target integration_test\example_test.dart --verbose` 

### Actual results

Hey,
I have multiple apps where I'm trying to work with patrol. Everything works fine, if all my tests are successful. Somehow - if there's one test, or a part of a test failing, the test is stuck forever and never finishes with a fail message. I tried reproducing it, but I can't with a plain project :(. I moved over all configuration files in a new test project, but the error is not showing up. I somehow have a feeling it has to do with dependencies - maybe Firebase? Any hint would be really appreciated, since I'm running out of ideas.

Here's a video running the test. It will stay in the this last line until I force close it:
"Starting 1 tests on Pixel_3a_API_31(AVD) - 12"

https://github.com/leancodepl/patrol/assets/2575205/ca546d84-9fd2-4359-a322-252eaf68a1ec

I do have a "WidgetsFlutterBinding.ensureInitialized();" in my main.dart. I originally had it in my AppConfig.initializeApp() but thought it could be an issue -> where do I need to put it? Should it run in the patrol test or not?

**Device:**
Android Emulator, Pixel 3A, API 31

Any help, hints, whatever would be highly appreciated!
I'm trying my best to get it reproducable somehow...

### Logs

<details>
<summary>Logs</summary>

```
I can't paste the logs because it's too many characters and then I can't open the issue :( 
```

</details>

### Patrol version

patrol: ^2.2.0

[...]

PR #1762, opened 2023-10-02
Add info about not finishing 

tenhobi:
After trying Patrol for a while and not knowing why my tests do not finish when they couldn't find a widget to tap on, etc., I would appreciate a warning in docs for that! 🙏 

Also related to https://github.com/leancodepl/patrol/issues/1602

### DIFF
--- a/docs/getting-started.mdx
+++ b/docs/getting-started.mdx
@@ -291,6 +291,13 @@ up. To run `integration_test/example_test.dart` on a local Android or iOS device
         ```
       </Info>
 
+      <Error>
+        To prevent issues during Patrol tests, please follow these guidelines:
+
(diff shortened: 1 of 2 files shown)


bartekpacia:
Thanks for this PR @tenhobi!

Small thing: I'm not sure `docs/patrol/finders` is the right place for it. That section in the docs talks about Patrol finders, which can be used in both widget test and in integration tests. Not all Patrol users use custom finders, some are only interested in native automation, and then they likely wouldn't see this warning.

I think the better place for it would be [the Getting Started guide](https://patrol.leancode.co/getting-started#create-a-simple-integration-test). What you think?

tenhobi:
@bartekpacia I moved it to the section you suggested, is this better? ☺️

bartekpacia:
Yup, thanks!

PR #1776, opened 2023-10-05
Example app: fix crashing on Android 14

bartekpacia:
Discovered it while trying to reproduce another issue.


bartekpacia:
WIP.

The notification has to be scheduled "at an exact time", otherwise it shows up too late.

bartekpacia:
see #1827

PR #1778, opened 2023-10-06
update outdated docs of the patrol build command

bartekpacia:


bartekpacia:
@fylyppo Do you maybe know when did this issue start occuring?

bartekpacia:
merging because the test is flaky

Issue #1770, opened 2023-10-04
Long press gesture feature


Tamerlanchiques:
### Use case

In my integration test I need to make long press gesture on a button. As I see there is no trickless solution for this.

### Proposal

`await $(PK.timeslotInfoPageSdizTextField).longPress();` makes long press (around 2 seconds) gesture.

Or maybe modify existing tap method:
`await $(PK.timeslotInfoPageSdizTextField).tap(pressDuration: Duration(seconds:2);`

bartekpacia:
Hi 👋🏻 

Looks like `WidgetTester` has the [longPress()](https://api.flutter.dev/flutter/flutter_test/WidgetController/longPress.html) method. We should add it to [`PatrolTester`](https://github.com/leancodepl/patrol/blob/c08337011c4ba02d75173cf400aa31c8ba3ddc5b/packages/patrol_finders/lib/src/custom_finders/patrol_tester.dart) and simply forward.

Would you be willing to contribute this feature? :)

Tamerlanchiques:
Hi. Hm-m… looks like something not that hard so I could try to make my first code-contribute to open source. :)

bartekpacia:
It's [Hacktoberfest](https://hacktoberfest.com/), after all :)

Tamerlanchiques:
Yay, I'm FOSS Contributor now. Exited

_PeepoHappy_

PR #1806, opened 2023-10-15
#1770 | PatrolTester: added longPress gesture

Tamerlanchiques:
leancodepl#1770

### DIFF
--- a/packages/patrol_finders/lib/src/custom_finders/patrol_tester.dart
+++ b/packages/patrol_finders/lib/src/custom_finders/patrol_tester.dart
@@ -294,7 +294,7 @@ class PatrolTester {
   ///
   /// After long press gesture this method automatically calls
   /// [WidgetTester.pumpAndSettle]. If you want to disable this behavior,
-  /// set [andSettle] to false.
+  /// set [settlePolicy] to [SettlePolicy.noSettle].
   ///
   /// See also:
   ///  - [PatrolFinder.waitUntilVisible], which is used to wait for the widget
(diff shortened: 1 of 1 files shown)


packages/patrol_finders/lib/src/custom_finders/patrol_tester.dart
```diff
@@ -275,6 +275,55 @@ class PatrolTester {
...
+  /// long press on:
+  ///
+  /// ```dart
+  /// // long presses on the third TextButton widget
+  /// await $(TextButton).at(2).longPress();
+  /// ```
+  ///
+  /// After long press gesture this method automatically calls
+  /// [WidgetTester.pumpAndSettle]. If you want to disable this behavior,
+  /// set [andSettle] to false.
```
bartekpacia, line 297:
```suggestion
  /// [WidgetTester.pumpAndSettle]. If you want to disable this behavior,
  /// set [settlePolicy] to [SettlePolicy.noSettle].
```

bartekpacia (commented):
Thanks for contributing!

Please also add a simple test [in this file](https://github.com/leancodepl/patrol/blob/555609d24f03df3a1f42797ab53a91b643b5639c/packages/patrol_finders/test/patrol_finder_test.dart). You can take example from other tests in that file.

bartekpacia:
OK, thank you for contribution. On a second thought this is a simple method forward and doesn't require tests so badly.

Tamerlanchiques:
Oh I thought to add tests at weekends but spend time on IRL activities :) 
Will add test coverage soon…

PR #1535, opened 2023-07-13
Enable BrowserStack on iOS

bartekpacia:

PR #1820, opened 2023-10-20
set up example app run on browserstack

bartekpacia:
temporary draft PR for @fylyppo 

see also previous PRs:
- #1635
- #1535 
- #1419


bartekpacia:
this was only for reproduction purposes for @fylyppo 

closing.

PR #1829, opened 2023-10-25
Add Patrol DevTools extension

bartekpacia:
This will bump Patrol's minimum Flutter SDK version to the new stable (3.16, which will likely become available in November).

### DIFF
--- a/.github/workflows/patrol-prepare.yaml
+++ b/.github/workflows/patrol-prepare.yaml
@@ -211,11 +211,11 @@ jobs:
           flutter-version: ${{ matrix.flutter-version }}
           channel: ${{ matrix.flutter-channel }}
 
-      - name: Set up Melos
-        run: dart pub global activate melos
-
-      - name: Activate Melos workspace
-        run: cd ../.. && melos bootstrap
+      - name: Set up Melos and activate workspace
+        working-directory: .
+        run: |
+          dart pub global activate melos
+          melos bootstrap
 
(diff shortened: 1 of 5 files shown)


packages/patrol_cli/lib/src/commands/develop.dart
```diff
@@ -112,6 +112,12 @@ class DevelopCommand extends PatrolCommand {
     final displayLabel = boolArg('label');
     final uninstall = boolArg('uninstall');
 
+    String? iOSInstalledAppsEnvVariable;
+    if (device.targetPlatform == TargetPlatform.iOS) {
+      iOSInstalledAppsEnvVariable =
+          await _iosTestBackend.getInstalledAppsEnvVariable(device.id);
+    }
+
```
bartekpacia, line 120: This will work only on iOS simulator.

zltnDC: ios only needs this :)

bartekpacia: What about physical device?

This invocation: `[xcrun', 'simctl', 'listapps', $deviceId]` works only for the simulator.

bartekpacia: Though I'm not even sure now if `patrol develop` works on iOS, lol.

Anyway thanks for your comment:)

PR #1873, opened 2023-11-07
Update getting-started.mdx

### LINKED ISSUE #1872: Improve Android setup documentation for projects not using .MainActivity
### Use case

Sometimes in the android manifest, instead of using `android:name=".MainActivity"` it will be `android:name="io.flutter.embedding.android.FlutterActivity"` directly.

Because of that, the MainActivityTest won't work because it is trying to find the MainActivity class.

### Proposal

Add more information in the Android setup to say that `instrumentation.setUp(MainActivity.class);` in MainActivityTest can be replaced by `instrumentation.setUp(io.flutter.embedding.android.FlutterActivity.class);` if the project is not using the MainActivity file

SebastienBtr:
Fix #1872


bartekpacia (approved):
LGTM, thanks!

Issue #1369, opened 2023-06-19
Remove deprecated `andSettle` argument


jBorkowska:

bartekpacia:
fixed in #1892

PR #1892, opened 2023-11-13
Remove deprecated parameters from `patrolTest`

### LINKED ISSUE #1624: Consequences of PatrolFinders
We've extracted PatrolTester into a new package and we don't want to release another major version of the patrol package, so we need to avoid breaking changes. This issue collects small things that we may change in the future.

1. Delete typedef PatrolTester
2. patrolTest will always initialize NativeAutomator, so we can replace these lines of code in PatrolIntegrationTester:
```
final NativeAutomator? nativeAutomator;
NativeAutomator get native;
```
to
```
final NativeAutomator native;
```
or
```
final NativeAutomator nativeAutomator;
NativeAutomator get native => nativeAutomator;
```
3. Remove the nativeAutomation argument from the patrolTest() function

jBorkowska:
Closes #1624 and #1369

### DIFF
--- a/packages/patrol/example/integration_test/common.dart
+++ b/packages/patrol/example/integration_test/common.dart
@@ -10,13 +10,13 @@ final _nativeAutomatorConfig = NativeAutomatorConfig(
   findTimeout: Duration(seconds: 20), // 10 seconds is too short for some CIs
 );
 
-Future<void> createApp(PatrolTester $) async {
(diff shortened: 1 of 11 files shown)


mateuszwojtczak (approved):
LGTM

bartekpacia (approved):
LGTM as well.

For the record: the 3 failing `patrol_finders` tests will be fixed in a separate PR.

PR #1901, opened 2023-11-14
Fix `.gitignore` ignoring bundled DevTools extension code

bartekpacia:


neiljaywarner:
@bartekpacia  I still do not see build dir in master where i would expect https://github.com/leancodepl/patrol/tree/master/packages/patrol/extension/devtools

Issue #1654, opened 2023-08-21
Add a `exists` function to `NativeAutomator`


MarcinusX:
When writing tests that open a browser, on Firebase Test Lab there is a cookie banner but when I run them on my local device, the banner is already closed. This introduces two paths of execution in a native view.

What I'd like to do is to check if there is a text in the browser, and if there is, click it. Something like:
```dart
final cookieSelector = Selector(text: 'Accept cookies');
if ($.native.exists(cookieSelector)) {
  $.native.tap(cookieSelector);
}
```
Or, maybe allow `native.tap` to not find given selector:
```dart
final cookieSelector = Selector(text: 'Accept cookies');
$.native.tap(cookieSelector, throwIfNotFound: false);
```

Currently my work around is an extension on `PatrolIntegrationTester`:
```dart
  Future<void> nativeTryTapText(String text) async {
    try {
      await native.tap(Selector(text: text));
    } on PatrolActionException catch (e) {
      if (e.message.contains('NOT_FOUND')) {
        return;
      }
      rethrow;
    }
  }
  ```

Not sure if it's technically possible to check if there is a match on a `Selector` before trying to tap it but if so, such functions could be helpful. :)

Kendru98:
related to: https://github.com/leancodepl/patrol/issues/2280

Issue #1465, opened 2023-07-05
Impossible to check if permission dialog is visible when using non-US locale


wheeOs:
### Steps to reproduce

1. run the sample test below in an iOS simulator
2. the permission dialog shows up in the test
3. wait until the test finishes

```dart
import 'package:flutter/material.dart';
import 'package:geolocator/geolocator.dart';
import 'package:patrol/patrol.dart';

void main() {

  patrolTest("sample test", nativeAutomation: true, ($) async {
    
    await $.pumpWidgetAndSettle(MaterialApp(
      builder: (_, __) {
        Geolocator.requestPermission();
        return const Center(child: Text("Test"));
      },
    ));

    await $.pump(const Duration(seconds: 4));
    await $.native.grantPermissionWhenInUse();
    //NOTE: (await $.native.isPermissionDialogVisible()) will return false here!;

  });
}
```

### Actual results

On iOS the permission dialog seems not to be interactable for patrol. When the dialog is visible, neither `native.isPermissionDialogVisible` returns `true` nor can permissions be granted via `grantPermission..()`.

the following error occurs:

```
        Patrol (native): grantPermissionWhenInUse() failed
        ══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
        The following PatrolActionException was thrown running a test:
        Patrol action failed: GrpcError: grantPermissionWhenInUse() failed with code NOT_FOUND (button to
        allow permission only once doesn't exist)
        When the exception was thrown, this was the stack:
        #0      NativeAutomator._wrapRequest (package:patrol/src/native/native_automator.dart:200:7)
        <asynchronous suspension>
        #1      NativeAutomator.grantPermissionWhenInUse (package:patrol/src/native/native_automator.dart:659:5)
        <asynchronous suspension>

```

### Logs

<details>
<summary>Logs</summary>

```
Verbose mode enabled. More logs will be printed.
Received 1 test target(s)
Received test target: /Users/wheeOS/Development/Projects/wheeos_mobile_app/integration_test/customer_journeys/dev_test.dart
Generated entrypoint /Users/wheeOS/Development/Projects/wheeos_mobile_app/integration_test/test_bundle.dart with 1 bundled test(s)
$ flutter --no-version-check devices --machine
No device specified, using the first one (iPhone 13 Pro)
Received 1 device(s) to run on
Received device: iPhone 13 Pro
[...]

PR #1919, opened 2023-11-17
Fix/example app android 14

### LINKED ISSUE #1827: Example app's `notification_test` is crashing on Android 14
Android 14 made it harder to schedule notifications at exact time ([see docs](https://developer.android.com/about/versions/14/behavior-changes-all)). Because of that, the example app is crashing:

Screenshot 2023-10-24 at 12 33 02 PM
on Android 14.

- flutter_local_notifications has to be updated
- notifications_test should be updated as well

See PR #1776 where I started work but didn't have time to finish it.

piotruela:
Closes #1827


piotruela:
I could not find a way to check if dialog (which shows only on Android 14) showed up, so I implemented it so that the test tries to interact with permission page but if it's not present it continues to execute. 
Would be useful for this test to have a method like the one described [here].(https://github.com/leancodepl/patrol/issues/1465#issuecomment-1628730209)
`$.native.isNativeViewVisible(Selector)`

bartekpacia:
@piotruela See this https://github.com/leancodepl/patrol/issues/1654#issue-1858734974 for a workaround to check if a native UI element is visible.

.github/workflows/test-android-14-device.yaml
bartekpacia, line 1: Do not create a new workflow, we have too many of them already. Instead add a new `device_model and `os_version` to existing `test-android-device.yaml`.

bartekpacia (changes requested):

packages/patrol/example/integration_test/permissions/notifications_test.dart
```diff
@@ -11,6 +11,16 @@ void main() {
         await $.native.grantPermissionWhenInUse();
       }
 
+      // Android 14+ requires additional permission to schedule notifications
+      try {
+        await $.native.tap(
+          Selector(text: 'Allow setting alarms and reminders'),
+        );
```
bartekpacia, line 18: Doesn't the default `grantPermission` work in this case? Because it should and if it doesn't it's very likely a bug (that warrants a new issue).

piotruela: It doesn't work, but it's not a dialog, it opens the settings page. Should `grantPermissionWhileInUse` also work in this case?
Screenshot_1700218874

bartekpacia: Ah, okay. It's okay to leave it as it is.

.github/workflows/test-android-device.yaml
```diff
@@ -21,9 +21,13 @@ jobs:
       matrix:
         flutter-version: ['3.16.x']
         flutter-channel: ['stable']
-        device_model: ['oriole']
+        device_model: ['oriole', 'shiba']
```
bartekpacia, line 24: I think you do not need `device_model` and `os` in `matrix` now. Having them in `matrix.include` is enough.

Not sure though.

piotruela: I guess you're right. Afaik it will be included in the config anyway.

bartekpacia (approved):

packages/patrol/example/android/app/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
bartekpacia: I'm not against having consistent formatting of AndroidManifest.xml files but we should then use some XML formatter and enforce it on the CI.

packages/patrol/example/android/app/src/main/AndroidManifest.xml
```diff
@@ -35,10 +36,21 @@
...
+        </intent-filter>
+    </receiver>
         <!-- Don't delete the meta-data below.
              This is used by the Flutter tool to generate GeneratedPluginRegistrant.java -->
         <meta-data
             android:name="flutterEmbedding"
             android:value="2"/>
     </application>
-</manifest>
+</manifest>
```
bartekpacia, line 56: why this?

piotruela: I overlooked it. Fixing now

PR #1953, opened 2023-11-23
Add Discord link

jBorkowska:


bartekpacia (approved):
I think it shouldn't be at the top, but whatever I guess.

jBorkowska:
I thought it would be under twitter, but I don't know if this is possible, we could do other anchors as real anchors and maybe then it would be sorted as we want

PR #1984, opened 2023-12-05
Feature/gh actions macos test

piotruela:


piotruela:
[It works](https://github.com/leancodepl/patrol/actions/runs/7100681241) but without screen recording. I created an tech-debt [issue](https://github.com/leancodepl/patrol/issues/1986) for that.

PR #1988, opened 2023-12-05
Fix android and iOS scripts links

Rahiche:
Current links are no valid


bartekpacia (approved):
Hi Raouf 👋🏻 💙 😄

bartekpacia:
Thanks for fixing that!

Issue #1868, opened 2023-11-07
On real iOS device enterText() doesn't work when TextField is focused


aqrc:
### Steps to reproduce

I experienced the same problem described in https://github.com/leancodepl/patrol/issues/1483.
This is not an issue on simulator.

<details>
  <summary>Minimal reproducible example</summary>

```dart
import 'package:flutter/material.dart';
import 'package:flutter_test/flutter_test.dart';
import 'package:patrol/patrol.dart';

void main() {
  patrolTest(
    'should enter text into focused field but doesnt',
    nativeAutomation: true,
    ($) async {
      await $.pumpWidgetAndSettle(
        MaterialApp(
          home: Scaffold(
            appBar: AppBar(title: const Text('app')),
            body: const Column(
              children: [
                TextField(
                  key: Key('autofocused_field'),
                  autofocus: true,
                ),
                TextField(
                  key: Key('not_focused_field'),
                ),
              ],
            ),
          ),
        ),
      );

      const testText = 'text for test';
      await $(#autofocused_field).enterText(testText); // doesn't work, fails later
      // await $(#not_focused_field).enterText(testText); // works fine
      await Future.delayed(const Duration(seconds: 4));
      expect($(testText), findsOneWidget);
    },
  );
}

```

</details>

### Actual results

On real iOS device, when `TextField` is focused, text is not getting entered into this `TextField` using `.enterText()`.  
Field might be focused due to `autofocus` (`TextField(autofocus: true)`), or due to previous call of `.tap()` method (like in the referenced issue).  

### Logs

<details>
<summary>Logs</summary>

```
andrey@Andreys-MacBook-Pro-276 frontend % patrol test --target integration_test/patrol_issue_test.dart --release --verbose
In next major release, patrolTest method will be intended for UI tests.
If you want to use Patrol in your widget tests, use patrol_finders package.

For more information, see https://patrol.leancode.co/patrol-finders-release
Disable this warning by setting the PATROL_FINDERS environment variable.
      
Verbose mode enabled. More logs will be printed.

Update available! 2.1.5 → 2.2.2
Run patrol update to update

Received 1 test target(s)
Received test target: /Users/andrey/companyname/frontend/integration_test/patrol_issue_test.dart
Generated entrypoint /Users/andrey/companyname/frontend/integration_test/test_bundle.dart with 1 bundled test(s)
Received Android flavor: dev
Received iOS flavor: dev
$ flutter --no-version-check devices --machine
[...]

PR #2030, opened 2023-12-22
Make more `Selector` properties work on iOS

### LINKED ISSUE #1554: Disparity of `Selector` behavior between Android and iOS
Patrol has the [`Selector`](https://github.com/leancodepl/patrol/blob/d8e462d2f4418f9558f8bab7595e460f50042160/contracts.proto#L178-L191) and [`NativeView`](https://github.com/leancodepl/patrol/blob/d8e462d2f4418f9558f8bab7595e460f50042160/contracts.proto#L196-L206) classes. They provide cross-platform interaction with the underlying native UI.

Currently, there are many differences `Selector` and `NativeView` between its behavior on Android and iOS.

### More about these 2 classes

<details><summary>Selector</summary>

On Android, `Selector`s are translated to [`BySelector`](https://developer.android.com/reference/androidx/test/uiautomator/BySelector)s and/or [`UiSelector`](https://developer.android.com/reference/androidx/test/uiautomator/UiSelector).

On iOS, `Selector`s are translated to [`XCUIElementQuery`](https://developer.apple.com/documentation/xctest/xcuielementquery)-ies. Sometimes, we construct them using `NSSelector`. They allow for searching using many view properties: `identifier`, `title`, `label`, `value`, and `placeholderValue`.

</details> 

<details><summary>NativeView</summary>

On Android, `NativeView`s are created from [`UiObject2`](https://developer.android.com/reference/androidx/test/uiautomator/UiObject2)s.

On iOS, `NativeView`s are created from [`XCUIElement`](https://developer.apple.com/documentation/xctest/xcuielement)s.

</details> 

### Problems

- **on iOS, only `Selector.text` is taken into account** when searching for views
- some `Selector` fields, such as [`resourceId`](https://github.com/leancodepl/patrol/blob/d8e462d2f4418f9558f8bab7595e460f50042160/contracts.proto#L186), are Android-only. That's because the `Selector` API predates iOS support.
- some `NativeView` fields, such as [`resourceName`](https://github.com/leancodepl/patrol/blob/d8e462d2f4418f9558f8bab7595e460f50042160/contracts.proto#L203), are Android-only. Cause the same as above.
- #1403
- #663

bartekpacia:
This PR resolves #1554.


bartekpacia:
Some thoughts for when I'm gone:

- I really think that `Selector.text` and `Selector.contentDescription` should be the same thing (preferably named `Selector.label`) @piotruela

piotruela:
I made Selector's `textStartsWith` and `textContains` work with methods `tap`, `doubleTap`, `waitUntilVisible`, `enterText` and `getNativeViews`. I wrote some tests in `ios_selector_test.dart` but I spotted a weird behavior there. If I run just a single test it passes but when I try to run more than one at once they all fail. I could not find a cause of that but before we merge this PR we need to investigate this.

jBorkowska:
I investigated failing tests, it is caused by another bug: https://github.com/leancodepl/patrol/issues/1868, I suggest to merge this PR regardless to those tests failing

abhiche:
When do you plan to release this?

packages/patrol/darwin/Classes/AutomatorServer/Automator/Automator.swift
```diff
@@ -1,3 +1,93 @@
+extension Selector {
```
zltnDC, line 1: This extension should be rather below #if PATROL_ENABLED.

piotruela (approved):

zltnDC (approved):

Issue #492, opened 2022-10-21
Add support for test tags to the CLI


bartekpacia:
The built-in [test](https://pub.dev/packages/test) package lets developers add _tags_ to their test files, groups, and individual tests.

I won't ramble about tags here, instead I'll send you to the official docs:
- [Tagging tests](https://pub.dev/packages/test#tagging-tests) in `package:test`'s readme
- [Configuring tags](https://github.com/dart-lang/test/blob/master/pkgs/test/doc/configuration.md#configuring-tags) in `package:test`'s docs

We should research:
- if tags are good for marking platform-specific tests
- if there are any problems with them. If not, we should add support for running tests by tags to `patrol_cli`, and update docs to let our users know about this feature.

cc: @jakubfijalkowski @mateuszwojtczak @jBorkowska, wdyt?

bartekpacia:
`flutter drive` (which we currently use) doesn't support test tags. We'll get them for free once we migrate to `flutter test`. This migration is tracked in #421.

jBorkowska:
Related to this PR: https://github.com/leancodepl/patrol/pull/2052

PR #2052, opened 2024-01-09
Add support for running tagged test files

kostadin-damyanov-prime:
Hello.

This PR adds support for `--tags` option for the `test` command.
This way we can tag test files with the `@Tags` annotation like dart's test suite does.
Using this option we can now run only the test files matching a given tag or a list of tags.

### DIFF
--- a/packages/patrol_cli/lib/src/commands/test.dart
+++ b/packages/patrol_cli/lib/src/commands/test.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:patrol_cli/src/analytics/analytics.dart';
 import 'package:patrol_cli/src/android/android_test_backend.dart';
+import 'package:patrol_cli/src/base/exceptions.dart';
 import 'package:patrol_cli/src/base/extensions/core.dart';
 import 'package:patrol_cli/src/base/logger.dart';
 import 'package:patrol_cli/src/compatibility_checker.dart';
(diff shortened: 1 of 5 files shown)


mateuszwojtczak:
Hi @kostadin-damyanov-prime, thanks a lot for making this contribution! We have just discussed tags internally :)

I have a few questions:
1. Should we lock the `analyzer` package version to some more specific than `any`?
2. When I initially considered tags, I was thinking about somehow connecting to the official `test_api` way of doing that. I can see you implemented a separate mechanism for that. WDYT?

kostadin-damyanov-prime:
@mateuszwojtczak Hi.

1. Done.
2. I am not too familiar with test_api. Do they use a different approach? If there is some code that we can reuse I am all for it.

mateuszwojtczak:
Yes, look at this [information in their readme](https://github.com/dart-lang/test/blob/master/pkgs/test/README.md#tagging-tests).

They also support `dart_test.yaml` file. Your implementation is really fine with most cases and it's simple which is very beneficial. 

I'm just worried that the community would assume that Patrol uses the same tags as Dart/Flutter tests. This is a very frequent assumption that ends with people filing issues, because they assume Patrol uses the same approach. Although we can't always be 100% compatible with that approach, we still at least try to do so.

kostadin-damyanov-prime:
I agree that mplementation based on `test_api` will be more feature-rich. Maybe even supporting stuff like `--exclude-tags` and boolean selectors. Unfortunately we do not have the time to implement it now. You can either merge this PR or do implementation based on `test_api` yourself. We (and anyone else) can use this fork in the meantime.

mateuszwojtczak:
@kostadin-damyanov-prime We decided to implement tags and skips by passing them as arguments to the `patrolTest` function. This means we are no longer in need of annotations / other analyzer pre-processing work.

Thus, I'm closing this PR and we're gonna solve #492 with another approach :)

PR #2055, opened 2024-01-12
Update getting-started.mdx FAQ related to use with riverpod

StevePhuc:


mateuszwojtczak:
Hi @StevePhuc! Thanks for your contribution.

When it comes to this workaround - it's a bit of a hack and we would like to not encourage users to wait for an arbitrary duration just because it happens to solve the issue.

However, we want Riverpod users to be able to integrate Patrol into their apps. So we need to check if:
1. Patrol works with vanilla Riverpod app
2. Find a solution to this problem that is less hacky.

PR #2072, opened 2024-01-26
Fix `enterText` not being able to interact with other than first textfield on Android

jBorkowska:
Story time:
The bug occured when there were multiple native textFields on the screen and all of them had the same direct parent. The process of finding the textfield was taking into account that dev might not use a selector to the textfield, but to its parent (eg. when textfield is wrapped with another element and only the wrapping element has resourceId, not textfield). So after finding the element by provided selector, `enterText` took element's parent, and looked for the first textfield among children of this parent. It resulted in the first textfield to be picked always, even if selector argument in `enterText` found another one (correct from the perspective of the dev).


piotruela:
Please add this fix to changelog

Issue #1982, opened 2023-12-04
`patrol run` to quickly start tests that are already present on the device


bartekpacia:
### Use case

Currently the feedback loop when developing integrations tests is very slow because of the need to make a full rebuild after every change. I very often cannot use `patrol develop` because of its [severe limitations](https://patrol.leancode.co/cli-commands/develop#caveats).

I want to be able to **quickly** (i.e. without rebuilding) run the app that is **already installed** on the device.

### Proposal

If:
1. the app is already installed on the device, and
2. my `$PWD` is Flutter project root,

then I want to be able to:

```
patrol run
```

and it will run the tests for my app.

I also want to be able to pass:

```
patrol run --only-run 'example_test mainGroup mainFlowTest'
```

to only run the `mainFlowTest` which is in group `mainGroup` in `integration_test/example_test.dart` file (this is somewhat similar to #1713)

(similar to `xcodebuild`'s `-only-testing` flag on iOS).

Clemens-Toegel:
are there any updates on this?
Would the open PR look good in general? so that it could be merged if the checks would pass?
For our CI this would be a huge benefit.

jBorkowska:
Size: M (needs research)

piotruela:
Added in #3197.
See more about how to use it in docs: 
https://patrol.leancode.co/documentation/ci/build-time-test-discovery#running-tests-without-rebuilding

PR #2080, opened 2024-01-30
Add test-only command to bypass build.

scottandrewzip:
Great utility for CI. Add test-only for when you need to separate build and test steps. Our test step was taking 3 minutes after we would do a build. This was due to the shear size of our project.

### DIFF
--- a/packages/patrol_cli/README.md
+++ b/packages/patrol_cli/README.md
@@ -12,6 +12,8 @@ Command-line tool to make working with [patrol][pub_link_test] easier.
 ```console
 $ dart pub global activate patrol_cli
 ```
+### From our repo..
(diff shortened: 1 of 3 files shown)


bartekpacia:
see also https://github.com/leancodepl/patrol/issues/1982 which where I proposed this feature

piotruela:
Thanks @scottandrewzip for the input.
I checked how it works and looks like the behavior of this command varies between platforms. On Android, the `_execute` method invokes the `./gradlew connectedAndroidTest` task, which rebuilds the test by updating it. On iOS, however, the `xcodebuild --test-without-building` command is used, which utilizes already built instrumentation so even if we change something in the test code we won't see it.
We need to find a way to solve the disparity by making it work the same way on both platforms

pdenert:
We'll do another approach

Issue #2110, opened 2024-02-15
`enterText()` not working, when interacting with the same textfield twice


jBorkowska:
Eg. we have a textfield and we enter in it 'text1', then we go back to previous page, and then again to the screen with the textfield. When trying to enter text again, the text stays the same. Though the textfield gets focused, keyboard is shown, but text stays the same.
When using `patrol develop` the effect is the same - you can input a text only once in each textfield during the whole run of this command.

Spotted on Android, didn't check iOS

jBorkowska:
Related to https://github.com/flutter/flutter/issues/134604

PR #2111, opened 2024-02-15
Add registering text input

jBorkowska:
Temporary fix for https://github.com/leancodepl/patrol/issues/2110

### DIFF
--- a/packages/patrol_finders/lib/src/custom_finders/patrol_tester.dart
+++ b/packages/patrol_finders/lib/src/custom_finders/patrol_tester.dart
@@ -334,9 +334,12 @@ class PatrolTester {
     Duration? visibleTimeout,
     Duration? settleTimeout,
   }) {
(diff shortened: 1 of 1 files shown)


zltnDC (approved):

b055man:
@jBorkowska that change seems to singlehandledly break the behavior of the text input field in our app tests (run in the Android emulator) - downgrading to 2.0.1+1 resolves issues
1) the keyboard does not show up upon calling enterText if the field is not tapped first (less of a problem, though awkward)
2) the keyboard does not close when the text field edit is done / the focus is lost - it stays on forever (can't really close it even when tapping back in the emulator)

It doesn't look at first glance that there's something incorrect with the app, esp. as it behaves as expected during regular operations. Also, these tests worked fine with previous Patrol (2.3.2) and as mentioned they work fine if 2.0.1+1 version of the package is used.

PR #2120, opened 2024-02-20
Resolves an issue where test bundler would generate invalid code if the file path included dashes

kodypeterson:
Identified an issue where having dashes (-) in your file path would result in invalid dart code being generated.


piotruela:
Thanks for the PR. We recently opened a topic of the separators and wanted to change them into something which would work universally and I like this idea. But before we will go with it we need to make sure that it will not break running tests on device farms (some of them parses the output of the console and expect it to be in specific form).

mateuszwojtczak:
Hi @kodypeterson! 
Could you create the issue with a reproducible sample with this one? 
We would like to investigate it, but I'm not sure if the PR is the right place for it, especially that we might want to choose another approach to that issue.

Issue #1482, opened 2023-07-10
Enable using patrol_cli with FVM


lyio:
### Steps to reproduce

1. Use `fvm` in your project
2. Do not have a global Flutter version installed
3. Install `patrol_cli`
4. Try to run `patrol test` or `patrol drive`

### Actual results

`patrol_cli` attempts to invoke `flutter` which is not available (in our case). We use `fvm` in all our projects and manage our Flutter versions like that.

It would be nice if there was a way to run Patrol in this environment without having to manage a global Flutter version.

### Logs

<details>
<summary>Logs</summary>

```
<!-- Replace this line with your logs. Do not remove the backticks! -->
```

</details>

### Patrol version

partrol: any

### Patrol Doctor output

<details>
<summary>Patrol Doctor output</summary>

```
<!-- Replace this line with your logs. Do not remove the backticks! -->
```

</details>

### Flutter Doctor output

<details>
<summary>Flutter Doctor output</summary>

```
<!-- Replace this line with your logs. Do not remove the backticks! -->
```

</details>

bartekpacia:
Hi @lyio :)

This issue is a valid one.

The first solution that came to my mind is to have an env var `PATROL_USE_FVM`. If it is set, `patrol` will be doing `fvm flutter` and `fvm dart` internally.

Alternatively: we could automatically fallback to `fvm` if no global Flutter version is found. But I don't like such implicit behavior.

Maybe you have any other ideas?

lyio:
While writing the issue I thought, checking for `fvm` might be a good way to handle this. Basically, your second suggestion the other way around.

I personally do not like setting a bunch of env variables, but we could combine the variable with a cli flag maybe? 
That way you could set the `PATROL_USE_FVM` or `patrol test --fvm`.

bartekpacia:
Yeah, env var + cli flag is OK with me.

I'll try to push this issue forward on sprint planning ;P

Jjagg:
@bartekpaci What about a `PATROL_FLUTTER_COMMAND` env variable instead? It seems more flexible and does not tie the flag and implementation to FVM specifically. Users can specify `PATROL_FLUTTER_COMMAND=fvm flutter` to use fvm or they could specify a full path to a flutter installation.

vlazdra:
Hey @bartekpacia has this feature request made it to sprint planing? 😁

mrocha98:
[...]

PR #2124, opened 2024-02-22
Flutter command arg

mvarendorff:
Aims to address #1482 

This PR builds on top of the [amazing work by Jjagg](https://github.com/leancodepl/patrol/issues/1482#issuecomment-1956208862). It introduces a new CLI flag `--flutter-command` which may be used to provide a custom command used when running the Flutter CLI to support version managers like puro and fvm. It also supports a matching env-Variable `PATROL_FLUTTER_COMMAND`.

I believe all occurences where `flutter` was previously used directly in the `patrol_cli` package are now replaced with usage of the `FlutterCommand` instance. If I missed any, let me know!

### DIFF
--- a/packages/patrol_cli/lib/src/runner/patrol_command.dart
+++ b/packages/patrol_cli/lib/src/runner/patrol_command.dart
@@ -1,6 +1,7 @@
 import 'package:args/command_runner.dart';
 import 'package:patrol_cli/src/base/exceptions.dart';
 import 'package:patrol_cli/src/ios/ios_test_backend.dart';
+import 'package:patrol_cli/src/runner/flutter_command.dart';
 
 abstract class PatrolCommand extends Command<int> {
   /// Seconds to wait after the individual test case finishes executing.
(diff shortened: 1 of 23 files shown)


packages/patrol_cli/lib/src/runner/patrol_command.dart
```diff
@@ -181,6 +182,21 @@ abstract class PatrolCommand extends Command<int> {
     return argResults![name]! as List<String>? ?? <String>[];
   }
 
+  FlutterCommand get flutterCommand {
+    final arg = globalResults!['flutter-command'] as String?;
+
+    var cmd = arg;
+    if (cmd == null || cmd.isEmpty) {
```
piotruela, line 189: Have you tested the scenario with defining flutter command as environment variable? In `patrol_command_runner.dart` `flutter-command` defaults to `flutter`. I think it will cause that this condition will always be false.

piotruela (commented):
Great job. Just please make sure that defining the flutter command as env variable also works

packages/patrol_cli/lib/src/runner/patrol_command.dart
mvarendorff: Good catch, sorry about that! That's fixed now:

grafik

mvarendorff:
I also fixed an issue related to the changes to the doctor command printing the version; because of a missing `runInShell: true` that previously caused those to fail / require the `.bat` extension on Windows.

Additional note: While testing running with `fvm flutter` as the command, I experienced issues with FVM crashing while running `patrol doctor`, causing the output "Invalid Flutter command: fvm flutter". This happened with FVM@2.4.1, and does *not* happen anymore with FVM 3.0.14 (I have not tested the version-boundaries of the bug). So in case that's an issue that comes up in your tracker, have them try updating FVM using `dart pub global activate fvm` 🤞

piotruela (approved):
LGTM 👌

PR #2193, opened 2024-05-13
Bump dependencies to match Flutter 3.22

piotruela:
Upcoming version of Flutter depends on `test_api` 0.7.0. It also makes `duration` and `phase` parameters in `pumpWidget` method named so we have to adjust or implementation of `patrol_finders` and bump the dependencies.

### DIFF
--- a/packages/patrol/CHANGELOG.md
+++ b/packages/patrol/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 3.7.0-dev.2
 
 - Bump `test_api` dependency to 0.7.0.
+- Bump `patrol_devtools_extension` dependencies. 
 
 ## 3.6.1
 
--- a/packages/patrol/CHANGELOG.md
+++ b/packages/patrol/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.7.0-dev.2
+
+- Bump `test_api` dependency to 0.7.0.
+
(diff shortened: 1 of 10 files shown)


zltnDC (approved):

rydmike:
@piotruela, 

Will you release the 2.0.3-dev.1 version as 2.0.3 soonish with the Flutter 3.22 pumpWidget support? Got some tests not running on Flutter 3.22 due to this. Thanks in advance.

rishabnayak (changes requested):
appears you might need to bump flutter to 3.22.0 [here](https://github.com/leancodepl/patrol/blob/chore/deps/.cirrus.yml)

piotruela:
> @piotruela,
> 
> Will you release the 2.0.3-dev.1 version as 2.0.3 soonish with the Flutter 3.22 pumpWidget support? Got some tests not running on Flutter 3.22 due to this. Thanks in advance.

@rydmike
Do you mean that the 2.0.3-dev.1 version doesn't work for you? What exactly is the problem there?

rydmike:
It certainly works and I used it, but I would prefer to not depend on a .dev release, especially not in another package. However, since it is only a dev/test dependency I did. :)

But the released none dev version of your package now no longer works at all with the latest stable Flutter 3.22 release, so it should be bumped.

Issue #1930, opened 2023-11-20
Remove workaround for `debugDefaultTargetPlatformOverride` being changed during test by DevTools


bartekpacia:
We were forced to add a workaround because of a bug in DevTols. See https://github.com/flutter/devtools/issues/6719

Long story short: the DevTools change a global Flutter value `debugDefaultTargetPlatformOverride`, but don't reset it at the end of the test. This results in assertions being broken in `flutter_test` binding.

There are 3 parts to the workaround:

https://github.com/leancodepl/patrol/blob/f0cbe876911115f74ac3c4044b3345a1bf94a436/packages/patrol/lib/src/binding.dart#L149-L153

https://github.com/leancodepl/patrol/blob/f0cbe876911115f74ac3c4044b3345a1bf94a436/packages/patrol/lib/src/common.dart#L116-L117

https://github.com/leancodepl/patrol/blob/f0cbe876911115f74ac3c4044b3345a1bf94a436/packages/patrol_cli/lib/src/test_bundler.dart#L142-L144

### What does success look like

1. Flutter team acknowledges the issue
2. The fix is merged in Flutter
3. We can remove the workaround code from Patrol

bartekpacia:
Adding `blocked` since it doesn't depend entirely on us.

daohoangson:
I'm getting `Cannot modify debugDefaultTargetPlatformOverride in non-debug builds.` when testing with Firebase Test Labs. 

```
May 20 09:36:28 iPhone Runner(Flutter)[534] <Notice>: flutter: Cannot modify debugDefaultTargetPlatformOverride in non-debug builds.
May 20 09:36:28 iPhone Runner(Flutter)[534] <Notice>: flutter: #0      debugDefaultTargetPlatformOverride= (package:flutter/src/foundation/platform.dart:106)
May 20 09:36:28 iPhone Runner(Flutter)[534] <Notice>: flutter: #1      patrolTest.<anonymous closure> (package:patrol/src/common.dart:140)
May 20 09:36:28 iPhone Runner(Flutter)[534] <Notice>: flutter: <asynchronous suspension>
May 20 09:36:28 iPhone Runner(Flutter)[534] <Notice>: flutter: #2      testWidgets.<anonymous closure>.<anonymous closure> (package:flutter_test/src/widget_tester.dart:183)
May 20 09:36:28 iPhone Runner(Flutter)[534] <Notice>: flutter: <asynchronous suspension>
[...]

PR #2203, opened 2024-05-22
Compare our value before setting `debugDefaultTargetPlatformOverride`

daohoangson:
This PR is related to #1930.

Since flutter/flutter@7ca4b7b, Flutter prevents changes to `debugDefaultTargetPlatformOverride` in non-debug mode so the test will fail when being run on a device farm like Firebase Test Lab. This PR compares the value before changing it, the logic should be the same as before:

- In debug mode, the value is reset as needed
- In production mode, the value should not be changed anyway so we will skip the setter and avoid the thrown error


piotruela (approved):
LGTM, thanks for the contribution!

PR #2219, opened 2024-06-11
Add openUrl method to native automator

### LINKED ISSUE #556: Testing deep linking
Hi, 

is there a way to test deep links with patrol? I've only found the open App method, but I want to specify a URL and check if my app opens and navigates to the correct page.

e.g.: `await $.native.openLink(link: 'http://www.example.com/passwordrest');`

pdenert:
Close #556


packages/patrol/lib/src/native/native_automator2.dart
```diff
@@ -240,6 +240,14 @@ class NativeAutomator2 {
     );
   }
 
+  /// Opens the URL specified by [url].
+  Future<void> openUrl({required String url}) async {
```
piotruela, line 244: Imo `url` could be a positional parameter here. It's quite clear that this method takes a url

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
```diff
@@ -53,6 +57,18 @@
       }
     }
 
+    func openUrl(_ urlString: String) {
+      guard let url = URL(string: urlString) else {
+        Logger.shared.i("Invalid URL string: \(urlString)")
+
+        return
```
piotruela, line 64: Maybe throw a `PatrolError` here

packages/patrol_devtools_extension/lib/api/contracts.dart
piotruela, line 1: I know it's generated from `schema.dart` but I'm wondering if this is a necessary change. As far as I see, we don't use this method in the devtools extension

pdenert: I can remove it, but it'll be back with each subsequent generation of contracts

piotruela (approved):
👍

PR #2223, opened 2024-06-12
Add press volume up and down actions

### LINKED ISSUE #2220: Add `volumeUp` and `volumeDown` native capabilities
### Use case

I have test cases where I need to simulate pressing physical volume up and down buttons

### Proposal

Android docs:
- https://developer.android.com/reference/androidx/test/uiautomator/UiDevice#pressKeyCodes(int[])

iOS docs:
- https://developer.apple.com/documentation/xctest/xcuidevice/button/volumeup
- https://developer.apple.com/documentation/xctest/xcuidevice/button/volumedown

pdenert:
Close #2220

### DIFF
--- a/packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
+++ b/packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
@@ -362,14 +362,14 @@
 
     // MARK: Volume settings
     func pressVolumeUp() throws {
-      runAction("pressing volume up") {
-        self.device.press(XCUIDevice.Button.pressVolumeUp)
+      try runVolumeAction("pressing volume up") {
+        self.device.press(XCUIDevice.Button.volumeUp)
       }
     }
 
     func pressVolumeDown() throws {
-      runAction("pressing volume down") {
-        self.device.press(XCUIDevice.Button.pressVolumeDown)
+      try runVolumeAction("pressing volume down") {
+        self.device.press(XCUIDevice.Button.volumeDown)
       }
     }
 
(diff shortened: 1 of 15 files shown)


packages/patrol/lib/src/native/native_automator2.dart
```diff
@@ -336,6 +336,16 @@ class NativeAutomator2 {
...
 
+  /// Press volume up
+  Future<void> pressVolumeUp() async {
+    await _wrapRequest('pressVolumeUp', _client.pressVolumeUp);
+  }
+
+  /// Press volume down
+  Future<void> pressVolumeDown() async {
+    await _wrapRequest('pressVolumeDown', _client.pressVolumeDown);
+  }
```
piotruela, line 347: Does those methods work on iOS Simulator and Android emulator? If no, please describe the limitation here

packages/patrol/lib/src/native/native_automator.dart
```diff
@@ -452,6 +452,16 @@ class NativeAutomator {
...
+  /// Press volume up
+  Future<void> pressVolumeUp() async {
+    await _wrapRequest('pressVolumeUp', _client.pressVolumeUp);
+  }
+
+  /// Press volume down
+  Future<void> pressVolumeDown() async {
+    await _wrapRequest('pressVolumeDown', _client.pressVolumeDown);
+  }
+
```
piotruela, line 464: ditto

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
```diff
@@ -346,6 +346,19 @@
       }
     }
 
+    // MARK: Volume settings
+    func pressVolumeUp() throws {
+      runAction("pressing volume up") {
+        self.device.press(XCUIDevice.Button.pressVolumeUp)
+      }
```
piotruela, line 353: Acording to the [documentation](https://developer.apple.com/documentation/xctest/xcuidevice/button) this is not available on simulator. I think we should check that before calling just as we do for `runControlCenterAction`

packages/patrol/lib/src/native/native_automator2.dart
pdenert: It works on Android emulator, but doesn't run on iOS Simulator. I've added documentation for that

packages/patrol/lib/src/native/native_automator.dart
pdenert: same

piotruela (approved):
LGTM

PR #2232, opened 2024-06-19
Add dart define from file

### LINKED ISSUE #1742: Patrol CLI doesn't support `--dart-define-from-file`
See discussion in #1477, especially [this comment](https://github.com/leancodepl/patrol/pull/1477#issuecomment-1635860323).

If we want to be consistent with Flutter, we should wait for https://github.com/flutter/flutter/issues/130599 to be resolved.

pdenert:
Close #1742


README.md
```diff
@@ -153,6 +153,21 @@ abstract class NativeAutomator<IOSServer, AndroidServer, DartClient> {
...
+If you have previously activated patrol_cli run:
+
+```bash
+dart pub global deactivate patrol_cli
+```
+
+then
+
+```bash
+cd patrol/packages/patrol_cli
```
mateuszwojtczak, line 167: I'd say `cd packages/patrol_cli`, because it's best to assume that the working directory would be the README's one

packages/patrol_cli/lib/src/commands/build_android.dart
```diff
@@ -121,12 +122,31 @@ class BuildAndroidCommand extends PatrolCommand {
...
+        'Received path for --dart-define-from-file: $dartDefineFromFilePath',
+      );
+      dartDefineFromFiles = mergeKeys(
+        json: dartDefineFromFiles,
+        dartDefines:
+            _dartDefinesReader.fromConfigFile(path: dartDefineFromFilePath),
+      );
+    }
+
+    final dartDefinesMerged =
```
mateuszwojtczak, line 140: `mergedDartDefines` sounds more English

packages/patrol_cli/lib/src/commands/test.dart
```diff
@@ -179,12 +180,31 @@ See https://github.com/leancodepl/patrol/issues/1316 to learn more.
       );
     }
 
+    final dartDefineFromFilePaths = stringsArg('dart-define-from-file');
```
mateuszwojtczak, line 183: L183-200 are pretty much the same as L125-142 in `build_android.dart` and L192-209 in `develop.dart`.

I think it's the same behavior and we could extract it and have in one place, WDYT?

mateuszwojtczak: in the end you just want some function that returns `mergedDartDefines` and `dartDefineFromFilePaths`

packages/patrol_cli/lib/src/crossplatform/app_options.dart
```diff
@@ -305,3 +319,17 @@ class MacOSAppOptions {
     return cmd;
   }
 }
+
+Map<String, String> mergeKeys({
```
mateuszwojtczak, line 323: why is this here? is it related in any way to `AppOptions`?

packages/patrol_cli/lib/src/dart_defines_reader.dart
```diff
@@ -25,6 +27,19 @@ class DartDefinesReader {
     return _parse(lines);
   }
 
+  Map<String, dynamic> fromConfigFile({required String path}) {
+    final filePath = _fs.path.join(_projectRoot.path, path);
+    final file = _fs.file(filePath);
+
+    if (!file.existsSync()) {
```
mateuszwojtczak, line 34: I think this is a very controversial decision. I can imagine this situation: you want to inject dart defines from a file and you make a typo in the path, so the file doesn't exist and your tests are being run but it just behaves in a different way than you expected.

This kind of behaviors can lead to hours-long debugging that gives you a headache but you just made a typo in a path.

BUT if we just gave a warning to the user, that this file doesn't exist, I would be so grateful as a user :D

packages/patrol_cli/lib/src/dart_defines_reader.dart
```diff
@@ -25,6 +27,19 @@ class DartDefinesReader {
...
+      return {};
+    }
+
+    final jsonString = file.readAsStringSync();
+    final json = jsonDecode(jsonString) as Map<String, dynamic>;
+    return json;
```
mateuszwojtczak, line 40: we can just return in L39

mateuszwojtczak (changes requested):

packages/patrol_cli/lib/src/dart_defines_reader.dart
```diff
@@ -25,6 +27,19 @@ class DartDefinesReader {
...
+  }
+
```
nikitadol, line 42: If the decision was made to completely parse the files into separate `define`'s, then I would like it to work the same way as in `flutter_tools`:
- fail if file is missing
- do not add `project Root.path` at beginning of file path
- support more than just `JSON` (maybe someone is using this)

How this is implemented in `flutter_tools`:
https://github.com/flutter/flutter/blob/64dd1ca9bc32fbc767a3f63e4c9c5ae2fc355380/packages/flutter_tools/lib/src/runner/flutter_command.dart#L1452-L1494

packages/patrol_cli/lib/src/dart_defines_reader.dart
pdenert: Ok, I've add logger warning

packages/patrol_cli/lib/src/commands/test.dart
pdenert: I've extracted this common code to new function

dev/e2e_app/integration_test/define_test.dart
piotruela, line 1: Won't this test cause all workflows with patrol tests in our repository to fail? I think we should exclude this test from all of them

packages/patrol_cli/lib/src/dart_defines_reader.dart
pdenert: Thanks. Good notice! I reused flutter_tools code for parsing. You can check it out now

mateuszwojtczak (changes requested):
Great, one last thing - if we allow for .env style files, let's have at least one test about it

pdenert:
> Great, one last thing - if we allow for .env style files, let's have at least one test about it 

Done @mateuszwojtczak

dev/e2e_app/integration_test/define_test.dart
pdenert: You're right. Need to manage that

mateuszwojtczak (approved):
LGTM

piotruela (approved):

packages/patrol_cli/lib/src/commands/dart_define_utils.dart
```diff
@@ -0,0 +1,28 @@
...
+/// Merge two map with overriding values for duplicated keys, by [dartDefines] values.
+Map<String, String> mergeKeys({
+  required Map<String, dynamic> json,
+  required Map<String, dynamic> dartDefines,
+}) {
+  final modified = Map<String, String>.from(json);
+
+  if (dartDefines.isNotEmpty) {
+    dartDefines.forEach((key, value) {
+      modified[key] = value as String;
```
nikitadol, line 12: @pdenert 
I tried this change in our project and realized that not everything is in the form of strings, so the code crashes here - I suspect that other projects may have a similar problem

pdenert: You're right. Can you create issue for that? We will try to address this ASAP

nikitadol: Done https://github.com/leancodepl/patrol/issues/2242

PR #2233, opened 2024-06-20
Fix complete breakage of MacOS automations introduced in #2219

pdenert:


piotruela (approved):

packages/patrol/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 3.8.1
+
+- Fix MacOSAutomator
```
mateuszwojtczak, line 3: Please let's link some issue and give more info here than just `Fix MacOSAutomator` so that when we look at it in a year, we know what happened :)

mateuszwojtczak (changes requested):
Let's add more details to the changelog entry

packages/patrol/CHANGELOG.md
pdenert: There's no issue for that. Should I create new one?

jBorkowska: I think we can link this PR here. Also we used to write (Unreleased) next to version if it's not released yet

mateuszwojtczak:
Maybe let's change the title of this PR and changelog entry to: _Fix complete breakage of MacOS automations introduced in #2219_

mateuszwojtczak (approved):
LGTM, this situation proves we need to have any minimal tests for macOS 😭

PR #2235, opened 2024-06-20
Don't use a system proxy

nikitadol:
This PR fixes an error that occurs if the system has a proxy enabled that is configured incorrectly (returns an empty response, tries to convert http to https):
```
 2024-06-20 19:14:27.173794+0200 RunnerUITests-Runner[84261:55333049] listDartTests(): failed, err: Invalid response: Optional( { URL: http://localhost:8082/listDartTests } { Status Code: 502, Headers {
 } }) Optional(0 bytes)
```


mateuszwojtczak (changes requested):
all good, please add a changelog entry for your PR like here: https://github.com/leancodepl/patrol/pull/2236

packages/patrol/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - Add privacy manifest for Apple.
+- Disabled use of system proxy in iOS/macOS (could lead to broken tests)
```
pdenert, line 4: Just one small thing, to be consistent

```suggestion
- Disabled use of system proxy in iOS/macOS (could lead to broken tests).
```

pdenert (approved):
LGTM. Thanks for contribution

PR #2237, opened 2024-06-21
Adding support to null-safety

acacioveit:
I need this to run tests on my project, but I haven't done anything according to the project standard, you can close it if you think it's necessary, but it could be useful for anyone who has the same problem.

### DIFF
--- a/packages/patrol_cli/lib/src/crossplatform/app_options.dart
+++ b/packages/patrol_cli/lib/src/crossplatform/app_options.dart
@@ -25,6 +25,7 @@ class FlutterAppOptions {
       ...['flutter', 'attach'],
       '--no-version-check',
       '--debug',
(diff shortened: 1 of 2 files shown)


mateuszwojtczak:
Hi @acacioveit, thanks for your contribution! Unfortunately, the `--no-sound-null-safety` flag has been removed with Dart 3 release (more info here https://github.com/dart-lang/sdk/issues/50348). Thus, we can't support it in `patrol_cli`.

I'll close this PR for the sake of housekeeping, people can still find closed PRs using Github search.

acacioveit:
Yes, in my current project I can't update to the latest versions of Flutter/Dart, that's why I needed to make this adjustment, then I decided to share it here, not a big deal, don't worry

PR #2254, opened 2024-07-05
Implement enableBluetooth and disableBluetooth for Android > 11

piotruela:
For Android 12 and higher we can use `svc bluetooth enable/disable` command but I couldn't find a way to implement it for lower versions.


pdenert (commented):
I think that support for android > 11 is enough for us. Could update feature-parity and changelog

pdenert (commented):
I think that support for android > 11 is enough for us. Could update feature-parity and changelog

pdenert (approved):
lgtm

PR #2265, opened 2024-07-12
Fix checking java version

pdenert:
Fallback to JAVA_HOME path when flutter doctor don't find java from flutter doctor output


piotruela (approved):
lgtm

PR #2270, opened 2024-07-18
Fix workflows for running patrol tests on Android

piotruela:
As in the title. Also changed runner for ios tests to M1.

### DIFF
--- a/dev/e2e_app/integration_test/permissions/permissions_location_test.dart
+++ b/dev/e2e_app/integration_test/permissions/permissions_location_test.dart
@@ -64,8 +64,8 @@ void main() {
         await $.native.grantPermissionOnlyThisTime();
       }
       await $.pump();
-      await $.pump(Duration(milliseconds: 500));
-      await $.pump(Duration(milliseconds: 500));
+      await $.pump(Duration(seconds: 5));
+      await $.pump(Duration(seconds: 5));
       await $.pump();
       await $.pump();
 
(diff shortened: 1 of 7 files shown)


pdenert (approved):
lgtm

dev/e2e_app/integration_test/permissions/permissions_location_test.dart
```diff
@@ -64,6 +64,10 @@ void main() {
         await $.native.grantPermissionOnlyThisTime();
       }
       await $.pump();
+      await $.pump(Duration(seconds: 5));
+      await $.pump(Duration(seconds: 5));
+      await $.pump();
+      await $.pump();
 
       await tapOkIfGoogleDialogAppears($);
```
fylyppo, line 72: It could be moved to `tapOkIfGoogleDialogAppears()` and use only one pump with longer duration if really needed

dev/e2e_app/integration_test/permissions/permissions_location_test.dart
```diff
@@ -88,6 +92,10 @@ void main() {
         await $.native2.grantPermissionOnlyThisTime();
       }
       await $.pump();
+      await $.pump(Duration(seconds: 5));
+      await $.pump(Duration(seconds: 5));
+      await $.pump();
+      await $.pump();
```
fylyppo, line 99: same

dev/e2e_app/integration_test/webview_leancode_test.dart
```diff
@@ -5,12 +5,15 @@ void main() {
     await createApp($);
 
     await $('Open webview (LeanCode)').scrollTo().tap();
+    await $.pump(Duration(seconds: 8));
+    await $.pumpAndSettle();
```
fylyppo, line 9: do we really need `pumpAndSettle` here?

fylyppo:
I see that webview fails on Android, have you tried to fix that also?

PR #2286, opened 2024-08-01
Add tags and exclude-tags support

### LINKED ISSUE #492: Add support for test tags to the CLI

pdenert:
fix #492


packages/patrol/lib/src/common.dart
```diff
@@ -169,12 +170,15 @@ DartGroupEntry createDartTestGroup(
...
   int maxTestCaseLength = global_state.maxTestLength,
+  String? tags,
+  String? excludeTags,
 }) {
   final groupDTO = DartGroupEntry(
     name: name,
     type: GroupEntryType.group,
     entries: [],
     skip: parentGroup.metadata.skip,
+    tags: parentGroup.metadata.tags.toList(),
```
mateuszwojtczak, line 181: what about `excludeTags` in here?

mateuszwojtczak (approved):
LGTM

packages/patrol/lib/src/common.dart
mateuszwojtczak: Ah, it doesn't make sense, right?

pdenert: Yeah

piotruela (approved):
Nice 👌

PR #2294, opened 2024-08-06
[WIP] Code coverage

gogolon:
This adds a possibility to collect code coverage from patrol tests using `patrol test --coverage`. Unwanted files can be ignored using glob patterns, i.e: `patrol test --coverage --coverage-ignore="**/*.g.dart"`

The PR is not finished yet - there is one thing that needs to be addressed. Currently, the coverage is only collected from isolates that are running when particular test finishes - if an isolate exits prior to this, it will be omitted. The `--pause-on-exit` flag is already set, but coverage has to be collected from the paused isolate separately, and it has to be resumed (other isolates might be waiting for its result).


packages/patrol_cli/lib/src/coverage/run_code_coverage.dart
```diff
@@ -0,0 +1,190 @@
...
+  required String flutterPackageName,
+  required Directory flutterPackageDirectory,
+  required TargetPlatform platform,
+  required Logger logger,
+  required Set<Glob> ignoreGlobs,
+}) async {
+  final homeDirectory =
+      Platform.environment['HOME'] ?? Platform.environment['USERPROFILE'];
+
+  final logsProcess = await Process.start(
```
bartekpacia, line 80: I think a `DisposeScope` should be passed here, and when it is disposed, processes should be killed and subscriptions canceled. This is a general pattern in this repo - see [AndroidTestBackend](https://github.com/leancodepl/patrol/blob/9f97af3615374ba8d5996e5ba6ad75496d72b7c5/packages/patrol_cli/lib/src/android/android_test_backend.dart#L29) as an example.

packages/patrol_cli/lib/src/coverage/run_code_coverage.dart
```diff
@@ -0,0 +1,190 @@
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
```
bartekpacia, line 3: Don't import `dart:io`, it makes mocking&testing impossible/hard. Prefer `package:file` to `dart:io`. If using some symbols from `dart:io` is  necessary, import `dart:io as io`, to make it clear it's being depended at.

Examples:
- https://github.com/leancodepl/patrol/blob/9f97af3615374ba8d5996e5ba6ad75496d72b7c5/packages/patrol_cli/lib/src/ios/ios_test_backend.dart#L56
- https://github.com/leancodepl/patrol/blob/9f97af3615374ba8d5996e5ba6ad75496d72b7c5/packages/patrol_cli/lib/src/pubspec_reader.dart#L84-L91

packages/patrol_cli/lib/src/coverage/run_code_coverage.dart
```diff
@@ -0,0 +1,190 @@
...
+
+  final map = await HitMap.parseJson(
+    coverage['coverage'] as List<Map<String, dynamic>>,
+  );
+
+  return map;
+}
+
+Future<ProcessResult> _forwardAdbPort(String host, String guest) async {
+  return Process.run('adb', ['forward', 'tcp:$host', 'tcp:$guest']);
```
bartekpacia, line 50: IIRC this will blow up on Windows, always pass `runInShell: true` in such cases

example:

https://github.com/leancodepl/patrol/blob/9f97af3615374ba8d5996e5ba6ad75496d72b7c5/packages/adb/lib/src/adb.dart#L199

packages/patrol_cli/lib/src/coverage/run_code_coverage.dart
```diff
@@ -0,0 +1,190 @@
...
+  required Logger logger,
+  required Set<Glob> ignoreGlobs,
+}) async {
+  final homeDirectory =
+      Platform.environment['HOME'] ?? Platform.environment['USERPROFILE'];
+
+  final logsProcess = await Process.start(
+    'flutter',
+    ['logs'],
+    workingDirectory: homeDirectory,
```
bartekpacia, line 83: ditto - `runInShell: true`

packages/patrol/lib/src/binding.dart
```diff
@@ -93,6 +95,27 @@ class PatrolBinding extends LiveTestWidgetsFlutterBinding {
       final nameOfRequestedTest = await patrolAppService.testExecutionRequested;
 
       if (nameOfRequestedTest == _currentDartTest) {
+        if (const bool.fromEnvironment('COVERAGE_ENABLED')) {
+          postEvent(
+            'waitForCoverageCollection',
+            {'mainIsolateId': Service.getIsolateId(Isolate.current)},
+          );
+
+          var stopped = true;
```
bartekpacia, line 104: I'd rename `stopped` -> `coverageCollected`

packages/patrol_cli/lib/src/coverage/run_code_coverage.dart
```diff
@@ -0,0 +1,190 @@
...
+}
+
+Uri _createWebSocketUri(Uri uri) {
+  final pathSegments = uri.pathSegments.where((c) => c.isNotEmpty).toList()
+    ..add('ws');
+  return uri.replace(scheme: 'ws', pathSegments: pathSegments);
+}
+
+Future<void> _saveCoverage(String report) async {
+  final coverageDirectory = Directory('coverage');
```
bartekpacia, line 60: it's not clear where this directory will be created?

packages/patrol_cli/lib/src/coverage/run_code_coverage.dart
```diff
@@ -0,0 +1,190 @@
...
+      }
+
+      final port = RegExp(':([0-9]+)/').firstMatch(vmLink)!.group(1)!;
+      final auth = RegExp(':$port/(.+)').firstMatch(vmLink)!.group(1);
+
+      final String? hostPort;
+
+      switch (platform) {
+        case TargetPlatform.android:
+          await _forwardAdbPort('61011', port);
```
bartekpacia, line 106: where does this number come from?

packages/patrol_cli/lib/src/coverage/run_code_coverage.dart
```diff
@@ -0,0 +1,190 @@
...
+import 'package:vm_service/vm_service.dart';
+import 'package:vm_service/vm_service_io.dart';
+
+Future<Map<String, HitMap>> _collectCoverage(
+  VmService client,
+  Uri vmUri,
+  String packageName,
+  String mainIsolateId,
+) async {
+  final coverage = await collect(
```
bartekpacia, line 18: maybe use a named import? it's not clear where this function comes from

packages/patrol_cli/lib/src/coverage/run_code_coverage.dart
```diff
@@ -0,0 +1,190 @@
...
+          if (event.extensionKind == 'waitForCoverageCollection') {
+            hitMap.merge(
+              await _collectCoverage(
+                serviceClient,
+                serviceUri,
+                flutterPackageName,
+                event.extensionData!.data['mainIsolateId'] as String,
+              ),
+            );
+            await serviceClient.dispose();
```
bartekpacia, line 168: should also be disposed when parent DisposeScope is disposed.

bartekpacia (commented):
nice idea:)

packages/patrol_cli/lib/src/coverage/run_code_coverage.dart
bartekpacia: Also, don't use `Process` from `dart:io`, use `package:process` to allow for dependency injection (required for testing). CTRL+F "package:process" to see examples in this repo

bartekpacia: Also, if you look around patrol, you'll notice that a wrapper around [adb](https://pub.dev/packages/adb) is used.

packages/patrol_cli/lib/src/coverage/run_code_coverage.dart
gogolon: I pass a `FileSystem` instance now and use it to create the directory, so I guess it should be clear now

packages/patrol_cli/lib/src/coverage/run_code_coverage.dart
gogolon: fully random, should we just pick a port and use it or search for an unused one at runtime?

bartekpacia: That's okay, let's just extract it to a constant and add a doc comments saying it's a randomly selected number.

packages/patrol/lib/src/binding.dart
```diff
@@ -93,6 +95,27 @@ class PatrolBinding extends LiveTestWidgetsFlutterBinding {
...
+            {'mainIsolateId': Service.getIsolateId(Isolate.current)},
+          );
+
+          var coverageCollected = false;
+
+          registerExtension('ext.patrol.markTestCompleted',
+              (method, parameters) async {
+            coverageCollected = true;
+            return ServiceExtensionResponse.result(jsonEncode({}));
+          });
```
bartekpacia, line 110: trailing comma

packages/patrol/lib/src/common.dart
```diff
@@ -237,19 +238,34 @@ String deduplicateGroupEntryName(String parentName, String currentName) {
   );
 }
 
-/// Recursively prints the structure of the test suite.
+/// Recursively prints the structure of the test suite and reports test count
+/// of the top-most group
 @internal
-void printGroupStructure(DartGroupEntry group, {int indentation = 0}) {
+int reportGroupStructure(DartGroupEntry group, {int indentation = 0}) {
```
bartekpacia, line 244: This function now also calls `postEvent() from `dart:developer`, update the comment accordingly.

packages/patrol_cli/lib/src/coverage/coverage_tool.dart
```diff
@@ -0,0 +1,213 @@
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
```
bartekpacia, line 3: `import 'dart:io' as io;`, to make it obvious

packages/patrol_cli/lib/src/coverage/coverage_tool.dart
```diff
@@ -0,0 +1,213 @@
...
+  final DisposeScope _disposeScope;
+
+  Future<void> run({
+    required String flutterPackageName,
+    required TargetPlatform platform,
+    required Logger logger,
+    required Set<Glob> ignoreGlobs,
+  }) async {
+    final homeDirectory =
+        Platform.environment['HOME'] ?? Platform.environment['USERPROFILE'];
```
bartekpacia, line 43: This is also a hidden dependency on `io.Platform` now. Use `Platform` from [`package:platform`](https://pub.dev/packages/platform) (just like is now done with `FileSystem` and `ProcessManager`)

Example: https://github.com/leancodepl/patrol/blob/370a0fe808074d7edfae5cbcc48249be01c16d47/packages/patrol_cli/lib/src/commands/doctor.dart#L13

packages/patrol_cli/lib/src/coverage/coverage_tool.dart
```diff
@@ -0,0 +1,213 @@
...
+            .asyncMap(_collectTotalTestCount)
+            .first;
+        logger.info('Total test count: $totalTestCount');
+
+        var count = 0;
+        final coverageCollectionCompleter = Completer<void>()
+          ..disposedBy(
+            scope,
+            null,
+          );
```
bartekpacia, line 81: can be single line

packages/patrol_cli/lib/src/coverage/coverage_tool.dart
```diff
@@ -0,0 +1,213 @@
...
+    );
+    _disposeScope.addDispose(serviceClient.dispose);
+
+    await serviceClient.streamListen('Extension');
+    final completer = Completer<int>()..disposedBy(_disposeScope, 0);
+    serviceClient.onExtensionEvent.listen((event) async {
+      if (event.extensionKind == 'testCount') {
+        completer.complete(event.extensionData!.data['testCount'] as int);
+      }
+    }).disposedBy(_disposeScope);
```
bartekpacia, line 126: cleaning up after ourselves
noice👍

packages/patrol_cli/lib/src/coverage/coverage_tool.dart
```diff
@@ -0,0 +1,213 @@
...
+
+extension<T> on Completer<T> {
+  void disposedBy(DisposeScope disposeScope, T disposeValue) {
+    disposeScope.addDispose(() {
+      if (!isCompleted) {
+        complete(disposeValue);
+      }
+    });
+  }
+}
```
bartekpacia, line 213: This seems like a good addition to [`package:dispose_scope`](https://pub.dev/packages/dispose_scope) by @RobertOdrowaz.

packages/patrol_cli/lib/src/coverage/device_to_host_port_transformer.dart
```diff
@@ -0,0 +1,80 @@
...
+        final forwardList = await _processManager.run(
+          // TODO: Add to wrapper
+          ['adb', 'forward', '--list'],
+          runInShell: true,
+        );
+        final output = forwardList.stdout as String;
+        hostPort = int.tryParse(
+          RegExp('tcp:([0-9]+) tcp:$devicePort').firstMatch(output)?.group(1) ??
+              '',
+        );
```
bartekpacia, line 61: Be careful with calling `adb` directly – sometimes the adb server might not be up, and then it will print more than just output of `adb forward --list`. For example:

```
$ adb kill-server # explicitly kill adb server
$ adb forward --list
* daemon not running; starting now at tcp:5037
* daemon started successfully
```

I think the above output will make `int.tryParse` return null.

It might not be the case here (`flutter run` (or equivalent) had likely been execute before, so adb connection is there), but let's be on the safe side w.r.t input validation.

(btw, `adb forward` sounds like it should be added to `package:adb`, but i don't wanna complain too much)

packages/patrol_cli/lib/src/coverage/vm_connection_details.dart
```diff
@@ -0,0 +1,36 @@
...
+  final String auth;
+
+  Uri get uri => Uri.parse('http://127.0.0.1:$port/$auth');
+  Uri get webSocketUri {
+    final pathSegments = uri.pathSegments.where((c) => c.isNotEmpty).toList()
+      ..add('ws');
+    return uri.replace(scheme: 'ws', pathSegments: pathSegments);
+  }
+
+  static VMConnectionDetails? tryExtractFromLogs(String logsLine) {
```
bartekpacia, line 17: A few tests for this would be nice to see

packages/patrol_cli/lib/src/coverage/coverage_tool.dart
```diff
@@ -0,0 +1,213 @@
...
+            .transform(
+              DeviceToHostPortTransformer(
+                processManager: _processManager,
+                devicePlatform: platform,
+                adb: _adb,
+                logger: logger,
+              ),
+            )
+            .whereNotNull()
+            .asBroadcastStream();
```
bartekpacia, line 69: If `rxdart` is used only for `whereNotNull` here, I'd consider using plain old `map` instead. Let's keep dependencies to reasonable minimum. No strong opinion though.

bartekpacia (approved):
:)

packages/patrol/lib/src/common.dart
gogolon: the comment says `reports test count of the top-most group`, isn't this enough?

packages/patrol_cli/lib/src/coverage/device_to_host_port_transformer.dart
gogolon: afair, if the `adb` server is not running, and we start it by calling `adb forward --list`, the list will be empty anyway. What they say [here](https://developer.android.com/studio/run/emulator-networking#adbredir) implies that stopping adb server also stops the redirections. Therefore, it seems reasonable to me that we return null in this case

gogolon: if it did return a proper output after a few lines of startup logs, we would parse the ports properly, wouldn't we? the regex takes the whole output and searches for a match, so I think that would not be an issue

packages/patrol_cli/lib/src/coverage/vm_connection_details.dart
gogolon: I refactored the method to make it use `Uri.parse` instead of parsing manually. I added two very simple test cases. Any specific test cases you had in mind?

bartekpacia: Nope, the ones you added look good enough!

packages/patrol_cli/lib/src/coverage/device_to_host_port_transformer.dart
bartekpacia: if you tested that situation and it works, then that's cool. i just wanted to make sure it's handled.

btw, in the future I'd really advise to implement this "adb forward" functionality in your `adb` package, now i remembered it has some [additional checks](https://github.com/leancodepl/patrol/blob/9db9449afe2410b221ef61e890d6b9ba83889ada/packages/adb/lib/src/internals.dart#L33-L98) that are there for a reason.

packages/patrol/lib/src/binding.dart
```diff
@@ -93,6 +95,29 @@ class PatrolBinding extends LiveTestWidgetsFlutterBinding {
...
+              coverageCollected = true;
+              return ServiceExtensionResponse.result(jsonEncode({}));
+            },
+          );
+
+          while (!coverageCollected) {
+            // The loop is needed to keep this isolate alive until the coverage
+            // data is collected.
+            await Future<void>.delayed(const Duration(seconds: 1));
+          }
```
piotruela, line 118: Will the Isolate be terminated while waiting for `Completer`?

packages/patrol_cli/lib/src/coverage/coverage_tool.dart
```diff
@@ -0,0 +1,213 @@
...
+    required Set<Glob> ignoreGlobs,
+  }) async {
+    final homeDirectory =
+        _platform.environment['HOME'] ?? _platform.environment['USERPROFILE'];
+    final hitMap = <String, coverage.HitMap>{};
+
+    await _disposeScope.run(
+      (scope) async {
+        final logsProcess = await _processManager.start(
+          ['flutter', 'logs'],
```
piotruela, line 52: Pass `FlutterCommand` here, instead of calling hardcoded `flutter`.

packages/patrol_cli/lib/src/coverage/run_code_coverage.dart
piotruela: Actually, I think we should search for unused port instead of hardcoding it. We already have some issues with other ports being hardcoded so we should avoid it whenever possible

packages/patrol/lib/src/binding.dart
gogolon: I'm not sure I understand the question. Could you be more specific?

packages/patrol_cli/lib/src/coverage/coverage_tool.dart
gogolon: This is a special case here. The command is ran in home directory to avoid any project related errors like the one below

```
The Xcode project defines schemes: dev, prod
You must specify a --flavor option to select one of the available schemes.
```

If it's run in the home dir, does it make sense to run it using fvm (or are there any other use cases for setting flutter command)?

piotruela: Then I'm not sure if that's good approach. Globally set flutter version (using fvm / the standard way) may be different from the one set in the project. Isn't there a risk that this will fail due to versions incompatibility?

packages/patrol/lib/src/binding.dart
piotruela: I mean, instead of doing this loop you could use `Completer` to wait for `ext.patrol.markTestCompleted` to finish:
```dart
  final completer = Completer<void>();

  registerExtension(
    'ext.patrol.markTestCompleted',
    (method, parameters) async {
      completer.complete();
      return ServiceExtensionResponse.result(jsonEncode({}));
    },
  );

  // Wait until the coverage is collected
  await completer.future;
  ```
  But I'm not sure if the Isolate will be terminated while waiting here `await completer.future`

File: packages/patrol_cli/lib/src/coverage/coverage_tool.dart (hunk 21 above)
Comment (gogolon) [reply packages/patrol_cli/lib/src/coverage/coverage_tool.dart:52]: `flutter logs` seems to just listen to adb on android, and on iOS it connects to the VM and listens to stdout. It seems unlikely to me that flutter changes so much that it fails.
The only other solution I could think of is somehow getting the path to the flutter used by flutter command (fvm), and then running it from home

Links to log readers below
https://github.com/flutter/flutter/blob/9d971fada15eb6570a54b65bbd13bdef4a4a16d5/packages/flutter_tools/lib/src/ios/devices.dart#L1154
https://github.com/flutter/flutter/blob/9d971fada15eb6570a54b65bbd13bdef4a4a16d5/packages/flutter_tools/lib/src/android/android_device.dart#L790

File: packages/patrol/lib/src/binding.dart (hunk 20 above)
Comment (gogolon) [reply packages/patrol/lib/src/binding.dart:118]: 🙈 it should work just fine

File: packages/patrol_cli/lib/src/coverage/coverage_tool.dart (hunk 21 above)
Comment (piotruela) [reply packages/patrol_cli/lib/src/coverage/coverage_tool.dart:52]: Okey right, it seems unlikely. Then there is yet another risk that someone will only have fvm flutter on their machine. Then `flutter` command will fail as unrecognised.

Comment (gogolon) [reply packages/patrol_cli/lib/src/coverage/coverage_tool.dart:52]: the installation guide of `fvm` recommends to have a globally installed flutter instance
https://fvm.app/documentation/getting-started/installation#recommendation
so I guess it's theoretically possible but probably not very common.

File: packages/patrol_cli/lib/src/coverage/run_code_coverage.dart (hunk 7 above)
Comment (gogolon) [reply packages/patrol_cli/lib/src/coverage/run_code_coverage.dart]: done, please have a look

File: packages/patrol_cli/lib/src/coverage/coverage_tool.dart (hunk 21 above)
Comment (piotruela) [reply packages/patrol_cli/lib/src/coverage/coverage_tool.dart:52]: Okey, I'm good with that

**Review by piotruela — APPROVED** [review]
Nice work, lgtm 👍

File: packages/patrol_cli/lib/src/coverage/device_to_host_port_transformer.dart (hunk 17 above)
Comment (gogolon) [reply packages/patrol_cli/lib/src/coverage/device_to_host_port_transformer.dart:76]: added in #2332

**Comment by pdenert** [comment]
@gogolon good job! Please bump versions of patrol and patrol_cli, remember about `packages/patrol_cli/lib/src/base/constants.dart` and compatibility_table. Also add Changelog for both packages

File: packages/patrol_cli/lib/src/test_bundler.dart [hunk 22]
```diff
@@ -92,7 +92,7 @@ Future<void> main() async {
     );
     testExplorationCompleter.complete(dartTestGroup);
     print('patrol_test_explorer: obtained Dart-side test hierarchy:');
-    printGroupStructure(dartTestGroup);
+    reportGroupStructure(dartTestGroup);
```
pdenert, line 95: This probably force incompatibility with old patrol versions

packages/patrol_cli/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 3.2.0
+- Add code coverage collection support (#2294)
```
pdenert, line 2:
```suggestion

- Add code coverage collection support. (#2294)
```

packages/patrol/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## Unreleased
+## 3.11.0
 
+- Add code coverage collection support (#2294)
 - No throw error in `selectFineLocation` when it's already selected. (#2302)
 - Add Option to select tap Location in enterText and enterTextByIndex (#2312)
```
pdenert, line 5:
```suggestion
- Add code coverage collection support. (#2294)
- No throw error in `selectFineLocation` when it's already selected. (#2302)
- Add option to select tap location in `enterText` and `enterTextByIndex`. (#2312)
```

pdenert (approved):
LGTM - Just two small changes in changelogs and you can merge it

PR #2295, opened 2024-08-06
Delete GoogleService-Info.plist from example project

AnthonyDadeWT:
I raised this issue and the discord and got feedback from @mateuszwojtczak 

My project runs security scans on our Flutter app binaries and we had a critical violation due to the file:

ios/.symlinks/patrol/example/ios/Runner/GoogleService-Info.plist

The security scan detected sensitive information (in this case it thought it was a credit card #) in this file, and is probably related to the API key or GCM_SENDER_ID values.

GoogleService-Info.plist is a Firebase configuration file that contains secrets/API keys. This file is usually not version controlled to keep these values private. Due to this, I don't think this file belongs here, and having this present could also cause the same security violation for others using Patrol that also scan their app(s). 

Read more in the discord thread where I brought this up:

https://discord.com/channels/1167030497612922931/1270048825569972394


piotruela (commented):
Thanks for the contribution and sorry it slipped under the radar. I think this change still makes sense: the plist serves no purpose anymore (Firebase in the example initializes from `firebase_options.dart` behind a flag) and it only trips security scanners.

@AnthonyDadeWT are you still up for finishing it? If so:
- rebase to the latest `master`,
- remove the 4 `GoogleService-Info.plist` references from `project.pbxproj` (file reference, group entry and the `Resources` build phase). Deleting just the file breaks the example iOS build with "Build input file cannot be found",
- delete `packages/patrol/example/android/app/google-services.json` too. It carries the same kind of API key and nothing applies the `google-services` gradle plugin, so it's unused.

If not, no worries, we'll take it over and finish it.

PR #2296, opened 2024-08-07
Disable ios workflows on PR

piotruela:
We don't want to run `test-ios-simulator-webview` and `test-ios-simulator` on every PR since they are not working currently.


pdenert (approved):
lgtm

PR #2300, opened 2024-08-08
Change java path parsing

### LINKED ISSUE #2288: Can't build a test when Java path is different than the one from Android Studio
Building a test hangs on the `./gradlew...` command due to wrong parsing of java path obtained from `flutter doctor -v`.

pdenert:
fix #2288

### DIFF
--- a/packages/patrol_cli/lib/src/android/android_test_backend.dart
+++ b/packages/patrol_cli/lib/src/android/android_test_backend.dart
@@ -128,13 +128,14 @@ class AndroidTestBackend {
         (line) async {
           if (line.contains('• Java binary at:') &&
               javaCompleterPath.isCompleted == false) {
-            final path = line
-                .replaceAll('• Java binary at:', '')
-                .replaceAll(
-                  _platform.isWindows ? r'\bin\java' : '/bin/java',
-                  '',
-                )
-                .trim();
+            var path = line.replaceAll('• Java binary at:', '').trim();
+            if (path != '/usr/bin/java') {
+              path = path.replaceAll(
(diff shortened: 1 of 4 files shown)


packages/patrol_cli/lib/src/android/android_test_backend.dart
```diff
@@ -128,13 +128,14 @@ class AndroidTestBackend {
         (line) async {
```
gogolon, line 128: shouldn't we pass the entire flutter command here and invoke it along with its arguments to avoid the same problem that was in #2320?

gogolon: I also see we do the same thing in `buildApkConfigOnly`

packages/patrol_cli/lib/src/android/android_test_backend.dart
```diff
@@ -128,13 +128,14 @@ class AndroidTestBackend {
...
               javaCompleterPath.isCompleted == false) {
-            final path = line
-                .replaceAll('• Java binary at:', '')
-                .replaceAll(
-                  _platform.isWindows ? r'\bin\java' : '/bin/java',
-                  '',
-                )
-                .trim();
+            var path = line.replaceAll('• Java binary at:', '').trim();
+            if (path != '/usr/bin/java') {
```
gogolon, line 132: i think a comment explaining this special case would be nice

piotruela (approved):

gogolon (approved):

gogolon (approved):

PR #2312, opened 2024-08-13
EnterText TapLocation iOS

Surio89:
Since I had problems inserting a text into a keycloak password field on iOS Emulator, I analyzed the problem and found that a button placed right above the input field was the problem (Toggle Password Visibility Button). 

image

By default for iOS, Patrol clicks on the relative position x:0.9, y:0.9 of the text field to be entered. So at the bottom right of the text field by default. However, in my example there is a button that intercepts pointer input. I have created a possibility to determine the relative position where the click in the text field takes place.

I added an `Offset tapLocation` to `enterText` and `enterTextByIndex` for both NativeAutomators. 

This only affects iOS, because as far as I have seen, android does not simulate a click on a specific position.

### DIFF
--- a/packages/patrol/CHANGELOG.md
+++ b/packages/patrol/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.11.0
+
+- Add Option to select tap Location in enterText and enterTextByIndex (#2312)
+
 ## 3.10.0
 
 - Implement `enableBluetooth` and `disableBluetooth` methods for Android > 11. (#2254)
--- a/packages/patrol/android/src/main/kotlin/pl/leancode/patrol/Automator.kt
+++ b/packages/patrol/android/src/main/kotlin/pl/leancode/patrol/Automator.kt
@@ -367,8 +367,9 @@ class Automator private constructor() {
         val uiObject = uiDevice.findObject(uiSelector)
 
         if (keyboardBehavior == KeyboardBehavior.showAndDismiss) {
-            val x = uiObject.width() * dx
-            val y = uiObject.height() * dy
+            val bounds = uiObject.visibleBounds
+            val x = bounds.width() * dx
+            val y = bounds.height() * dy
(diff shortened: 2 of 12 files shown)


piotruela (changes requested):
Thanks for the input. The code looks good, just update the CHANGELOG and we can proceed with merging the PR.

lunaticcoding:
> Thanks for the input. The code looks good, just update the CHANGELOG and we can proceed with merging the PR.

Done :)

piotruela (approved):
lgtm

lunaticcoding:
@piotruela somehow the commits made by @Surio89 are now attributed to you :S so he does not show up as a contributor even though he contributed to the project :/ 

Can we do something about that so he shows up as well? Would be important to us.

mateuszwojtczak:
@lunaticcoding that's very unfortunate, let's see what we can do about it

mateuszwojtczak:
@lunaticcoding @Surio89 I think it's because @Surio89 committed changes just as a git author Jannik Neidig and not as Surio89 and Github doesn't know that it's the same person. Look at commits tab at the top of this page and you can see that on the other hand commits made by @lunaticcoding are recognized correctly. When @piotruela squashed and merged, he only committed them, but original authors remained and it should be treated as contributions. Also, @lunaticcoding commits are being seen as his contributions in the official repo's contributors page.

So, what can we do? Maybe we could somehow rewrite the history or let Github know that this git author is actually @Surio89?

mateuszwojtczak:
Also: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/troubleshooting-commits/why-are-my-commits-linked-to-the-wrong-user

Surio89:
@mateuszwojtczak I have realized that unfortunately I have been using the wrong email address. I have added this email address as a secondary address to my github account. Now I appear in the list thanks a lot!

PR #2321, opened 2024-08-22
Add steps allowing conditional running test workflows on fork PRs

piotruela:
We want to be able to perform e2e tests on PRs that comes from forks. This change makes it possible but only when the workflow is triggered by someone with `write` access to the repository.

Inspired **heavily** by [this article](https://michaelheap.com/access-secrets-from-forks/).


pdenert (approved):
lgtm

PR #2355, opened 2024-10-07
Add a guide on how to debug Patrol CLI

karolinaoparczyk:


.run/Patrol CLI.run.xml
pdenert, line 1: In the docs you mention file: `.run/patrol_cli.run.xml` and I think it's better name then `.run/Patrol CLI.run.xml` - I don't like spaces in names of that type of files

CONTRIBUTING.md
```diff
@@ -0,0 +1,47 @@
...
+```json
+{
+  "version": "0.1.0",
+  "configurations": [
+    {
+      "name": "Patrol CLI",
+      "request": "launch",
+      "type": "dart",
+      "program": "patrol/packages/patrol_cli/bin/main.dart",
+      "cwd": "absolute_path_to_my_project",
```
pdenert, line 21: Again, it does not have to be an absolute path. It can be relative, but taking the current working directory as the starting point

pdenert (approved):
lgtm

Issue #2357, opened 2024-10-07
When copying native view props as Selector arguments, they are copied with double quotes


jBorkowska:
Could be copied with single quotes, as usually linters suggest

PR #2358, opened 2024-10-07
Devtools fixes

### LINKED ISSUE #2356: Native view details panel in devtools is not scrolling even when its content overflows
Panel on the right is not scrollable.

Image

jBorkowska:
This PR fixes #2356 and #2357


pdenert (commented):
It looks good, but I guess you have to bump version, right? Also I can't found any changelog in this package - do we want to add one?

piotruela (approved):

PR #2360, opened 2024-10-08
Fix test name for androidx.test:orchestrator:1.5.0

tddang-linagora:
## Why?
- https://github.com/android/android-test/issues/2255
- Test name is currently under the form of `test_file test_group Test description` which contain spaces
- According to https://firebase.google.com/docs/test-lab/android/instrumentation-test#orchestrator: `Test Lab always uses the latest version of Orchestrator.`, which is 1.5.0 at the time of creating this PR
- Our Firebase Test Lab always throw the exact same exception and stacktrace as https://github.com/android/android-test/issues/2255, confirming that Firebase Test Lab ignore androidx.test:orchestrator version in app:build.gradle
- You can reproduce this error by using the Patrol example, with the latest version of both CLI and pub.dev dependency, by updating androidx.test:orchestrator from 1.4.2 to 1.5.0

### DIFF
--- a/packages/patrol/android/src/main/kotlin/pl/leancode/patrol/PatrolJUnitRunner.java
+++ b/packages/patrol/android/src/main/kotlin/pl/leancode/patrol/PatrolJUnitRunner.java
@@ -115,8 +115,8 @@ public class PatrolJUnitRunner extends AndroidJUnitRunner {
             List<DartGroupEntry> dartTestCases = ContractsExtensionsKt.listTestsFlat(dartTestGroup, "");
             List<String> dartTestCaseNamesList = new ArrayList<>();
             for (DartGroupEntry dartTestCase : dartTestCases) {
-                dartTestCaseSkipMap.put(dartTestCase.getName(), dartTestCase.getSkip());
-                dartTestCaseNamesList.add(dartTestCase.getName());
+                dartTestCaseSkipMap.put(dartTestCase.getName().replace(" ", "-"), dartTestCase.getSkip());
+                dartTestCaseNamesList.add(dartTestCase.getName().replace(" ", "-"));
             }
             Object[] dartTestCaseNames = dartTestCaseNamesList.toArray();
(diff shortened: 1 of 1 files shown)


piotruela (commented):
Thanks for opening this PR. We are aware of this issue and we are also currently discussing the best way to solve it. I'm not sure if solution proposed by you will work in 100% cases.
I.e.
Test named "Smoke test - 1" will be changed into "Some-test---1" during discovery step. Then when calling for execution, it will be called as "Some test   1" which does not match original name.
On top of my head, I think we will have to forbid using the character that we want to use as a separator.

piotruela:
Closing this PR as issue was fixed in #2361 and deployed in `patrol: 3.11.1`.
Thanks again for the input!

Wreos:
Should I update "androidx.test:orchestrator version as well? Currently I have 1.4.2 (as in documentation) ?

Because after updating to Patrol 3.11.1 I have another problem when run tests in firebase
`java.lang.RuntimeException: Could not open stream for output
     FATAL EXCEPTION: AndroidTestOrchestrator
Process: androidx.test.orchestrator, PID: 8626
java.lang.RuntimeException: Could not open stream for output
	at androidx.test.orchestrator.AndroidTestOrchestrator.getOutputStream(AndroidTestOrchestrator.java:5)
	at androidx.test.orchestrator.AndroidTestOrchestrator.executeNextTest(AndroidTestOrchestrator.java:11)
	at androidx.test.orchestrator.AndroidTestOrchestrator.runFinished(AndroidTestOrchestrator.java:10)
	at androidx.test.orchestrator.TestRunnable.run(TestRunnable.java:11)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:644)
	at java.lang.Thread.run(Thread.java:1012)`

Issue #2366, opened 2024-10-09
Clearing permissions state between test on iOS


jBorkowska:
We'd like to clear permissions state like on Android.

We'll use [resetAuthorizationStatus](https://developer.apple.com/documentation/xctest/xcuiapplication/3526066-resetauthorizationstatus) method.
Permissions to be cleared (from [XCUIProtectedResource](https://developer.apple.com/documentation/xctest/xcuiprotectedresource)):

Requiring iOS 13.4:
- location
- contacts
- calendar
- reminders
- photos
- bluetooth
- microphone
- camera
- homeKit
- mediaLibrary
- keyboardNetwork
- systemRootDirectory
- userDesktopDirectory
- userDocumentsDirectory
- userDownloadsDirectory

Requiring iOS 14.0:
- health

Requiring iOS 15:
- userTracking
- focus
- removableVolumes
- networkVolumes
- appleEvents

Requiring iOS 15.4:
- localNetwork

PR #2367, opened 2024-10-09
Clear permissions between tests on iOS

karolinaoparczyk:
Issue: #2366 

Resetting permissions on iOS is available from iOS 13.4, but some permissions require a higher version (check #2366).

Permissions with error `use of undeclared identifier`:
```
XCUIProtectedResourceAppleEvents
XCUIProtectedResourceRemovableVolumes
XCUIProtectedResourceNetworkVolumes
XCUIProtectedResourceUserDownloadsDirectory
XCUIProtectedResourceUserDocumentsDirectory
XCUIProtectedResourceUserDesktopDirectory
XCUIProtectedResourceSystemRootDirectory
```

These should be investigated further.

The `clear_permissions_test.dart` has only few permissions tested, because there is an issue with getting `PermissionStatus.permanentlyDenied` status for them.

I've reverted ios target to 11.0 and the test passed 👀 

TODO:
- Changes should also include a parameter that would turn clearing permissions on.

https://github.com/user-attachments/assets/59088842-26c9-45bc-8a9a-951757caff18

### DIFF
--- a/.github/workflows/test-ios-simulator.yaml
+++ b/.github/workflows/test-ios-simulator.yaml
@@ -104,7 +104,7 @@ jobs:
           sleep 10 # See https://github.com/leancodepl/patrol/issues/1282
 
           TESTS_EXIT_CODE=0
-          patrol test --exclude ${{ env.EXCLUDED_TESTS }} --verbose || TESTS_EXIT_CODE=$?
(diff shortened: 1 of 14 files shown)


pdenert (approved):
lgtm

PR #2387, opened 2024-10-30
Change patrol output

pdenert:
This PR completely changes the current output when firing patrol tests. It also adds two new flags:
- showTestSteps,
- showFlutterLogs.

To reach final effect, we need to handle exceptions during test, but this require #2362.
It's also adding new package: patrol_log, that need setup of the pubspec, changelog etc.

Preview of the new output on the video:

https://github.com/user-attachments/assets/518f151a-b0a6-4fb7-87e7-ca9e5c86842f

PR #2394, opened 2024-11-05
Use new notification tile identifier for iOS 18

piotruela:
Fixes tapping on notification in iOS 18. 
Looks like `identifier` changed in the latest iOS version:
iOS 17:
image
iOS 18:
image


piotruela:
>Update changelog and version in pubspec

I've updated the changelog, but I didn't bump the version because I plan to release this change together with #2387.

pdenert (approved):
lgtm

PR #2434, opened 2024-11-26
Disable printing logs in nested `waitUntilVisible` and `waitUntilExits` calls

pdenert:


packages/patrol_finders/lib/src/custom_finders/patrol_finder.dart
```diff
@@ -400,11 +404,19 @@ class PatrolFinder implements MatchFinder {
...
   /// Timeout is globally set by [PatrolTester.config.visibleTimeout]. If you
   /// want to override this global setting, set [timeout].
-  Future<PatrolFinder> waitUntilVisible({Duration? timeout}) =>
+  Future<PatrolFinder> waitUntilVisible({
+    Duration? timeout,
+    bool enablePatrolLog = true,
+  }) =>
       wrapWithPatrolLog(
         action: 'waitUntilVisible',
         color: AnsiCodes.cyan,
```
mateuszwojtczak, line 413: Not related to this PR: I would totally remove color information from this method. It should be like `type: step` or something that under the hood does `color: AnsiCodes.cyan`. I would not see this as the concern of this method.

mateuszwojtczak (approved):
LGTM

PR #2361, opened 2024-10-08
fix: Modify test names inside PatrolJUnitRunner to remove spaces

piotruela:
Replaces spaces in the name returned by `listDartTest` with a version that does not contain spaces.
The change is also visible in the report, as it reflects the names of the tests (functions) that are executed as tests.

This PR is in fact slightly modified #2360 created by [tddang-linagora](https://github.com/tddang-linagora) so kudos to him for the initial work.

PR #2437, opened 2024-11-28
Remove whitespace replacement from Android runner

piotruela:
This PR reverts a change made [here](https://github.com/leancodepl/patrol/pull/2361). It was previously required as `androidx.test:orchestrator:1.5.0` [introduced a regression](https://github.com/android/android-test/issues/2255) which was later reverted in `androidx.test:orchestrator:1.5.1`.

Also I updated the setup instructions to suggest using the latest version of the orchestrator.

Added a check if we are running on Orchestrator 1.5.0. If we detect this version, we warn user about that version


pdenert (approved):
lgtm

Issue #2463, opened 2024-12-21
waitUntilVisible() fails on Columns with SizedBox


FritzMatthaeus:
I have found an answer on why the ``waitUntilVisible()`` test fails when using a **Key** or searching by **Class**. It has to do with the Widget containing a Column with a **SizedBox**. 

this will be found when calling ``$(Keys.elementA).waitUnitilVisible() ``: 

```dart
class ElementA extends StatelessWidget {
  ElementA() : super(key: Keys.elementA);

  @override
  Widget build(BuildContext context) {
    return const Column(
      children: [
        Text(
          'Element A',
        ),
      ],
    );
  }
}
```

this will **NOT** be found when calling ``$(Keys.elementB).waitUnitilVisible() `` or ``$(ElementB).waitUntilVisible()``: 

```dart 
class ElementB extends StatelessWidget {
  ElementB() : super(key: Keys.elementB);

  @override
  Widget build(BuildContext context) {
    return const Column(
      children: [
        Text(
          'Element B',
        ),
        SizedBox(height: 48)
      ],
    );
  }
}
```

I have created a reproducable example: [Patrol Bug Finder Example](https://github.com/mevi-media/patrol_example.git)

_Originally posted by @FritzMatthaeus in https://github.com/leancodepl/patrol/discussions/2459#discussioncomment-11637154_

FritzMatthaeus:
While the pull request will fix the problem on the given example, the overall problem seems to be bigger: The hitTestable will fail, even with Alignments, if we hit some "empty" space. In the example, the test will pass because we can get a result on ``finder.hitTestable(at: Alignment.topCenter)``. But it will fail, if we wrap the Column into a Padding. 

The Pull Request might be an enhancement, up to your decision. But i would recommend to update the Documentation and clarify, that a finder with byKey or byClass should not be used on Complex Widgets with Paddings, Columns inside itself or it's children. At least if you want to check the visibility. It should be set on lowest child inside the widget tree. 

Setting the Key as in this example (though this is what i want to check, if ElementB is visible), will fail. With or without alignments: 

```dart
class ElementB extends StatelessWidget {
  ElementB() : super(key: Keys.elementB);

  @override
  Widget build(BuildContext context) {
    return Padding(
      padding: const EdgeInsets.all(24.0),
      child: const Column(
        children: [
          Text(
            'Element B',
          ),
          SizedBox(height: 48), 
          Text('Some other widget'),
        ],
      ),
    );
  }
}
```

[...]

PR #2425, opened 2024-11-22
Add android flavor to dartDefines.

GChanathip:
this PR fix issue https://github.com/leancodepl/patrol/issues/2256

PR #2464, opened 2024-12-21
fix: Bug on waitUntilVisible() and visible

FritzMatthaeus:
This fix is related to the issue #2463 

The methods ``waitUntilVisible()`` and the getter ``visible`` default to ``Alignment.center`` when calling ``hitTestable()`` on the Flutter test framework. In case the tested Widget has an empty SizedBox in it's middle, both methods will fail. This can be fixed by allowing to pass an Alignment. 

Example:

```dart
class ElementB extends StatelessWidget {
  ElementB() : super(key: Keys.elementB);

  @override
  Widget build(BuildContext context) {
    return const Column(
      children: [
        Text(
          'Element B',
        ),
        SizedBox(height: 48)
      ],
    );
  }
}
```

Calling ``$(Keys.elementB).waitUntilVisible()`` on this widget will fail as the SizedBox will cover the middle of the Widget.
Calling ``$(Keys.elementB).waitUntilVisible(alignment: Alignment.topCenter)`` on this widget will pass.

The ``visible`` getter has to be refactored as a function to be able to pass an Alignment parameter: ``visible({Alignment alignment = Alignment.center})`` 

An alternative Approach would be to evaluate the finder for all possible Alignments when calling ``waitUntilVisible()`` and ``visible``. A separate Pull Request will be created for that solution. You are in a better position to decide which way would suit best for the package.

### DIFF
--- a/dev/e2e_app/integration_test/common.dart
+++ b/dev/e2e_app/integration_test/common.dart
@@ -5,7 +5,7 @@ import 'package:patrol/patrol.dart';
 export 'package:flutter_test/flutter_test.dart';
 export 'package:patrol/patrol.dart';
 
-final _patrolTesterConfig = PatrolTesterConfig();
+final _patrolTesterConfig = PatrolTesterConfig(printLogs: true);
 final _nativeAutomatorConfig = NativeAutomatorConfig(
   findTimeout: Duration(seconds: 20), // 10 seconds is too short for some CIs
 );
--- a/dev/e2e_app/integration_test/text_fields_test.dart
+++ b/dev/e2e_app/integration_test/text_fields_test.dart
@@ -1,0 +1,32 @@
+import 'package:e2e_app/text_fields_screen.dart';
+import 'package:flutter/material.dart';
+
+import 'common.dart';
+
+void main() {
+  patrol('Can enter text into same field twice', ($) async {
+    await $.pumpWidgetAndSettle(const TextFieldsScreen());
+    await $.pumpAndSettle();
+
+    // Enter text into the first text field.
+    await $(const Key('textField1')).enterText('User');
+    await $(const Key('buttonFocus')).tap();
+    expect($('User'), findsOneWidget);
+
+    // Enter text into the first text field again. After focusing on the button.
+    await $(const Key('textField1')).enterText('User2');
+    await $(const Key('buttonUnfocus')).tap();
+    expect($('User'), findsNothing);
+    expect($('User2'), findsOneWidget);
+
+    // Enter text into the first text field again. After unfocusing the button.
+    // Then enter text into the second text field and into the first field again.
+    await $(const Key('textField1')).enterText('User3');
+    await $(const Key('textField2')).enterText('User4');
+    await $(const Key('textField1')).enterText('User5');
+    expect($('User2'), findsNothing);
+    expect($('User3'), findsNothing);
+    expect($('User4'), findsOneWidget);
+    expect($('User5'), findsOneWidget);
+  });
+}
--- a/dev/e2e_app/lib/text_fields_screen.dart
+++ b/dev/e2e_app/lib/text_fields_screen.dart
@@ -1,0 +1,49 @@
+import 'package:flutter/material.dart';
+
+class TextFieldsScreen extends StatefulWidget {
+  const TextFieldsScreen({super.key});
+
+  @override
+  State<TextFieldsScreen> createState() => _TextFieldsScreenState();
+}
+
+class _TextFieldsScreenState extends State<TextFieldsScreen> {
+  FocusNode focusNode = FocusNode();
+
+  @override
+  Widget build(BuildContext context) {
+    return MaterialApp(
+      home: Scaffold(
+        appBar: AppBar(
+          title: const Text('Text Fields'),
+        ),
+        body: Center(
+          child: SingleChildScrollView(
+            child: Column(
+              children: [
+                const TextField(
+                  key: Key('textField1'),
+                ),
+                ElevatedButton.icon(
+                  key: const Key('buttonFocus'),
+                  onPressed: () =>
+                      FocusScope.of(context).requestFocus(focusNode),
+                  label: const Icon(Icons.search),
+                  focusNode: focusNode,
+                ),
+                const TextField(
+                  key: Key('textField2'),
+                ),
+                ElevatedButton.icon(
+                  key: const Key('buttonUnfocus'),
+                  onPressed: () => FocusScope.of(context).unfocus(),
+                  label: const Icon(Icons.search),
+                ),
+              ],
+            ),
+          ),
+        ),
+      ),
+    );
+  }
+}
--- a/docs/debugging-patrol-tests.mdx
+++ b/docs/debugging-patrol-tests.mdx
@@ -20,10 +20,20 @@ Here is how you can do it:
 
 2. Run your patrol tests using `develop` command with the same arguments as you would normally do.
 
-3. When the tests will start running, at some point you will see a message with the VM service URI.
-   Copy the URI from the message.
+3. When the tests will start running, at some point you will see a message with a link to Patrol devtools extension.
+   Copy the last part of the URI from the message.
+   Eg. for this link:
 
-`The Dart VM service is listening on http://127.0.0.1:63725/57XmBI_pwSA=/`
+   `Patrol DevTools extension is available at http://127.0.0.1:9104/patrol_ext?uri=http://127.0.0.1:52263/F2-CH29gR1k=/`
+   
+   copy `http://127.0.0.1:52263/F2-CH29gR1k=/`.
+
+   <Warning>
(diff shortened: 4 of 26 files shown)


FritzMatthaeus:
the second commit offers an alternative solution to the one of the first commit. Here the API remains untouched but we check all possible Alignments for a possible hit test:

```dart
/// Waits until [finder] finds at least one visible widget.
  ///
  /// Throws a [WaitUntilVisibleTimeoutException] if more time than specified by
  /// the timeout passed and no widgets were found.
  Future<PatrolFinder> waitUntilVisible(
    Finder finder, {
    Duration? timeout,
    bool enablePatrolLog = true,
  }) {
    return TestAsyncUtils.guard(
      () => wrapWithPatrolLog(
        action: 'waitUntilVisible',
        finder: finder,
        color: AnsiCodes.cyan,
        enablePatrolLog: enablePatrolLog,
        function: () async {
          final duration = timeout ?? config.visibleTimeout;
          final end = tester.binding.clock.now().add(duration);
          final hitTestableFinders = alignments.map((alignment) => finder.hitTestable(at: alignment));
          final hitTestableEvaluations = hitTestableFinders.map((finder) => finder.evaluate());
          while (hitTestableEvaluations.map((result) => result.isNotEmpty).firstOrNull == null) {
            final now = tester.binding.clock.now();
            if (now.isAfter(end)) {
              throw WaitUntilVisibleTimeoutException(
                finder: finder,
                duration: duration,
              );
            }            
            await tester.pump(const Duration(milliseconds: 100));
          }
          return PatrolFinder(
            finder: hitTestableFinders.firstWhere((finder) => finder.evaluate().isNotEmpty),
            tester: this,
          );
        },
      ),
    );
  }
```

Looking forward to your feedback.

FritzMatthaeus:
Both solutions can be tested on this example: [Patrol Example](https://github.com/mevi-media/patrol_example.git)

mateuszwojtczak:
Hi @FritzMatthaeus! Welcome to the contributor lounge. 🎉

The idea of being "visible" is very hard to set in stone. We thought hit-testability is the closest we can get to visibility, since most of the time users want to interact with something if they wait for it to be visible. Thus, we check for `hitTestable()` evaluations.

In your case, the `Column` is a widget that takes all available height and is not responding to hit tests. So, when the default alignment is `Alignment.center`, then it hits a center of a Column that is "transparent" to hit tests and it can't find any hit-testable widgets.

Your first commit provides a nice fix of just passing by the alignment parameter to the user, so that they're in full control of where the hit test happens.

Your second commit tries to hit-test in 9 points which can lead to more unexpected issues for other users. Still, it won't cover all the cases because if the child is moved by a pixel (e.g. with a padding), then the hit-test fails.

That said, we think your first commit is the best fix for this issue right now.

I have generated a golden test image from your `ElementA` example (Column is in blue and doesn't react to hit tests, but it's center is there).

issue_test

FritzMatthaeus:
Hi @mateuszwojtczak,

i fully agree with your decision. This PR has been the result of my learning curve on "visible" and "hitTestable". My final learning is, to be aware of Columns and Rows when testing "visibility". Maybe this should be added to the docs somewhere. I'll think about it and maybe i come up with a PR on the docs. 

Have good year 2025 and thank you for your work on this package!

pdenert:
Hi @FritzMatthaeus, maybe it wasn't obvious from @mateuszwojtczak answer, but we would like to merge your first commit, as it extends user possibilities, so it would be nice to have it in Patrol. 
You can reopen the PR and restore changes from first commit or create new one. If you don't have time for that, let us know and we'll do it by ourself.

Thanks again for your contribution.

pdenert (commented):
Very good job! Just few organizing things:
1. Please format changed files. You can use `dart format ./packages/patrol_finders`
2.  We try to stick with the standard flutter max 80 chars per line. Please change your comments to fit this requirement.
3. Add new unreleased entry to patrol_finders changelog. See example here: https://github.com/leancodepl/patrol/pull/2425/files#diff-e17f3235f9e47a742b45c5a87b2ce49cb726b37d26fb9d478ce3880c6e61ba98

FritzMatthaeus:
> Very good job! Just few organizing things:
> 
> 1. Please format changed files. You can use `dart format ./packages/patrol_finders`
> 2. We try to stick with the standard flutter max 80 chars per line. Please change your comments to fit this requirement.
> 3. Add new unreleased entry to patrol_finders changelog. See example here: https://github.com/leancodepl/patrol/pull/2425/files#diff-e17f3235f9e47a742b45c5a87b2ce49cb726b37d26fb9d478ce3880c6e61ba98

I did as requested. Thanks for guiding me through this process. Crash course in good package maintainance for me ;)

pdenert (commented):
One last thing. There's one warning from analyzer:
```
warning - lib/src/custom_finders/patrol_tester.dart:5:8 - Unused import: 'package:patrol_finders/src/custom_finders/utils.dart'. Try removing the import directive. - unused_import
```

Please remove this import and then we can merge it

FritzMatthaeus:
> One last thing. There's one warning from analyzer:
> 
> ```
> warning - lib/src/custom_finders/patrol_tester.dart:5:8 - Unused import: 'package:patrol_finders/src/custom_finders/utils.dart'. Try removing the import directive. - unused_import
> ```
> 
> Please remove this import and then we can merge it

fixed :)

FritzMatthaeus:
After working a while with Patrol i am not quite sure anymore if merging this PR is recommended. It does enhance the methods ``visibleAt`` and ``waitUntiVisible`` if the user calls them directly. But it also produces inconsistencies as ``waitUntilVisible`` is also invoked by the methods ``enterText``, ``tap`` and ``longPress``. I have been experimenting with adding the alignment property to those methods as well. And i have found out that this breaks ``tap``. While the invoked waitUntilVisible call is successful, the following underlying tap call of the finder fails as it implicitly taps the center and there is no way of configuring it:

this example fails with a ``WaitUntilVisibleTimeoutException`` if any other alignment than Alignment.center is required to successfully hittest it.

```dart
 Future<void> tap(
    Finder finder, {
    SettlePolicy? settlePolicy,
    Duration? visibleTimeout,
    Duration? settleTimeout,
    bool enablePatrolLog = true,
  }) {
    return TestAsyncUtils.guard(
      () => wrapWithPatrolLog(
        action: 'tap',
        finder: finder,
        color: AnsiCodes.yellow,
        enablePatrolLog: enablePatrolLog,
        function: () async {
          final resolvedFinder = await waitUntilVisible(
            finder,
            timeout: visibleTimeout,
            enablePatrolLog: false,
          );
          await tester.tap(resolvedFinder.first);
          await _performPump(
            settlePolicy: settlePolicy,
            settleTimeout: settleTimeout,
          );
        },
      ),
    );
  }
```

this example fails also, though it is no longer a ``WaitUntilVisibleTimeoutException`` but the ``tester.tap(resolvedFinder.first)`` throws an exception.

```dart

$(Widget).tap(at: Alignment.topCenter);

Future<void> tap(
    Finder finder, {
    SettlePolicy? settlePolicy,
    Duration? visibleTimeout,
    Duration? settleTimeout,
    bool enablePatrolLog = true,
    Alignment alignment = Alignment.center
  }) {
    return TestAsyncUtils.guard(
      () => wrapWithPatrolLog(
        action: 'tap',
        finder: finder,
        color: AnsiCodes.yellow,
        enablePatrolLog: enablePatrolLog,
        function: () async {
          final resolvedFinder = await waitUntilVisible(
            finder,
            timeout: visibleTimeout,
            enablePatrolLog: false,
            alignment: alignment
          );
          await tester.tap(resolvedFinder.first); // <- the error occurs here
          await _performPump(
            settlePolicy: settlePolicy,
            settleTimeout: settleTimeout,
          );
        },
      ),
    );
  }
```
So passing an Alignment property to this method is not possible.

This might end up in an inconsistency for the user as he might successfully test the visibility and hittest the Widget by calling
```dart
await $(Widget).waitUntilVisible(at: Alignment.topCenter);
```
but when he tries to tap it, the test will fail with a ``WaitUntilVisibleTimeoutException``
```dart
await $(Widget).tap();
```

I did not test ``longPress`` and ``enterText``.

I have come to the conclusion that it might be a better solution to stick with the current implementation of the visibility checks without Alignment parameters but to improve the documentation to clarify the visibility concept when it comes to Widgets including Rows and Columns. Or just take in the ``visibleAt`` method and leave ``waitUntilVisible`` as is.

As the PR is now complete, i will leave this decision up to you. ;)
Thank you for your good work and this great package!

Best regards,
Fritz

pdenert (approved):
LGTM. 
I've added issue to enhance the tap method: https://github.com/leancodepl/patrol/issues/2484

pdenert:
One last request @FritzMatthaeus - please resolve changelog conflict and then I merge this

FritzMatthaeus:
> One last request @FritzMatthaeus - please resolve changelog conflict and then I merge this

oh my god, i feel like a total newbie (which i am)! hopefully my next PR is more smoothly ;)

PR #2481, opened 2025-01-08
Remove exception from step entry

jBorkowska:


pdenert (commented):
Please update changelogs and versions in all this packages.
Also after publishing `patrol_log`, you have to update it's version in `patrol_finders`, and then do the same thing in `patrol` package. Maybe it'll be easier to split this into 3 PRs? Or for now you can update only changelogs with "unreleased"

jBorkowska:
I updated changelogs, but I didn't bump the versions yet (regardless of what I wrote in those changelogs) - it has to be done during release

pdenert (approved):
lgtm

PR #2503, opened 2025-01-23
Bump Gradle in example apps so it's possible to build them on the latest JDK

bartekpacia:
Gradle 8.9 doesn't support the latest stable Java 23. ([see Gradle-Java compatibility matrix](https://docs.gradle.org/current/userguide/compatibility.html))

Java 23 is the default when using `brew install openjdk` (which is arguably a very popular way to install JDK on macOS). I use it.

### Repro

```
$ brew install openjdk
```

```
$ echo $JAVA_HOME
/Library/Java/JavaVirtualMachines/openjdk.jdk/Contents/Home
```

```
$ cd packages/patrol/example
```

```
$ flutter build apk --config-only
```

```
cd android && ./gradlew :app:assembleDebug
```

Results in:

```
FAILURE: Build failed with an exception.

* What went wrong:
BUG! exception in phase 'semantic analysis' in source unit '_BuildScript_' Unsupported class file major version 67
> Unsupported class file major version 67

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

BUILD FAILED in 26s
Running Gradle task 'assembleDebug'...                             27.5s
```

> [!NOTE]
>
>Using `flutter build apk --debug` works because by default, it ignores `$JAVA_HOME` and defaults to the Java bundled in IntelliJ IDEA/Android Studio. Still, when the developer calls `flutter config --jdk-dir "$JAVA_HOME` (and JDK 23 is in `$JAVA_HOME`), the same problem occurs.

### DIFF
--- a/dev/e2e_app/android/gradle/wrapper/gradle-wrapper.properties
+++ b/dev/e2e_app/android/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.9-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.12-all.zip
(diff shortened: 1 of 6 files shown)


pdenert (commented):
It's ok, but please add entry to patrol package changelog

bartekpacia:
Sure, done!

pdenert:
Thanks! And now resolve the conflict please 😉

pdenert (approved):
LGTM

Issue #2460, opened 2024-12-19
Disable Analytics


soyllamas:
### Use case

The CLI currently fails to execute commands when analytics tracking calls are blocked by custom DNS configurations (e.g., DNS-level ad/tracking blockers). This makes the CLI unusable for users who implement network-level privacy controls.

<img width="1832" alt="image" src="https://github.com/user-attachments/assets/0b0e3edb-5b65-431d-abef-8a0574d9a6c3" />

### Proposal

1. Add a user-configurable analytics flag:
    - Allow users to disable analytics via CLI flags (e.g., `patrol --disable-analytics` or `patrol --analytics=false`).
    - This gives users explicit control over analytics collection.
2. Implement "fire and forget" analytics:
    - Make analytics non-blocking by running it asynchronously.
    - Analytics failures won't affect command execution or user experience.
    - Main command continues regardless of analytics success/failure.

pdenert:
Hi @soyllamas. Thanks for the issue. 
Currently, at the first run, patrol asks whether you agree for analytics or not. If you don't agree, patrol won't be sending analytics.
If you agreed for analytics, but now want to disable it, you can find file `$Home/.config/patrol_cli/analytics.json` and change: `"enabled":true` to `"enabled":false`. Alternatively, you can delete this file and the next time you launch patrol it will ask for analytics permission again.

soyllamas:
@pdenert Thanks it worked! 🎉

pdenert:
So, I'm closing the issue. We also work on adding environment variable to steer that. Should be available soon

PR #2509, opened 2025-01-28
Gracefully handle analytics failing to send (patrol_cli)

alexandradeas:
Handle exceptions raised when analytics fail to be sent.

This is related to https://github.com/leancodepl/patrol/issues/2460 where I ran into the same due to a DNS block on the domain - which I'd expect to be common in CI environments where egress is tightly controlled. As analytics aren't an essential part of the CLI, my rationale is that if the call fails it should not halt execution

### DIFF
--- a/packages/patrol_cli/CHANGELOG.md
+++ b/packages/patrol_cli/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+- Gracefully handle when analytics fail to send (#2460)
+
 ## 3.5.0
 
 - Add `PATROL_ANALYTICS_ENABLED` environment variable to disable analytics. (#2483)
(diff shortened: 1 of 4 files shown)


packages/patrol_cli/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 3.5.1
+- Gracefully handle when analytics fail to send (#2460)
```
pdenert, line 2: Please change version to `Unreleased` and add  empty line between (for formatting purposes)

```suggestion
## Unreleased

- Gracefully handle when analytics fail to send. (#2460)
```

pdenert (commented):
@alexandradeas thanks for contribution! You're right, this should be done like that from the begging. Please change the changelog, and than we can merge this.

alexandradeas:
> @alexandradeas thanks for contribution! You're right, this should be done like that from the begging. Please change the changelog, and than we can merge this.

my bad for not reading the instructions properly, updated that now

pdenert (approved):
LGTM

pdenert:
Thanks again for contribution @alexandradeas

PR #2510, opened 2025-01-28
Remove build phases ios

pdenert:
As we checked - adding build phases is unneeded in patrol iOS setup.


packages/patrol/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- Remove adding build phases from iOS setup. (#2510)
```
piotruela, line 3: Should we mention that change in changelog? 🤔

piotruela (commented):
Just one question

packages/patrol/CHANGELOG.md
pdenert: I added this for 2 reasons:
1. I want to keep track of when it happened,
2. so that users can see that we've removed it.

piotruela (approved):

PR #2516, opened 2025-02-02
Fix running ios tests in example project with dev flavor

piotruela:
This PR fix running iOS tests in example app using `dev` flavor. 
Before the fix, process was failing with `xcodebuild: error: Failed to build workspace Runner with scheme dev.: There are no test bundles available to test.` error which was caused by multiple `.xctest` files generated for `dev` scheme upon building the tests.
To fix that, `shouldAutocreateTestPlan = "YES"` line was removed from `dev.xcscheme` file.
PR also contains other changes which were required in order to make project compatible with Flutter 3.27.0, but those changes were also introduced in #2510.


pdenert (approved):
lgtm

PR #2528, opened 2025-02-17
Add assertions for to coordinates in swipe method

hamed-rezaee:
Ensured `to.dx` and `to.dy` are within the inclusive `0-1` range.
This aligns with the existing assertions for from and prevents out-of-bounds errors.


pdenert (approved):
LGTM. Thanks for contribution

PR #2536, opened 2025-02-24
Add `--coverage-package` flag to `patrol_cli`

### LINKED ISSUE #2491: add --coverage-package to cli
### Use case

i have an app structured as a monorepo (without Melos, just a dart monorepo as all packages are only local): 

structure:
- packages/foo
- lib/myapp
 
Running ``patrol test --coverage`` only covers the app files (which are in /lib) but not the files in the packages which contain the major part of the UI. This might be a case of a badly structured monorepo but it is what it is. It would be great if ``foo`` would be covered too. 

I have [found](https://github.com/LouiseHsu/flutter/commit/5a3a09f8475d68484913e4322acb03d5fc7cd5ff) that flutter test does support including packages into the coverage by running:
``flutter test --coverage --coverage-package foo``

currently this is not supported by the Patrol CLI. For my purpose this would be a valuable feature

### Proposal

include ``--coverage-package [mypackage]`` to Patrol CLI like 
``patrol test --coverage --coverage-package foo``

gogolon:
Closes #2491

### DIFF
--- a/packages/patrol_cli/CHANGELOG.md
+++ b/packages/patrol_cli/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 - Gracefully handle when analytics fail to send. (#2460)
 - Handle uninstallation of the app after tests finish when using AGP 8.2+. (#2535)
+- Add `--coverage-packages` flag to `patrol test` which specifies which packages should be included 
+in the coverage report. (#2536)
 
 ## 3.5.0
 
--- a/packages/patrol_cli/CHANGELOG.md
+++ b/packages/patrol_cli/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 - Gracefully handle when analytics fail to send. (#2460)
 - Handle uninstallation of the app after tests finish when using AGP 8.2+. (#2535)
-- Add `--coverage-packages` flag to `patrol test` which specifies which packages should be included 
+- Add `--coverage-packages` flag to `patrol test` that specifies which packages should be included 
 in the coverage report. (#2536)
 
 ## 3.5.0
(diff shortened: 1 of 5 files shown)


packages/patrol_cli/lib/src/commands/test.dart
```diff
@@ -254,10 +255,13 @@ See https://github.com/leancodepl/patrol/issues/1316 to learn more.
       unawaited(
         _coverageTool.run(
           device: device,
-          flutterPackageName: config.flutterPackageName,
           platform: device.targetPlatform,
           logger: _logger,
           ignoreGlobs: ignoreGlobs,
+          packagesRegExps: switch (coveragePackagesRegExps.length) {
+            0 => {RegExp(config.flutterPackageName)},
+            _ => coveragePackagesRegExps.map(RegExp.new).toSet(),
```
pdenert, line 263: Will this contain also app under test package?

packages/patrol_cli/lib/src/commands/test.dart
```diff
@@ -254,10 +255,13 @@ See https://github.com/leancodepl/patrol/issues/1316 to learn more.
       unawaited(
         _coverageTool.run(
           device: device,
-          flutterPackageName: config.flutterPackageName,
           platform: device.targetPlatform,
           logger: _logger,
           ignoreGlobs: ignoreGlobs,
+          packagesRegExps: switch (coveragePackagesRegExps.length) {
+            0 => {RegExp(config.flutterPackageName)},
```
pdenert, line 262: Shouldn't there be used just packageName from L164?

pdenert (commented):
Please also update changelog

packages/patrol_cli/lib/src/commands/test.dart
gogolon: I thought `packageName` refers to native package name? What we need here is the name of the flutter package

packages/patrol_cli/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - Gracefully handle when analytics fail to send. (#2460)
 - Handle uninstallation of the app after tests finish when using AGP 8.2+. (#2535)
+- Add `--coverage-packages` flag to `patrol test` which specifies which packages should be included
```
pdenert, line 5:
```suggestion
- Add `--coverage-packages` flag to `patrol test` that specifies which packages should be included 
```

pdenert (approved):
LGTM

Issue #2383, opened 2024-10-29
Allow quitting `patrol develop` with `q`


jBorkowska:
Now we support only ctrl+c

Issue #2143, opened 2024-03-05
Print help during develop mode


jBorkowska:
Like Flutter handles `h` key during debugging in `flutter run` and shows help regarding how to hot restart or quit, we could do it as well. Information to show there could be link to devtools (now it's at the beginning and if a user decides to open it after a lot of actions, he has a lot of logs to scroll through). Also we could show there how to hot restart (`r`) and how to quit (`ctrl+c`).

PR #2577, opened 2025-03-28
Add ability to quit the `patrol develop` process by pressing q on the keyboard

AnthonyDadeWT:
This PR implements the requested functionality described in #2383 

I've added user-facing messaging to the console logs during the patrol develop process to advertise this new feature.

## Changes
- Added `onQuit` callback parameter to `FlutterTool` class to handle cleanup operations
- Modified the quit handler to call the uninstall function before exiting
- Updated `attachForHotRestart` and `attach` methods to accept and pass through the `onQuit` callback
- Modified `develop.dart` to pass the uninstall function as the `onQuit` callback

This change ensures that when users quit the Flutter tool using 'q', all associated processes are properly cleaned up, preventing any lingering processes from running in the background.

## Example
When running the Flutter develop command and pressing 'q':
1. The Flutter attach process is killed
2. The uninstall function is called to clean up the app and test processes
3. The program exits cleanly

### DIFF
--- a/packages/patrol_cli/lib/src/crossplatform/flutter_tool.dart
+++ b/packages/patrol_cli/lib/src/crossplatform/flutter_tool.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 import 'dart:io' as io;
+import 'dart:io' show exit;
 
 import 'package:dispose_scope/dispose_scope.dart';
 import 'package:meta/meta.dart';
@@ -17,6 +18,7 @@ class FlutterTool {
     required Platform platform,
     required DisposeScope parentDisposeScope,
     required Logger logger,
+    this.onQuit,
   })  : _stdin = stdin,
         _processManager = processManager,
         _platform = platform,
@@ -30,6 +32,7 @@ class FlutterTool {
   final Platform _platform;
   final DisposeScope _disposeScope;
   final Logger _logger;
+  final Future<void> Function()? onQuit;
 
   bool _hotRestartActive = false;
   bool _logsActive = false;
@@ -43,6 +46,7 @@ class FlutterTool {
     required Map<String, String> dartDefines,
     required bool openDevtools,
     bool attachUsingUrl = false,
+    Future<void> Function()? onQuit,
   }) async {
     if (io.stdin.hasTerminal) {
       _enableInteractiveMode();
@@ -64,6 +68,7 @@ class FlutterTool {
         debugUrl: url,
         dartDefines: dartDefines,
         openBrowser: openDevtools,
+        onQuit: onQuit,
       );
     } else {
       await Future.wait<void>([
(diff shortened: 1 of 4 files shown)


packages/patrol_cli/lib/src/crossplatform/flutter_tool.dart
```diff
@@ -213,7 +237,9 @@ class FlutterTool {
           observationUrlCompleter?.complete(url);
         }
         if (line.startsWith('Showing ') && line.endsWith('logs:')) {
-          _logger.success('Hot Restart: logs connected');
+          _logger
```
piotruela, line 240: The way you implemented it will cause the hints for the "r" and "q" key commands to be printed separately. I'm not sure if it wouldn't be a better idea to display all key commands together in a single block, similar to how Flutter does it (see screenshot below)
image

@jBorkowska what are your thoughts on that?

piotruela (commented):
Thanks for the contribution ❤️
I left a few minor comments, but overall I like the solution.
Please also add an entry to `patrol_cli` changelog

packages/patrol_cli/lib/src/crossplatform/flutter_tool.dart
AnthonyDadeWT: Thanks, I would be happy to make both of these changes. 

Are there any other commands available for use while using the develop command other than r (hot restart) and q (quit) ?

AnthonyDadeWT: I guess a follow up question I have is when do we want to display all commands? My thinking is that it would make most sense to do this when hot restart is attached and is available for use.

jBorkowska: What do you mean by `printed separately` @piotruela ?

jBorkowska: Also if we'd like to have this help to be shown like in Flutter (ofc we'd like to 😄), then we have to add support for `h` key

jBorkowska: Which is a separate task

piotruela: We spoke offline with @jBorkowska, and concluded that it would be better to print it all at the same time. This will be a good start for [help command](https://github.com/leancodepl/patrol/issues/2143) which we want to implement in future.
>Are there any other commands available for use while using the develop command other than r (hot restart) and q (quit) ? Those are the only ones I know of.

Nope, there is no other commands for now.

>I guess a follow up question I have is when do we want to display all commands? My thinking is that it would make most sense to do this when hot restart is attached and is available for use.

Yes, let's wait for hot restart to be attached and log it then.

piotruela (commented):
Please update develop tests in `patrol_develop_test.dart` to look for  `r Hot restart` instead of `press "r" to restart`. This should fix patrol develop tests that are currently failing

AnthonyDadeWT:
Thanks, Ill make sure to run tests locally before pushing next time.

PR #2578, opened 2025-03-30
Update Android setup instruction to use kotlin DSL syntax

### LINKED ISSUE #2546: getting started instructions for native for build.gradle.kts files
Recent flutter versions file new project uses build.gradle.kts files.
https://patrol.leancode.pl/getting-started uses old style files.

File-new project flutter stable and add patrol should "just work"
We would sure appreciate it
thanks

piotruela:
Closes #2546 
I've also updated files in `patrol/example` to use latest syntax


packages/patrol/example/android/settings.gradle.kts
```diff
@@ -0,0 +1,26 @@
...
+    repositories {
+        google()
+        mavenCentral()
+        gradlePluginPortal()
+    }
+}
+
+plugins {
+    id("dev.flutter.flutter-plugin-loader") version "1.0.0"
+    id("com.android.application") version "8.7.0" apply false
```
pdenert, line 21: We have 8.6.0 before

docs/documentation/index.mdx
```diff
@@ -106,6 +106,107 @@ Check out our video version of this tutorial on YouTube!
     Psst... Android is a bit easier to set up, so we recommend starting with it!
 
 <Accordion defaultOpen title="Android setup">
+  <Steps>
```
pdenert, line 109: I would add some description here that there are 2 different versions of gradle

pdenert:
What about dev/e2e_app gradle files?

packages/patrol/example/android/settings.gradle.kts
piotruela: I generated this file using `flutter create` command. I guess this is now the default version

packages/patrol/example/android/settings.gradle.kts
```diff
@@ -0,0 +1,26 @@
...
+    id("com.google.gms.google-services") version "4.4.0" apply false
+    id("org.jetbrains.kotlin.android") version "1.8.22" apply false
```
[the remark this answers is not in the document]
piotruela: I changed it to match the version pointed in the [table](https://docs.gradle.org/current/userguide/compatibility.html#kotlin)

docs/documentation/index.mdx
piotruela: I've renamed Android setup to point out that it is for Flutter projects created with 3.29.0+. I don't think we need to go deeper with explanation of Kotlin Gradle DSL etc

pdenert: My idea was to point out that it is needed for projects that have Gradle DSL files, that includes all flutter projects create with version 3.29+, but not only, as projects created with older flutter can be migrated

PR #2581, opened 2025-04-01
Display a friendly device name

### LINKED ISSUE #2561: Display a user-friendly device name
When running Patrol tests without explicitly selecting the target device, the device ID is printed:
```console
$ dart run patrol_cli:main test -t integration_test/{path-to-test}
No device specified, using the first one (3C201FDJG00376)
...
```

Instead, I suggest displaying the name of the device, similar to how `flutter run` does it:

```console
$ flutter run --flavor dev -t lib/main_dev.dart
Launching lib/main_dev.dart on Pixel 8 Pro in debug mode...
...
```

PiotrRogulski:
Removes `resolvedName` and replaces it with `device.name` and/or `device.id`.

- Closes #2561


piotruela (approved):
🙇 ✅ 💯

PR #2584, opened 2025-04-02
Add alignment to all finder methods

### LINKED ISSUE #2484: Add `alignment` parameter to $.tap
As [FritzMatthaeus](https://github.com/FritzMatthaeus) noticed in: [this comment](https://github.com/leancodepl/patrol/pull/2464#issuecomment-2577232122), after adding option to pass Alignment to waitUntilVisible, we need to enhance other methods with that option. So user can tap, on something that was found as visible with alignment different than center.
We need to:
1. Add `alignment` to tap.
2. Replace `tap` with `tapAt` when using `alignment`.

jBorkowska:
This PR closes #2484

### DIFF
--- a/packages/patrol_finders/test/patrol_tester_test.dart
+++ b/packages/patrol_finders/test/patrol_tester_test.dart
@@ -6,8 +6,8 @@ import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:patrol_finders/src/custom_finders/custom_finders.dart';
 
-import 'utils/long_screen_with_long_button.dart';
-import 'utils/screen_with_long_button.dart';
+import 'utils/long_screen_with_partially_visible_widget.dart';
+import 'utils/screen_with_partially_visible_widget.dart';
 import 'utils/set_physical_size.dart';
 
 void main() {
@@ -96,7 +96,7 @@ void main() {
           setPhysicalSize(tester.tester, width);
 
           await tester.pumpWidget(
-            ScreenWithLongButton(
+            ScreenWithPartiallyVisibleWidget(
               width: width,
               testedWidget: ElevatedButton(
                 onPressed: () {},
@@ -121,7 +121,7 @@ void main() {
           var counter = 0;
 
           await tester.pumpWidget(
-            ScreenWithLongButton(
+            ScreenWithPartiallyVisibleWidget(
               width: width,
               testedWidget: ElevatedButton(
                 onPressed: () {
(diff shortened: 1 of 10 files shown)


packages/patrol_finders/test/patrol_tester_test.dart
```diff
@@ -211,6 +292,87 @@ void main() {
...
+
+          final controller = TextEditingController();
+
+          await tester.pumpWidget(
+            MaterialApp(
+              home: Scaffold(
+                body: SingleChildScrollView(
+                  scrollDirection: Axis.horizontal,
+                  child: Container(
+                    color: Colors.blue,
```
piotruela, line 346: Is color necessary?

packages/patrol_finders/test/patrol_tester_test.dart
```diff
@@ -211,6 +292,87 @@ void main() {
...
+                        SizedBox(
+                          width: width,
+                          child: TextField(),
+                        ),
+                      ],
+                    ),
+                  ),
+                ),
+              ),
+            ),
```
piotruela, line 321: Can we extract this widget into something more generic that will take `width` and `testedWidget` as a parameter from outside? We could then add some comments to this generic widget with explanation why we set those width in such way etc.

piotruela (commented):
Just a few small comments. 
+ Please add an entry to `patrol_finders` changelog.

packages/patrol_finders/test/patrol_tester_test.dart
jBorkowska: No, was needed for debugging, will remove

packages/patrol_finders/test/utils/screen_with_long_button.dart
```diff
@@ -0,0 +1,30 @@
+import 'package:flutter/material.dart';
+
+class ScreenWithLongButton extends StatelessWidget {
```
piotruela, line 3: The name is misleading, we are not always passing button. How about `ScreenWithPartiallyVisibleWidget`?

packages/patrol_finders/test/utils/screen_with_long_button.dart
```diff
@@ -0,0 +1,30 @@
...
+        scrollDirection: Axis.horizontal,
+        child: SizedBox(
+          width: 2 * width,
+          child: Center(
+            child: SizedBox(
+              width: width,
+              child: testedWidget,
+            ),
+          ),
+        ),
```
piotruela, line 26: Also please add some comments explaining what is the idea here.

packages/patrol_finders/test/utils/long_screen_with_long_button.dart
```diff
@@ -0,0 +1,39 @@
+import 'package:flutter/material.dart';
+
+class LongScreenWithLongButton extends StatelessWidget {
```
piotruela, line 3: ditto

piotruela (approved):
Awesome 💯  LGTM

PR #2590, opened 2025-04-07
Add `--build-name` and `--build-number` flags to Patrol CLI commands

### LINKED ISSUE #2391: Add flags for `build-number` and `build-name`
### Use case

Right now you can't pass `--build-number` and `--build-name` flags to a Flutter build when building Patrol tests with the app.

Thus, it's hard to work force update features around and the other way around -- it's hard to test the "force update" feature itself. :)

### Proposal

There should be `--build-number` and `--build-name` flags in `patrol build`, `patrol test` and `patrol develop` and they should just behave the same as in Flutter CLI.

pdenert:
Fixes #2391

# Summary

Adds `--build-name` and `--build-number` flags to `patrol test`, `patrol develop`, and `patrol build` commands, enabling developers to specify custom app version information during testing.

# Changes

- Added flag definitions to `PatrolCommand` base class
- Extended `FlutterAppOptions` to include build name and build number parameters
- Updated all Flutter build invocations to pass version flags to underlying Flutter CLI
- Updated documentation with examples and usage information
- Fixed formatting in `http_parser.h` file
- Updated iOS simulator to iPhone 16 (18.5) in `prepare-ios` job


packages/patrol_cli/lib/src/crossplatform/app_options.dart
```diff
@@ -142,6 +146,14 @@ class AndroidAppOptions {
       effectiveDartDefines['FLUTTER_APP_FLAVOR'] = flavor;
     }
 
+    // Add build number and build name to dart defines
+    if (flutter.buildNumber != null) {
+      effectiveDartDefines['FLUTTER_BUILD_NUMBER'] = flutter.buildNumber!;
+    }
+    if (flutter.buildName != null) {
+      effectiveDartDefines['FLUTTER_BUILD_NAME'] = flutter.buildName!;
+    }
```
gogolon, line 155: consider pattern matching to avoid null assertion

gogolon (approved):

packages/patrol_cli/lib/src/commands/build_android.dart
```diff
@@ -42,6 +42,14 @@ class BuildAndroidCommand extends PatrolCommand {
...
 
     usesAndroidOptions();
+    argParser.addOption(
+      'build-number',
+      help: 'A build number to be used as --build-number in Flutter build.',
+    );
+    argParser.addOption(
+      'build-name',
+      help: 'A build name to be used as --build-name in Flutter build.',
+    );
```
piotruela, line 52: Consider moving those two into `usesBuildVersion()`(or use some other name) method instead of pasting it multiple times.

piotruela (commented):
I'd love to see a unit test that checks if those new params actually end up in the `flutter build` command invocation when passed.

packages/patrol/example/integration_test/app_version_test.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:example/main.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:patrol/patrol.dart';
+
+void main() {
+  // This test must be run with correct build-name (1.2.3) and
+  // build-number (123) flags.
```
pdenert: Maybe add ready to use patrol command here

packages/patrol/example/integration_test/app_version_test.dart
pdenert, line 1: I would prefer this test in the dev/e2e_app

Issue #2566, opened 2025-03-18
`patrol build` doesn't check if CLI and patrol versions are compatible


jBorkowska:
CI in one of our projects used `patrol build` with CLI 3.5.1 and patrol 3.13.0 and succeeded, but `patrol test` ran locally with exact same versions fails on compatibility check

AnthonyDadeWT:
I can probably pick this up after https://github.com/leancodepl/patrol/pull/2579 is merged

jBorkowska:
Closing as resolved in https://github.com/leancodepl/patrol/pull/2597

PR #2597, opened 2025-04-14
[patrol_cli]: Add throwToolExit to build commands when packages are incompatible

AnthonyDadeWT:
This PR adds an improved error system during the build and test processes that alerts users when they are using incompatible versions of patrol and patrol_cli. It enhances the error message shown during these commands when the compatibility check fails, providing clearer guidance on how to resolve version compatibility issues.

Addresses #2566.

-----------------

**Changes**

- Added version compatibility checks in all build commands (Android, iOS, macOS)
- Created a new helper method in CompatibilityChecker to show warnings without failing the build process
- Enhanced compatibility  error messages to include:
  - Clear explanation of the incompatibility issue
  - The maximum compatible CLI version for the project's patrol version
  - Specific command to downgrade to the compatible version
  - Option to upgrade both dependencies
  - Link to the compatibility table for reference
- Fixed a bug in build_macos.dart where it was incorrectly using iOS flavor configuration instead of macOS

-----------------

Given the scenario from #2566, a user will see the following when running patrol build or patrol test now when the compatibility check fails:

Screenshot 2025-04-14 at 10 49 13 PM

### DIFF
--- a/packages/patrol_cli/lib/src/compatibility_checker/compatibility_checker.dart
+++ b/packages/patrol_cli/lib/src/compatibility_checker/compatibility_checker.dart
@@ -28,6 +28,7 @@ class CompatibilityChecker {
   final DisposeScope _disposeScope;
   final Logger _logger;
 
+  /// Checks if the version compatibility and throws an error if incompatible
   Future<void> checkVersionsCompatibility({
     required FlutterCommand flutterCommand,
     required TargetPlatform targetPlatform,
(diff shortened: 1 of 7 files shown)


packages/patrol_cli/lib/src/compatibility_checker/compatibility_checker.dart
```diff
@@ -86,10 +87,75 @@ class CompatibilityChecker {
...
+      final isCompatible = areVersionsCompatible(cliVersion, packageVersion);
+      if (!isCompatible) {
+        // Find the maximum compatible CLI version for this patrol version
+        final maxCliVersion = getMaxCompatibleCliVersion(packageVersion);
+
+        if (maxCliVersion != null) {
+          _logger.warn(
+            '''
+Patrol version $packageVersion defined in your project is not compatible with patrol_cli version $cliVersion.
+This will cause issues when running 'patrol test'.
```
piotruela, line 133: Whenever we introduce a change to patrol/patrol_cli that makes it incompatible with previous version, it's usually because we add a new field or method to `patrol` and we need `patrol_cli` be compatible with that change. Otherwise, both commands (`patrol test` and `patrol build` which is not running the tests, but preparing/building the app and tests) will most probably fail.
Therefore, I think we should change this warning to say `This will cause issues when running the tests` or maybe even `throwToolExit` the same way as we do for `patrol test`.

packages/patrol_cli/lib/src/compatibility_checker/compatibility_checker.dart
```diff
@@ -86,10 +87,75 @@ class CompatibilityChecker {
...
+
+To resolve this issue, you can:
+1. Downgrade patrol_cli to a compatible version by running: 
+   ${lightCyan.wrap('dart pub global activate patrol_cli $maxCliVersion')}
+   
+2. Or upgrade both "patrol_cli" and "patrol" dependencies to the latest versions.
+
+Check the compatibility table at: ${lightCyan.wrap('https://patrol.leancode.co/documentation/compatibility-table')}
```
piotruela, line 141: In the screenshot you've attached, this part lost the yellow color. Can we do something about it to make it more clear that this is the continuation of the warning message?

packages/patrol_cli/lib/src/compatibility_checker/compatibility_checker.dart
```diff
@@ -86,10 +87,75 @@ class CompatibilityChecker {
...
+      final maxCliVersion = getMaxCompatibleCliVersion(patrolVersion);
+
+      if (maxCliVersion != null) {
+        throwToolExit(
+          'Patrol version $patrolVersion defined in the project is not compatible with patrol_cli version $cliVersion\n'
+          'To resolve this issue, you can:\n'
+          '1. Downgrade patrol_cli to a compatible version by running: \n'
+          '   ${lightCyan.wrap('dart pub global activate patrol_cli $maxCliVersion')}\n'
+          '2. Or upgrade both "patrol_cli" and "patrol" dependency in project to the latest versions.\n\n'
+          'Check the compatibility table at: ${lightCyan.wrap('https://patrol.leancode.co/documentation/compatibility-table')}',
```
piotruela, line 100: Ditto (here red color is lost)

packages/patrol_cli/lib/src/compatibility_checker/compatibility_checker.dart
```diff
@@ -86,10 +87,75 @@ class CompatibilityChecker {
...
+This will cause issues when running 'patrol test'.
+
+To resolve this issue, you can:
+1. Downgrade patrol_cli to a compatible version by running: 
+   ${lightCyan.wrap('dart pub global activate patrol_cli $maxCliVersion')}
+   
+2. Or upgrade both "patrol_cli" and "patrol" dependencies to the latest versions.
+
+Check the compatibility table at: ${lightCyan.wrap('https://patrol.leancode.co/documentation/compatibility-table')}
+''',
```
piotruela, line 142: Maybe we could extract error/warning message to a common method to have something like:
```dart
String _incompatibilityMessage(
  String packageVersion,
  String? additionalInfo,
  String cliVersion,
  String? resolveSteps,
) {
  return '''
Patrol version $packageVersion defined in your project is not compatible with patrol_cli version $cliVersion.
$additionalInfo

${resolveSteps != null ? 'To resolve this issue, you can:\n$resolveSteps' : ''}

Check the compatibility table at: ${lightCyan.wrap('https://patrol.leancode.co/documentation/compatibility-table')}
''';
}
```
This way we will prevent duplicated parts of text

piotruela (commented):
Thanks for (yet another) contibution! I left some comments.

packages/patrol_cli/lib/src/compatibility_checker/compatibility_checker.dart
AnthonyDadeWT: Ok, thanks for the info. I will make change it to throw a ToolExit

packages/patrol_cli/lib/src/compatibility_checker/compatibility_checker.dart
AnthonyDadeWT: I've changed this to throwToolExit given your above comment. I've updated the screenshots in the PR description that show the new error messaging.

AnthonyDadeWT:
@piotruela Thank you for your review! I have added new screenshots of the error message shown to users when compatibility check fails on patrol build and test. I have also made the error messaging the same for both now.

piotruela (approved):
LGTM! 👏

PR #2608, opened 2025-04-18
Introduce native swipe back mimic

papeye:
Adds `$.native.swipeBack()` for easy to use testing of navigation with gestures. 

Motivation:
I've stumbled upon necessity to do swipe back gesture in patrol tests in my current project. Necessity arose for iOS because on Android we can always do `pressBack()` but there is no corresponding method for iOS which is somewhat limiting patrol use-cases because requires introduction of back arrow, which although conventional might not be sth the decision maker wants in their app. Even with back button available it might still be handy to be able to verify if navigation with gestures work as intended.

Introduced method is supported for all current patrol compatible platforms (android, iOS, macOS).  

Please let me know if using already defined  `swipe()` is ok or I should introduce a separate `swipeBack` request to Automator Servers. 

Current status: Bumping API version on emulator.wtf caused one of the tests `mock_location_test.dart` to fail. Looking into it now...

### DIFF
--- a/dev/e2e_app/integration_test/swipe_back_test.dart
+++ b/dev/e2e_app/integration_test/swipe_back_test.dart
@@ -1,0 +1,18 @@
+import 'common.dart';
+
+void main() {
+  patrol(
+    'performs swipe back gesture ',
+    ($) async {
+      await createApp($);
+
+      final openLoadingScreenButton = $(find.text('Open loading screen'));
+
+      await openLoadingScreenButton.scrollTo().tap();
+
+      await $.native.swipeBack(dy: 0.6);
+
+      await openLoadingScreenButton.waitUntilExists();
+    },
+  );
+}
--- a/packages/patrol/lib/src/native/native_automator.dart
+++ b/packages/patrol/lib/src/native/native_automator.dart
@@ -827,7 +827,7 @@ class NativeAutomator {
   ///
   /// Example usage:
   /// ```dart
-  /// await tester.swipeBack(height: 0.2); // Swipe back at 1/5 height of the screen
+  /// await tester.swipeBack(dy: 0.8); // Swipe back at 1/5 height of the screen
   /// await tester.swipeBack(); // Swipe back at the center of the screen
   /// ```
   Future<void> swipeBack({double dy = 0.5, String? appId}) {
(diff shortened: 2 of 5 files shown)


packages/patrol/lib/src/native/native_automator.dart
```diff
@@ -810,6 +810,43 @@ class NativeAutomator {
     );
   }
 
+  /// Mimics the swipe back (left to right) gesture.
+  ///
+  /// [height] determines the vertical position of the swipe, as a fraction
+  /// of the screen height (0.0 is the bottom, 1.0 is the top). It must be
```
piotruela, line 816: > 0.0 is the bottom

I think this might be counterintuitive. Axis in Flutter have value of (0,0) for top left corner. I think we should stick with that convention
image

dev/e2e_app/integration_test/swipe_back_test.dart
```diff
@@ -0,0 +1,14 @@
...
+void main() {
+  patrol(
+    'performs swipe back gesture ',
+    ($) async {
+      await createApp($);
+
+      await $(find.text('Open loading screen')).scrollTo().tap();
+
+      await $.native.swipeBack(height: 0.2);
+    },
```
piotruela, line 12: We need something in that test to verify if `swipeBack` worked

papeye: Agree 💯 I just removed this in the last commit to see if then the CI tests pass. Navigation gestures need to be turned on for android in settings and I'll investigate if that's possible to do for CI

packages/patrol/lib/src/native/native_automator.dart
papeye: Makes total sense. I wasn't sure which way to go. Thanks for the suggestion ☺️ I'll make the change  🚀

piotruela:
Looks like `mock_location_test` fails because we bumped API version of the emulators to 33 (which we needed to enable gesture). I'll look into it

papeye:
> Looks like `mock_location_test` fails because we bumped API version of the emulators to 33 (which we needed to enable gesture). I'll look into it

thank you <3 This test failing is weird - I'm able to make tests pass on 2 out of 3 failing devices by increasing 'cool down' times after setting mock location in the test, but it seems to behave somewhat undeterministicly anyway

pdenert (commented):
Looks good. I think that we can skip mock_location_test fail, because, we have to fix it anyway. 
This PR needs only changelog entry in patrol package

pdenert (approved):
LGTM

PR #2613, opened 2025-04-25
fix: Fix gathering exceptions in patrol develop

piotruela:
Follow up to #2362.
Turned out that when running `patrol develop` we cannot overwrite `FlutterError.onError` as the framework is then throwing this:
```
          'package:flutter_test/src/binding.dart': Failed assertion: line 1018 pos 9: '_pendingExceptionDetails != null': A test overrode FlutterError.onError but either failed to return it to its original state, or had unexpected additional errors that it could not handle. Typically, this is caused by using expect() before restoring FlutterError.onError.
          dart:core-patch/errors_patch.dart 63:4        _AssertionError._doThrowNew
          dart:core-patch/errors_patch.dart 45:5        _AssertionError._throwNew
          package:flutter_test/src/binding.dart 1018:9  TestWidgetsFlutterBinding._runTest.handleUncaughtError
          package:flutter_test/src/binding.dart 1032:9  TestWidgetsFlutterBinding._runTest.<fn>
          dart:async/zone.dart 1178:14                  _Zone._processUncaughtError
          dart:async/zone.dart 1399:5                   _CustomZone.handleUncaughtError
          dart:async/future_impl.dart 870:24            Future._propagateToListeners
          dart:async/future_impl.dart 730:5             Future._completeError
```
But we also don't need to overwrite default behaviour in develop mode as we are not reporting it back to native side anyway

### DIFF
--- a/packages/patrol/lib/src/binding.dart
+++ b/packages/patrol/lib/src/binding.dart
@@ -16,6 +16,8 @@ import 'constants.dart' as constants;
 
 const _success = 'success';
 
+bool get _isDevelopMode => constants.hotRestartEnabled;
+
 void _defaultPrintLogger(String message) {
   // TODO: Use a logger instead of print
   // ignore: avoid_print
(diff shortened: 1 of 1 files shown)


pdenert (approved):
LGTM

PR #2642, opened 2025-06-06
Add feature guide for taking photos using camera in Patrol framework

Kendru98:


docs/feature-guide/take-a-photo.mdx
```diff
@@ -0,0 +1,76 @@
+---
+title: Take photo using camera
+---
+
+We have added a new feature to take a photo using the Android and iOS camera with Patrol. Due to many differences between devices, this method will not work on 100% of devices but should work on most of them.
```
jBorkowska: Let's not write "new feature", it's not a news post, it will be outdated quickly. Also the info in second part of this line can be in Info block

docs/feature-guide/take-a-photo.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+
+## How it works
+This method does two actions:
+1. Tap on shutter button
+2. Tap on confirm button
+
+The table below shows the native selectors that the `takeCameraPhoto()` method uses internally for each platform:
+
+| Platform | Shutter Button | Confirm Button |
```
jBorkowska, line 18: Is this table necessary? You can look it up in the code

Kendru98: Looking at the code is not easy, you need to go to both automators as user, need to know how patrol is implemented etc.

pdenert: I think that such tables and other UI enhancements makes docs nicer

Issue #2317, opened 2024-08-19
Android-BrowserStack test execution failures due to "PatrolAppServiceClientException: Invalid response 503"


BanuToros:
### Steps to reproduce

1- Execute Patrol tests on emulators locally (Successfull)
2- Execute [bs_android](https://github.com/leancodepl/mobile-tools/blob/master/bin/bs_android) scripts to run Patrol tests on BrowserStack App Automate
3- Expect to see that Patrol "Test starting" screen is displayed, the app is launched and all the 21 tests are ran successfully

### Actual results

1- Patrol "Test starting" is not displayed
2- The app is not launching (only an orange screen is displayed which belongs to the app)
3- Initialisation error is returned.

https://github.com/user-attachments/assets/6d955201-d113-4b22-b943-c8a3abf7d9ec

### Logs

<details>
<summary>Logs</summary>

```
<INSTRUMENTATION_STATUS: class=com.easyjet.travel.android.MainActivityTest
INSTRUMENTATION_STATUS: current=1
INSTRUMENTATION_STATUS: id=AndroidJUnitRunner
INSTRUMENTATION_STATUS: numtests=1
INSTRUMENTATION_STATUS: stream=
com.easyjet.travel.android.MainActivityTest:
INSTRUMENTATION_STATUS: test=initializationError
INSTRUMENTATION_STATUS_CODE: 1
INSTRUMENTATION_STATUS: class=com.easyjet.travel.android.MainActivityTest
INSTRUMENTATION_STATUS: current=1
INSTRUMENTATION_STATUS: id=AndroidJUnitRunner
INSTRUMENTATION_STATUS: numtests=1
INSTRUMENTATION_STATUS: stack=java.lang.RuntimeException: pl.leancode.patrol.contracts.PatrolAppServiceClientException: Invalid response 503, <!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta
      name="viewport"
      content="width=device-width, height=device-height, initial-scale=1.0, minimum-scale=1.0"
    />
    <title>Page not found</title>
    <style>
      * {
        margin: 0;
        padding: 0;
        outline: 0;
        box-sizing: border-box;
      }

      a {
        color: #0070f0;
      }

      html {
        background-color: #e9e7ea;
        color: #333333;
      }

      h1 {
        font-size: 26px;
        line-height: 32px;
        font-weight: 400;
        text-align: center;
        padding: 30px 20px;
[...]

PR #2645, opened 2025-06-09
Use HttpMultiServer and Ktor for handling raw HTTP requests

mateuszwojtczak:
We are using raw HTTP instead of HTTPS by our localhost communication. Because of that, we sometimes have trouble with OkHttp blocking our requests due to [Android clear-text traffic policies](https://developer.android.com/privacy-and-security/security-config#CleartextTrafficPermitted).

To fix that issue on some device farms (e.g. Browserstack), this PR uses Ktor for raw HTTP communication. It also introduces HttpMultiServer package for cases where there are multiple network interfaces (e.g. BrowserStack again 🤣 ) and we don't know which to arbitrarily choose.

The changes in this PR are non-breaking for consumer projects. They’re either internal to Patrol or don’t require any SDK / Gradle / pubspec updates on the app side.


jBorkowska:
Will this close this one? https://github.com/leancodepl/patrol/issues/2317

piotruela (commented):
Can you do a small exploration to check whether changing/bumping those dependencies could be a breaking (or blocking) change for any projects? (E.g. could it force an SDK bump?) And publish a summary in the PR description

pdenert (approved):

PR #2649, opened 2025-06-13
Revert dynamic ports

piotruela:
This PR reverts the changes introduced in #2527. 
Due to https://github.com/flutter/flutter/pull/161828 released in Flutter 3.32.0, platform plugins are being removed when running flutter commands like `flutter build` in release mode. This prevents us from passing dynamically selected port via native side + ffi.
Until we will find a way to not be excluded from the build (while still be included as dev_dependency) we need to remove that change from `master` branch as this is currently blocking any releases of the stable version


mateuszwojtczak (approved):
LGTM

pdenert (approved):
LGTM

PR #2651, opened 2025-06-16
docs: Add comprehensive documentation for Native Automation 2.0 (nati…

Kendru98:


docs/documentation/native/native2.mdx
```diff
@@ -0,0 +1,164 @@
+---
+title: Native Automation 2.0 (native2)
+---
+
+Patrol's **native2** introduces a unified approach to native mobile automation, providing cross-platform selectors and methods that work seamlessly across both iOS and Android platforms. This evolution from the original native automation API simplifies test writing while providing more powerful selector capabilities.
```
jBorkowska: It's not quite true - actually we separate selector classes to be platform-specific

jBorkowska (changes requested):
Ehh, this doc is not reflecting the changes we made. especially it doesn't mention why we did this and why this approach is better and how to use it at its full power

docs/documentation/native/native2.mdx
```diff
@@ -0,0 +1,115 @@
...
+<Info>
+Native2 is available starting from Patrol version `3.6.0`.
+</Info>
+
+## What is native2?
+
+Native2 was created to address fundamental limitations in the original native automation approach. The original native API was primarily designed for Android, and attempts to make a single `Selector` work across both platforms proved problematic because:
+
+- **Platform-specific identifiers**: iOS and Android use different identifier systems (Android's `resourceName` vs iOS's `identifier`)
+- **Flaky text selectors**: Text-based selectors can be unreliable as they may change with localization or UI updates
```
jBorkowska: Well, it's still a problem, native2 doesn't solve this

docs/documentation/native/native2.mdx
```diff
@@ -0,0 +1,115 @@
...
+- **Platform incompatibility**: A single selector approach couldn't effectively handle the fundamental differences between iOS and Android element identification
```
jBorkowska: It's the same as first point, but told differently. I would merge both points into one

docs/documentation/native/native2.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+Native2 provides a **single command** that can work with both Android and iOS selectors, eliminating the need to write separate platform-specific code. Instead of maintaining different methods for each platform, you can now use one method that accepts platform-specific selectors that acknowledge the inherent differences between platforms.
```
jBorkowska: This paragraph suggests, that we had separate methods for Android and iOS in old native, which is not true. There's still one method, the difference is in the selector object.

docs/documentation/native/native2.mdx
```diff
@@ -0,0 +1,115 @@
...
+  ),
+);
+```
+
+### Specifying App ID (iOS)
+
+When working with iOS, you may need to specify the `appId` parameter to interact with elements in specific applications. This is particularly useful when your test needs to interact with system apps like Safari, Settings, or other third-party applications.
+
+<Info>
+The `appId` parameter is primarily needed for iOS. On Android, the app context is usually handled automatically.
```
jBorkowska: I think that appId parameter is ignored on Android - if that's true, it's worth writing about it here

docs/documentation/native/native2.mdx
```diff
@@ -0,0 +1,115 @@
...
+  await $.native.tap(Selector(label: 'Take Picture'));
+}
+```
+
+### With native2
+
+Native2 provides a single method that works across both platforms:
+
+```dart
+// After: Single method with platform-specific selectors
```
jBorkowska: method -> method call

docs/documentation/native/native2.mdx
```diff
@@ -0,0 +1,115 @@
...
+await $.native2.tap(
+  NativeSelector(
+    android: AndroidSelector(
+      resourceName: 'com.android.camera2:id/shutter_button',
+    ),
+    ios: IOSSelector(label: 'Take Picture'),
+  ),
+);
+
+// Text input with unified method
```
jBorkowska: method -> selector

docs/documentation/native/native2.mdx
```diff
@@ -0,0 +1,115 @@
...
+      identifier: 'emailField',
+    ),
+  ),
+  text: 'john@example.com',
+);
+```
+
+### Text Input Operations
+
+```dart
```
jBorkowska, line 50: Does this example show something different than the one above?

jBorkowska (changes requested):
Please replace all occurences of native2 to `native2`

docs/documentation/native/native2.mdx
Kendru98: A little does, now you can  easily use 2 different resourceId and resourceName in one command, with native you was not able to do that, you need Ifs or if's to do that.

docs/documentation/native/native2.mdx
Kendru98: It is tricky because in many cases when you want to rely on resourceId/Name what is best practice I would say. You need to use 2 different native Selectors and commands.

docs/documentation/native/native2.mdx
Kendru98: Very similar, will remove :D

docs/documentation/native/native2.mdx
jBorkowska: but still you have to use text-based selectors - it suggests that you don't have to use text selectors any more

docs/documentation/native/native2.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+<Info>
+Native2 is available starting from Patrol version `3.6.0`.
+</Info>
+
+## What is `native2`?
+
+Native2 was created to address fundamental limitations in the original native automation approach. The original native API was primarily designed for Android, and attempts to make a single `Selector` work across both platforms proved problematic because:
+
+- **Platform-specific identifiers**: iOS and Android use different identifier systems (Android's `resourceName` vs iOS's `identifier`) and a single selector approach couldn't effectively handle the fundamental differences between iOS and Android element identification
```
jBorkowska: remove list, make it just a part of a sentence, maybe bolded. It looks weird when it's only one point

jBorkowska: Also the wording here is a bit strange - what's an identifier system? Did you mean selectors arguments? It looks here like only "Android's `resourceName` vs iOS's `identifier`" is different. I would ditch the word "identifier" and use selector instead - it's quite universal name used in many frameworks and doesn't indicate our `Selector`

docs/documentation/native/native2.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+Native2 provides **platform-specific selectors** that work with both Android and iOS, giving you more accurate platform-specific selectors instead of one shared selector.
```
jBorkowska: This sentence is much better than the one above and I think it tells exactly what the previous sentence was trying to say

docs/documentation/native/native2.mdx
```diff
@@ -0,0 +1,100 @@
...
+  ),
+);
+```
+
+### Specifying App ID (iOS)
+
+When working with iOS, you may need to specify the `appId` parameter to interact with elements in specific applications. This is particularly useful when your test needs to interact with system apps like Safari, Settings, or other third-party applications.
+
+<Info>
+The `appId` parameter is required for iOS. On Android, it will be ignored.
```
jBorkowska: It's not required, it's used

jBorkowska (changes requested):

jBorkowska:
Also make all native2 to be code (I mean bacticks), there are a few left

docs/documentation/native/native2.mdx
Kendru98: I think this sentence above is okay, it gives more context to the user. I rephrase it a little check now pls

jBorkowska (approved):

pdenert (approved):
LGTM

PR #2656, opened 2025-06-23
chore: Bump API version of emulators in android test workflow

piotruela:


dev/e2e_app/integration_test/mock_location_test.dart
pdenert, line 1: Mock location fails on android emulator. I guess the issue is elsewhere then in the length of waiting time

dev/e2e_app/emulatorwtf.yaml
```diff
@@ -29,13 +29,13 @@ all_devices:
...
+      version: 33
       gpu: auto
     - model: NexusLowRes
-      version: 31
+      version: 33
       gpu: auto
     - model: Tablet10
-      version: 31
+      version: 33
       gpu: auto
```
Kendru98: Maybe mix API lvls a little? for example 34/33/32?

PR #2659, opened 2025-06-25
Feat/handle more locales for native methods

Kendru98:


packages/patrol/android/src/main/kotlin/pl/leancode/patrol/LocalizationHelper.kt
```diff
@@ -0,0 +1,42 @@
+package pl.leancode.patrol
+
+import android.content.Context
+import android.os.Build
+
+object LocalizationHelper {
```
jBorkowska: Can we remove `Helper`?

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
```diff
@@ -1,4 +1,4 @@
-#if PATROL_ENABLED && os(iOS)
+#if PATROL_ENABLED  && os(iOS)
```
jBorkowska: hm?

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
```diff
@@ -389,17 +389,33 @@
...
           }
         #else
-          self.preferences.descendants(matching: .any)["Display & Brightness"].firstMatch.tap()
-          self.preferences.descendants(matching: .any)["Dark"].firstMatch.tap()
+          // Use system locale directly instead of calling getLocale()
+          let locale = Locale.current.languageCode ?? "en"
+          let localizedStrings = self.getLocalizedStrings(for: locale)
+
+          self.preferences.descendants(matching: .any)[
+            localizedStrings["display_brightness"] ?? "Display & Brightness"
```
jBorkowska: Why do we have this fallback here? We have "en" fallback above already, is this one necessary?

jBorkowska: This happens in many places

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
Kendru98: Additional space - removed

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
Kendru98: Yes, in the case that File will be not found - but probably this will not happen so we can remove it

Kendru98: @piotruela @pdenert On Android side it was easy, but on iOS should I remove fallback and use "!"?

packages/patrol/android/src/main/res/values-de/strings_de.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!-- Airplane mode toggle -->
+    <string name="airplane_mode">Flugmodus</string>
+    
+    <!-- Location settings -->
+    <string name="use_location">Standort verwenden</string>
+</resources>
```
piotruela: Add trailing newline

piotruela: And same for all other files with translation strings

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
```diff
@@ -912,6 +1003,34 @@
       element.typeText(delete + data)
     }
 
+    private func getLocalizedStrings(for languageCode: String) -> [String: String] {
```
piotruela: I'd say that method should take only string key as a parameter. The `languageCode` can be accessed inside of it right? 
If english string won't be found (which I guess is highly unlikely, we can then throw an exception from this method

packages/patrol/darwin/Resources/de.lproj/Localizable.strings
```diff
@@ -0,0 +1,23 @@
...
+/* Settings navigation */
+"developer" = "Entwickler";
+"dark_appearance" = "Dunkles Erscheinungsbild";
+"display_brightness" = "Anzeige & Helligkeit";
+"dark" = "Dunkel";
+"light" = "Hell";
+
+/* Location precision */
+"precise_off" = "Genau: Aus";
+"precise_on" = "Genau: An";
```
piotruela: Add trailing newline

piotruela: Same in all other `Localizable.strings`

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
Kendru98: Should be solved check current implementation

packages/patrol/darwin/Classes/AutomatorServer/Localization.swift
```diff
@@ -0,0 +1,108 @@
...
+        )
+        // Fallback to English if target language specific file is not found
+        if targetLanguage != "en" {
+          return try getLocalizedStringForLanguage(key: key, language: "en", bundle: bundle)
+        } else {
+          // If even English is not found or this is already 'en', we return the key itself
+          Logger.shared.e(
+            "Localizable.strings file not found for language: \(targetLanguage) and no English fallback."
+          )
+          return key  // Fallback: return the key itself
```
piotruela: What's the point in returning key? If fallback string is not found we are 100% sure that the method will fail so we can just throw an exception already

packages/patrol/darwin/Classes/AutomatorServer/Localization.swift
```diff
@@ -0,0 +1,108 @@
...
+    } catch let error as LocalizationError {
+      Logger.shared.e(
+        "Localization error for key '\(key)' in language '\(targetLanguage)': \(error.localizedDescription)"
+      )
+      return key  // Fallback: return the key itself on error
+    } catch {
+      Logger.shared.e(
+        "An unexpected error occurred during localization for key '\(key)': \(error.localizedDescription)"
+      )
+      return key  // Fallback: return the key itself on unexpected error
```
piotruela: Ditto. Let's just throw from this method and handle it where we use it

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
```diff
@@ -897,6 +935,24 @@
       }
     }
 
+    func getLocale() throws -> GetLocaleResponse {
+      return runAction("getting device locale") {
+        let locale = Locale.current
+        let languageCode = locale.languageCode ?? "en"
+        let regionCode = locale.regionCode ?? "US"
```
piotruela: Why would we return `en` and `US` as the default values. Maybe we should just return `null` when nothing was found instead?

packages/patrol/darwin/Classes/AutomatorServer/Localization.swift
```diff
@@ -0,0 +1,108 @@
+import Foundation
+import os
+
+class Localization {
+
+  /**
+   * Gets the current device locale
+   */
+  private static func getDeviceLocale() -> String {
+    return Locale.current.languageCode ?? "en"
```
piotruela: ditto

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
Kendru98: Please check, I made changes

CONTRIBUTING.md
```diff
@@ -77,3 +77,19 @@ Note: On iOS, during development you should remove `#if PATROL_ENABLED` from `Au
...
+2. Add a new language folder with the name of the language or use existing one.
+3. Add a new `Localizable.strings` or add strings to the existing file.
+4. Add the new language to the `getLocalizedStrings` function in `IOSAutomator.swift` file.
+5. If you added new strings, you need to find all functions where these strings are used and add them to the function. for example "labels" in allowPermissionOnce() in IOSAutomator.swift file.
+
+### Android
+
+1. Go to `packages/patrol/android/src/main/res/values-<language>/` folder.
+2. Add a new `strings.xml` with language folder name or add strings to the existing file.
+3. If you added new string, you need to add it to proper function in Automator.kt, for example look on toggleAirplaneMode() function.
```
gogolon:
```suggestion
3. If you added a new string, you need to add it to proper function in Automator.kt, for example look at the toggleAirplaneMode() function.
```

dev/e2e_app/integration_test/permissions/notifications_test.dart
```diff
@@ -87,5 +89,7 @@ void main() {
 
       await $('Tapped notification with ID: 1').waitUntilVisible();
     },
+    // Test fails on physical device and do not throw an error cause of timeout
```
gogolon:
```suggestion
    // Test fails on physical device and does not throw an error cause of timeout
```

gogolon: is this why the tag below is commented?

dev/e2e_app/integration_test/service_airplane_mode_test.dart
```diff
@@ -1,12 +1,23 @@
...
-    await $.native.enableAirplaneMode();
-  });
+  patrol(
+    'disables and enables airplane mode twice',
+    ($) async {
+      await createApp($);
+      if (await $.native.isVirtualDevice() && Platform.isIOS) {
+        $.log(
+          'Test will be skipped because of iOS simulator limitations',
+        );
```
gogolon: early return?

dev/e2e_app/lib/permissions_screen.dart
```diff
@@ -120,10 +151,25 @@ class _PermissionTile extends StatelessWidget {
   final bool granted;
   final VoidCallback onTap;
 
+  Key _getRequestButtonKey(String permissionName) {
+    switch (permissionName.toLowerCase()) {
```
gogolon: switch expression

dev/e2e_app/lib/permissions_screen.dart
```diff
@@ -120,10 +151,25 @@ class _PermissionTile extends StatelessWidget {
...
+      default:
+        return Key('request${permissionName}PermissionButton');
+    }
+  }
+
   @override
   Widget build(BuildContext context) {
     return Padding(
-      key: Key(name.toLowerCase()),
+      key: key,
```
gogolon: why add a key here?

packages/patrol/android/src/main/kotlin/pl/leancode/patrol/Automator.kt
```diff
@@ -832,7 +833,7 @@ class Automator private constructor() {
         targetContext.startActivity(intent)
 
         var uiSelector = UiSelector()
-        uiSelector = uiSelector.text("Airplane mode")
+        uiSelector = uiSelector.text(Localization.getLocalizedString(targetContext, "airplane_mode"))
```
gogolon: instead of using keys directly, do

`import pl.leancode.patrol.R.string as s`

and then `s.airplane_mode`

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
```diff
@@ -725,8 +739,9 @@
         Logger.shared.i("tapping on notification which contains text \(substring)")
         if self.isVirtualDevice() && self.isPhone() {
           // For some weird reason, this works differently on Simulator
```
gogolon, line 741: why the additional `isPhone` check? what about iPads?

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
```diff
@@ -824,17 +862,18 @@
...
       }
 
       let alerts = self.springboard.alerts
-      let button = alerts.buttons["Precise: On"]
+      let preciseOn = try Localization.getLocalizedString(key: "precise_on")
+      let button = alerts.buttons[preciseOn]
       let exists = button.waitForExistence(timeout: self.timeout)
-
       return exists
+
```
gogolon: revert?

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
```diff
@@ -843,14 +882,16 @@
         return
       }
 
-      if !isFineLocationEnabled() {
+      if try !isFineLocationEnabled() {
         Logger.shared.i("Coarse location is already enabled")
```
gogolon, line 885: is this logic correct?

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
```diff
@@ -1023,14 +1064,16 @@
...
+      let ok = try Localization.getLocalizedString(key: "ok")
+      let labels = [ok]
 
       if let button = self.waitForAnyElement(
         elements: labels.map { systemAlerts.buttons[$0] },
         timeout: self.timeout
       ) {
         button.tap()
       }
+
```
gogolon: revert

packages/patrol/darwin/Classes/AutomatorServer/Automator/MacOSAutomator.swift
```diff
@@ -27,6 +27,12 @@
       return device.system
     }()
 
+    func getLocale() throws -> GetLocaleResponse {
+      try runAction("getLocale") {
+        throw PatrolError.methodNotImplemented("getLocale")
```
gogolon: todo?

packages/patrol/darwin/Classes/AutomatorServer/Localization.swift
```diff
@@ -0,0 +1,100 @@
...
+    // Try to load the localized strings file for the target language
+    if let path = bundle.path(
+      forResource: "Localizable", ofType: "strings", inDirectory: "\(targetLanguage).lproj")
+    {
+      return try getLocalizedStringForLanguage(
+        key: key, language: targetLanguage, bundle: bundle, path: path)
+    } else {
+      Logger.shared.i(
+        "Could not find Localizable.strings for \(targetLanguage), trying English strings as fallback"
+      )
```
gogolon: move this into the if below

packages/patrol/darwin/Classes/AutomatorServer/Localization.swift
```diff
@@ -0,0 +1,100 @@
...
+      throw LocalizationError.parsingFailed(
+        "Could not parse Localizable.strings for language: \(language)")
+    }
+
+    guard let localizedString = dictionary[key] else {
+      throw LocalizationError.keyNotFound(
+        "Key '\(key)' not found in Localizable.strings for language: \(language)")
+    }
+
+    Logger.shared.i("Loaded localized string for key '\(key)' in language: \(language)")
```
gogolon:
```suggestion
    Logger.shared.d("Loaded localized string for key '\(key)' in language: \(language)")
```

dev/e2e_app/integration_test/permissions/deny_many_permissions_twice_test.dart
```diff
@@ -1,65 +1,88 @@
...
-    await _requestAndDenyMicrophonePermission($);
-    await _requestAndDenyMicrophonePermission($);
+      await _requestAndDenyMicrophonePermission($);
+      await _requestAndDenyMicrophonePermission($);
 
-    await _requestAndDenyLocationPermission($);
-    await _requestAndDenyLocationPermission($);
-  });
+      await _requestAndDenyLocationPermission($);
+      await _requestAndDenyLocationPermission($);
```
gogolon: please add a comment explaining such checks need to be done twice

packages/patrol/darwin/Classes/AutomatorServer/Localization.swift
```diff
@@ -0,0 +1,100 @@
+import Foundation
+import os
+
+class Localization {
+
+  /**
+   * Gets the current device locale
+   */
+  private static func getDeviceLocale() -> String? {
+    return Locale.current.languageCode
```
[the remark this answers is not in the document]
Kendru98: don't think so that we want to support region-specific variations

dev/e2e_app/integration_test/permissions/notifications_test.dart
Kendru98: Yes

dev/e2e_app/lib/permissions_screen.dart
Kendru98: Need to indentify Tile with permission to check if permission was granted or denied

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
Kendru98: Not sure need to ask @piotruela, I just changed virtual device check

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
Kendru98: I think yes, this is small switcher on the localization permission on iOS, tests working correctly as I checked. I just need to add try to handle errors, because of changes related to add more locales

packages/patrol/darwin/Classes/AutomatorServer/Automator/MacOSAutomator.swift
Kendru98: not sure there are others 46 occurances of this, and there is no todo at all

CONTRIBUTING.md
```diff
@@ -77,3 +77,21 @@ Note: On iOS, during development you should remove `#if PATROL_ENABLED` from `Au
 ## Working with patrol_devtools_extension
 
 If you plan to use the local version of Patrol to test/modify the Patrol DevTools extension, you need to deploy it first. To do this, navigate to the `patrol_devtools_extension` folder and run the `./publish_to_patrol_extension` script.
+
+## Adding a new language or new strings to existing language
```
gogolon: please specify that it refers to handling native OS interaction

dev/e2e_app/integration_test/service_bluetooth_test.dart
```diff
@@ -1,12 +1,18 @@
...
   patrol('disables and enables bluetooth twice', ($) async {
     await createApp($);
-
-    await $.native.disableBluetooth();
-    await $.native.enableBluetooth();
-    await $.native.disableBluetooth();
-    await $.native.enableBluetooth();
+    // TODO: will be handled with tags in the future
+    if (await $.native.isVirtualDevice() && Platform.isIOS) {
+      $.log('Test will be skipped because of iOS simulator limitations');
```
gogolon, line 10: I'd say an early return would be preferred in such cases to avoid having the whole actual code of the test in an 'else'

dev/e2e_app/lib/permissions_screen.dart
gogolon: I should have been more specific: the key is already attached to the widget itself as it's assigned to `super.key` in the constructor - no need to assign it to the padding too

packages/patrol/android/src/main/res/values/strings_en.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!-- Airplane mode toggle -->
+    <string name="airplane_mode">Airplane mode</string>
+    
+    <!-- Location settings -->
+    <string name="use_location">Use location</string>
+</resources>
```
gogolon: missing newline at the end of the file

packages/patrol/darwin/Classes/AutomatorServer/Localization.swift
```diff
@@ -0,0 +1,108 @@
...
+class Localization {
+
+  /**
+   * Gets the current device language
+   */
+  private static func getDeviceLanguage() -> String? {
+    // Get the user's preferred language instead of system locale
+    if let preferredLanguage = Locale.preferredLanguages.first {
+      // Extract language code from language identifier (e.g., "en-US" -> "en")
+      return String(preferredLanguage.prefix(2))
```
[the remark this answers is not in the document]
gogolon: I think they are all of length 2 but the suggested solution seems more elegant (if it works ofc)

dev/e2e_app/lib/permissions_screen.dart
Kendru98: skill issue - removed

packages/patrol/android/src/main/kotlin/pl/leancode/patrol/Localization.kt
```diff
@@ -0,0 +1,47 @@
...
+        // Check if the detected language is supported
+        val supportedLanguages = setOf("en", "de", "fr", "pl")
+        if (language !in supportedLanguages) {
+            throw UnsupportedOperationException(
+                "Language '$language' is not supported. Supported languages: ${supportedLanguages.joinToString(", ")}"
+            )
+        }
+
+        return if (resourceId != 0) {
+            context.resources.getString(resourceId)
```
gogolon: I think we can just return this. The 0 check does not seem to make sense

CONTRIBUTING.md
```diff
@@ -77,3 +77,21 @@ Note: On iOS, during development you should remove `#if PATROL_ENABLED` from `Au
...
+
+## Adding a new language or new strings to existing language for native OS interactions
+
+### iOS
+
+1. Go to `packages/patrol/darwin/Resources/` folder.
+2. Add a new language folder with the name of the language or use existing one.
+3. Add a new `Localizable.strings` or add strings to the existing file.
+4. Add the new language to the `getLocalizedStrings` function in `IOSAutomator.swift` file.
+5. If you added new strings, you need to find all functions where these strings are used and add them to the function. for example "labels" in `allowPermissionOnce()` in IOSAutomator.swift file.
```
pdenert:
```suggestion
5. If you added new strings, you need to find all functions where these strings are used and add them to the function. For example "labels" in `allowPermissionOnce()` in IOSAutomator.swift file.
```

dev/e2e_app/integration_test/service_wifi_test.dart
```diff
@@ -1,12 +1,17 @@
...
 void main() {
   patrol('disables and enables wifi twice', ($) async {
     await createApp($);
-
-    await $.native.disableWifi();
-    await $.native.enableWifi();
-    await $.native.disableWifi();
-    await $.native.enableWifi();
+    if (await $.native.isVirtualDevice() && Platform.isIOS) {
+      $.log('Test will be skipped because it of iOS simulator limitations');
```
pdenert:
```suggestion
      $.log('The test will be skipped due to the limitations of the iOS simulator.');
```

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
piotruela: tbh, I don't know why is this `.isPhone()` check used here. I guess we might try removing it in some other PR to see if it's really necessary

PR #2653, opened 2025-06-17
Support inline and full-line comments in .patrol.env files

lockieRichter:
### Problem

Currently, `.patrol.env` files do not support comments. This makes it difficult to document the purpose of environment variables, provide context, or temporarily disable lines without deleting them.

### Solution

This PR adds support for both full-line and inline comments:

- Lines starting with `#` are now ignored completely.
- Inline comments are supported — any text following a `#` on the same line as a variable is stripped out.

### Example

```env
# This is the API key used in staging
API_KEY=abc123 # Replace with actual key

# This one is disabled for now
# DEBUG_MODE=true
```

### Benefits
- Improves readability and maintainability of .patrol.env files.
- Makes it easier for team members to understand or modify test configs.
- Compatible with existing .env file conventions.

PR #2669, opened 2025-07-05
fix: Support values with # in .patrol.env

piotruela:
Change introduced in #2653 disallowed having `#` in key-value pairs inside `.patrol.env` as everything after this character would be (sometimes incorrectly) treated as a comment.


pdenert (approved):
LGTM

Issue #2609, opened 2025-04-19
FormatException when wrong pattern is used for tags


vlazdra:
I managed to figure this out from the android logcat output where the exception occurred.

```bash
04-18 19:40:42.104  5600  5600 E flutter : [ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: Invalid argument(s): Invalid tag "5301". Tags must be (optionally hyphenated) Dart identifiers.
04-18 19:40:42.104  5600  5600 E flutter : #0      Metadata._validateTags (package:test_api/src/backend/metadata.dart:299:5)
04-18 19:40:42.104  5600  5600 E flutter : #1      new Metadata.parse (package:test_api/src/backend/metadata.dart:254:5)
04-18 19:40:42.104  5600  5600 E flutter : #2      Declarer.test (package:test_api/src/backend/declarer.dart:201:32)
04-18 19:40:42.104  5600  5600 E flutter : #3      test (package:flutter_test/src/test_compat.dart:186:13)
04-18 19:40:42.104  5600  5600 E flutter : #4      testWidgets (package:flutter_test/src/widget_tester.dart:175:5)
04-18 19:40:42.104  5600  5600 E flutter : #5      patrolTest (package:patrol/src/common.dart:108:3)
04-18 19:40:42.104  5600  5600 E flutter : #6      patrol (file:///Users/vlazdra/Documents/Projects/project/integration_test/utils/common.dart:56:3)
04-18 19:40:42.104  5600  5600 E flutter : #7      main (file:///Users/vlazdra/Documents/Projects/project/integration_test/features/deeplink_test.dart:15:3)
04-18 19:40:42.104  5600  5600 E flutter : #8      Declarer.group.<anonymous closure> (package:test_api/src/backend/declarer.dart:283:37)
04-18 19:40:42.104  5600  5600 E flutter : #9      _rootRun (dart:async/zone.dart:1525:13)
04-18 19:40:42.104  5600  5600 E flutter : #10     _CustomZone.run (dart:async/zone.dart:1422:19)
04-18 19:40:42.104  5600  5600 E flutter : #11     _runZoned (dart:async/zone.dart:2034:6)
04-18 19:40:42.104  5600  5600 E flutter : #12     runZoned (dart:async/zone.dart:1960:10)
04-18 19:40:42.104  5600  5600 E flutter : #13     Declarer.declare (package:test_api/src/backend/declarer.dart:183:7)
[...]

PR #2674, opened 2025-07-09
Fix patrol tags page; add warning about not number tags

Kendru98:
Resolving https://github.com/leancodepl/patrol/issues/2609, not sure if we want to handle just numbered tags.

### DIFF
--- a/docs/documentation/patrol-tags.mdx
+++ b/docs/documentation/patrol-tags.mdx
@@ -107,7 +107,6 @@ patrol test --tags='((login || payment) && (smoke || regression))'
 
 You can combine tag filtering with other Patrol CLI options:
 
-# Run specific test file with tag filtering
 ```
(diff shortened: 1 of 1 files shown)


docs/documentation/patrol-tags.mdx
```diff
@@ -81,6 +81,7 @@ Patrol supports complex tag expressions using logical operators:
 - `&&` - AND operator (run tests with both tags)
 - `!` - NOT operator (exclude tests with this tag)
 
+<Warning>Tag can't start or be just a number</Warning>
```
jBorkowska: let's write what actually a tag can be, as it's not the only condition. Maybe let's link: xhttps://pub.dev/packages/test#tagging-tests and cite: "Note that tags must be valid Dart identifiers, although they may also contain hyphens."

PR #2690, opened 2025-07-15
Bump leancode lint, min dart and patrol_log in patrol_cli and patrol

pdenert:


docs/documentation/compatibility-table.mdx
gogolon, line 1: this file seems to be duplicated. the second one is at `docs/compatibility-table.mdx`

packages/patrol/example/pubspec.yaml
```diff
@@ -36,12 +36,13 @@ dev_dependencies:
...
   flutter_test:
     sdk: flutter
-  leancode_lint: ^14.3.0
+  leancode_lint: ^17.0.0
   patrol:
     path: ..
   patrol_cli:
     path: ../../patrol_cli
 
+  test_api: any
```
gogolon: intentional?

packages/patrol/example/lib/firebase_options.dart
```diff
@@ -49,15 +49,15 @@ class DefaultFirebaseOptions {
     }
   }
 
-  static const FirebaseOptions android = FirebaseOptions(
+  static const android = FirebaseOptions(
```
gogolon: I see you applied dart fix here but this file is generated (by flutterfire configure) so I would say it should be exempt from lints

docs/documentation/compatibility-table.mdx
pdenert: Unfortunately we're forced to duplicating this, due to redirecting issue in docs.page

packages/patrol/example/pubspec.yaml
pdenert: Strange, it was added, by dart fix, but it doesn't required anymore

packages/patrol/example/lib/firebase_options.dart
gogolon: maybe there's a better way than adding an ignore in the file itself? if the file was ever regenerated, the ignore would be probably removed. I thought excluding in analysis options but I don't think it's possible to exclude from lints only there

gogolon (approved):

packages/patrol/example/lib/firebase_options.dart
pdenert: Agree, but as you yourself have pointed out: it's don't possible to exclude from lints only there

dev/e2e_app/integration_test/permissions/notifications_test.dart
```diff
@@ -3,89 +3,83 @@ import 'dart:io' as io;
...
-        // Android 14+ requires additional permission to schedule notifications.
-        // Workaround for conditionally granting permission.
-        final android14PermissionSelector = AndroidSelector(
-          text: 'Allow setting alarms and reminders',
-        );
-        final nativeViews = await $.native2.getNativeViews(
+    // Until we resolve the issue of invoking native methods without a
+    // selector intended for the platform on which we are running the test,
+    // we need to add this check.
+    if (io.Platform.isAndroid) {
```
gogolon, line 52: why not defaultTargetPlatform?

gogolon (approved):

Issue #2607, opened 2025-04-18
Keep the app running after last step in `patrol develop` test


piotruela:
### Use case

When I write a test using `patrol develop` I usually do it step by step. I'd like to see when I ended up after completing last step in my 'under construction' test. Currently after we execute the whole test app is closed and we see such screen (on Android):

<img width="352" alt="Image" src="https://github.com/user-attachments/assets/717a54c7-7480-4b67-9190-49a317e33be7" />

### Proposal

`patrol develop` should keep the app open after the last step is finished so I can see and decide what step I want to add next to the test code. We could achieve that by adding a step to `patrolTest` body that would be executed only if we are running test using `develop` command. We could report that we are stopped at that step to `patrol_log` so that the test execution will be reported as successful.

Kendru98:
> Currently after we execute the whole test app is closed and we see such screen (on Android):

Same thing for iOS

PR #2693, opened 2025-07-17
Feat/keep the app running after last step in patrol develop

Kendru98:
Resolve: #2607


packages/patrol/lib/src/common.dart
```diff
@@ -148,27 +148,25 @@ void patrolTest(
       );
       await callback(patrolTester);
 
-      // We need to silent this warning to avoid false positive
-      // avoid_redundant_argument_values
-      // ignore: prefer_const_declarations
-      final waitSeconds = const int.fromEnvironment('PATROL_WAIT');
-      final waitDuration = Duration(seconds: waitSeconds);
-
```
Kendru98, line 156: I think that this is outdated after we removed --wait, can you confirmed?

packages/patrol/lib/src/global_state.dart
```diff
@@ -34,3 +51,119 @@ String get currentTestIndividualName {
...
+}
+
+/// Checks if a root-level test is the last entry in the entire test file
+bool _isLastTestAtRootLevel(LiveTest currentTest) {
+  try {
+    // A root-level test should generally not stop the execution
+    // unless there's evidence it's truly the last test
+    // This is a conservative approach
+    return false;
+  } catch (err) {
```
konowrockis: Is there ever a chance `return false` throws? 😅

packages/patrol/lib/src/global_state.dart
```diff
@@ -34,3 +51,119 @@ String get currentTestIndividualName {
...
+  return true;
+}
+
+/// Checks if the current test is the last test in the given group
+bool _isLastTestInGroup(LiveTest currentTest, Group group) {
+  final allEntriesInGroup = <String>[];
+
+  // Get ALL entries (both tests and groups) in the group
+  for (final entry in group.entries) {
+    if (entry.name != 'patrol_test_explorer') {
```
konowrockis: `patrol_test_explorer` should be IMO extracted to some const or function `isInternalTestExplorer` or property on entry.

packages/patrol/lib/src/global_state.dart
```diff
@@ -34,3 +51,119 @@ String get currentTestIndividualName {
...
+      // Extract individual entry name by removing the group prefix
+      final fullName = entry.name;
+      final groupPrefix = '${group.name} ';
+      final individualEntryName = fullName.startsWith(groupPrefix)
```
konowrockis: Hmm, isn't it pretty much the same thing as [this function](https://github.com/leancodepl/patrol/blob/b5c7649cd8fcb5dffbb38e2cc9b4de98757a3af2/packages/patrol/lib/src/common.dart#L276)? 🤔
Shouldn't this be extracted and reused?

packages/patrol/lib/src/global_state.dart
```diff
@@ -34,3 +51,119 @@ String get currentTestIndividualName {
...
+
+  if (allEntriesInGroup.isEmpty) {
+    return false;
+  }
+
+  final currentTestName = currentTest.individualName;
+  final lastEntryName = allEntriesInGroup.last;
+
+  // The current test is last only if it's the very last entry (not just last test)
+  final isLast = currentTestName == lastEntryName;
```
konowrockis: I have no idea how patrol behaves in such cases but what would happen if two tests have the same name? 🤔

packages/patrol/lib/src/global_state.dart
Kendru98: Dont think so 🙈

packages/patrol/lib/src/global_state.dart
Kendru98: Good idea

packages/patrol/lib/src/global_state.dart
```diff
@@ -1,12 +1,45 @@
...
+const _patrolTestExplorerName = 'patrol_test_explorer';
+
+/// Returns true if the given test/entry name is the internal patrol test explorer.
+bool isInternalTestExplorer(String name) {
+  return name == _patrolTestExplorerName;
+}
+
+/// Returns true if the given entry is the internal patrol test explorer.
+bool isInternalTestExplorerEntry(GroupEntry entry) {
+  return entry.name == _patrolTestExplorerName;
```
konowrockis:
```suggestion
  return isInternalTestExplorer(entry.name);
```

packages/patrol/lib/src/global_state.dart
piotruela: When running `patrol test` command, tests are invoked by native side using by name (concatenated with groups names). So when two tests within the same group have the same name it will execute first test in that group twice (I checked it just now).

What we could do is, when making this comparison, take into account not only the test name but also the names of the groups in which the test is located. I think it’s quite a common situation to have the same test name in different groups, so it seems to me that we should handle such cases.

As for having the same name within the same group, I’ve noticed that the Flutter plugin for running tests also doesn’t handle this well — it only executes one of the two tests — so I think we can consider this a rare edge case.

packages/patrol/lib/src/common.dart
```diff
@@ -148,27 +148,28 @@ void patrolTest(
...
       if (debugDefaultTargetPlatformOverride !=
           patrolBinding.workaroundDebugDefaultTargetPlatformOverride) {
         debugDefaultTargetPlatformOverride =
             patrolBinding.workaroundDebugDefaultTargetPlatformOverride;
       }
 
-      if (waitDuration > Duration.zero) {
-        final stopwatch = Stopwatch()..start();
+      // In develop mode (hot restart enabled), wait indefinitely to allow user interaction
+      // and avoid the "Test Completed" screen, but only after the last test in the group
```
gogolon: this comment seems to be somewhat duplicated in ln 170

packages/patrol/lib/src/common.dart
```diff
@@ -148,27 +148,28 @@ void patrolTest(
...
+        // If test fails this code will not be executed
+        patrolLog.log(
+          LogEntry(
+            message:
+                'All tests were executed. Press "r" to start again or "q" to quit',
+          ),
+        );
+        // Wait indefinitely in develop mode after the last test
         await Future.doWhile(() async {
           await widgetTester.pump();
```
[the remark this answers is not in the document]
gogolon: could be a while(true) loop?

packages/patrol/lib/src/global_state.dart
```diff
@@ -34,3 +67,96 @@ String get currentTestIndividualName {
 bool get isCurrentTestPassing {
   return Invoker.current!.liveTest.state.result.isPassing;
 }
+
+/// Returns whether the current test is the last test in its immediate group.
+bool get isCurrentTestLastInGroup {
+  final invoker = Invoker.current!;
+  final currentTest = invoker.liveTest;
```
gogolon: combine into 1 line

packages/patrol/lib/src/global_state.dart
```diff
@@ -34,3 +67,96 @@ String get currentTestIndividualName {
...
+      // This is a higher-level group - check if the child group is last entry
+      final childGroup = currentTest.groups[groupIndex + 1];
+      if (!_isLastEntryInGroup(childGroup, currentGroup)) {
+        return false;
+      }
+    }
+  }
+
+  // If we get here, the test is last at every level of the hierarchy
+
```
gogolon: excessivve newline

packages/patrol/lib/src/global_state.dart
```diff
@@ -34,3 +67,96 @@ String get currentTestIndividualName {
...
+  if (allEntriesInGroup.isEmpty) {
+    return false;
+  }
+
+  final currentTestName = currentTest.individualName;
+  final lastEntryName = allEntriesInGroup.last;
+
+  // The current test is last only if it's the very last entry (not just last test)
+  final isLast = currentTestName == lastEntryName;
```
gogolon: maybe return it immediately?

packages/patrol/lib/src/common.dart
```diff
@@ -148,29 +148,27 @@ void patrolTest(
...
+        while (true) {
           await widgetTester.pump();
-          if (stopwatch.elapsed > waitDuration) {
-            stopwatch.stop();
-            return false;
-          }
-
-          return true;
-        });
+          await Future.delayed(const Duration(milliseconds: 10), () {});
```
gogolon:
```suggestion
          await Future<void>.delayed(const Duration(milliseconds: 10));
```

no lint warning for not passing the generic type?

Kendru98: If add ,() {}); there is no warning :D

Kendru98: but should add <void> will fix

piotruela (approved):

Issue #2393, opened 2024-11-05
`disableDarkMode` and `enableDarkMode` don't work on iOS 18 Simulator.


piotruela:
I recently ran the `service_dark_mode_test.dart` test from `e2e_app` project in the repository on the iOS 18 Simulator. The test fails at the stage where it attempts to click on the 'Developer' option in the settings.
<img width="670" alt="image" src="https://github.com/user-attachments/assets/1ce66002-9fef-4960-9d09-3dfd0ef44bf8">

Probably something has changed in how this element is described in iOS 18.

### Patrol version

patrol: 3.12.0
patrol_cli: 3.3.0

Friesenjung:
+1 this is really annoying. Currently I cannot run our Patrol Test Suite due to this issue.

Crawl3rGer:
Any updates on this one, @Kendru98 ? Facing the same issue. Thank you!

Kendru98:
Already got the cause of issue, Developer page that we need to open is at the end of the list now, we need to add scroll there. Should be quick, but debugging which iOS version is differentcan be time-consuming 😓

Kendru98:
Are you saying guys that this is broken since iOS 18?

Crawl3rGer:
@Kendru98 yes, this is broken since iOS 18. We have used iOS 17 before and our system settings test worked fine.

Kendru98:
Fix on the way, guys. Thanks for your patience. 😓

Friesenjung:
How is it going? We are really looking forward to the fix :) ❤️

Kendru98:
This week for sure, maybe even today 👍

Kendru98:
@Friesenjung New patrol version was released today, please check if this fix/workaround working for you.

Friesenjung:
@Kendru98 i will Check it later this day. Thanks so far!

PR #2705, opened 2025-07-24
Fix/darkmode ios 18 simulator

Kendru98:
Fix for [#2393](https://github.com/leancodepl/patrol/issues/2393)
Add scroll on simulator iOS 18 devices to be able to click Developer options and change theme.

### DIFF
--- a/packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
+++ b/packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
@@ -396,6 +396,7 @@
           if let osVersionString = self.getOsVersion().split(separator: ".").first,
             let osVersion = Int(osVersionString)
           {
+            // For iOS 18 and above, we need to swipe to the developer option because it's is moved to the bottom of the list
             if osVersion >= 18 {
               let start = self.preferences.coordinate(
                 withNormalizedOffset: CGVector(dx: 0.5, dy: 0.9))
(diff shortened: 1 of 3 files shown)


packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
```diff
@@ -390,6 +390,17 @@
...
+            let osVersion = Int(osVersionString)
+          {
+            if osVersion >= 18 {
+              let start = self.preferences.coordinate(
+                withNormalizedOffset: CGVector(dx: 0.5, dy: 0.9))
+              let end = self.preferences.coordinate(
+                withNormalizedOffset: CGVector(dx: 0.5, dy: 0.3))
+              start.press(forDuration: 0.1, thenDragTo: end)
+            }
+          }
```
pdenert: Add comments to describe, why we need to scroll

pdenert: Are we sure, that this will work on every device, regardless of screen size?

Kendru98: added

Kendru98: iOS simulators are very similar tbh it should work on most for sure, I was testing 2 of them and it was working correctly

Issue #2506, opened 2025-01-23
Enable wifi is not working on iPhone real device


manojkumar-bandlamudi:
### Steps to reproduce

1. Clone the sample patrol tests repo 
2. add 'native.enableWifi()' in the test
3. Run the tests

### Actual results

- Control center is being opened but driver not able to turn off the wifi

### Logs

<details>
<summary>Logs</summary>

```

        2025-01-23 17:38:18.057002+1030 RunnerUITests-Runner[] PatrolServer: INFO: enabling wifi...
            t =    27.23s Press Application 'com.apple.springboard'[0.90, 0.01] for 0.1 seconds, then drag to Application 'com.apple.springboard'[0.90, 0.20] with velocity of 500.00 pixels per second, and then hold for 0.0 seconds 
            t =    27.23s     Wait for com.apple.springboard to idle
            t =    27.24s     Find the Application 'com.apple.springboard'
            t =    27.29s Ignoring failure to get hierarchy for remote element in process 32635 (Error getting main window kAXErrorServerNotFound)
            t =    27.30s Ignoring failure to get hierarchy for remote element in process 32635 (Error getting main window kAXErrorServerNotFound)
            t =    27.31s     Check for interrupting elements affecting " " Application
            t =    27.32s         Wait for au.com.youi.clientdev to idle
            t =    27.34s     Synthesize event
            t =    27.34s         Find the Application 'com.apple.springboard'
            t =    27.39s             Ignoring failure to get hierarchy for remote element in process 32635 (Error getting main window kAXErrorServerNotFound)
            t =    27.40s             Ignoring failure to get hierarchy for remote element in process 32635 (Error getting main window kAXErrorServerNotFound)
            t =    27.42s         Find the Application 'com.apple.springboard'
            t =    27.44s             Ignoring failure to get hierarchy for remote element in process 32635 (Error getting main window kAXErrorServerNotFound)
            t =    27.44s             Ignoring failure to get hierarchy for remote element in process 32635 (Error getting main window kAXErrorServerNotFound)
            t =    27.44s         Find the Application 'com.apple.springboard'
            t =    27.46s             Ignoring failure to get hierarchy for remote element in process 32635 (Error getting main window kAXErrorServerNotFound)
            t =    27.46s             Ignoring failure to get hierarchy for remote element in process 32635 (Error getting main window kAXErrorServerNotFound)
            t =    27.46s         Find the Application 'com.apple.springboard'
[...]

PR #2719, opened 2025-08-11
Add WiFi enable/disable methods in IOSAutomator to support iOS 18 and…

Kendru98:
Add enabling wifi in iOS 18. For some weird reason toggle button is not reachable so we use images of WiFi icon as workaround

### DIFF
--- a/packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
+++ b/packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
@@ -529,34 +529,77 @@
 
     func enableWiFi() throws {
       try runControlCenterAction("enabling wifi") {
-        let toggle = self.springboard.switches["wifi-button"]
-        let exists = toggle.waitForExistence(timeout: self.timeout)
-        guard exists else {
-          throw PatrolError.viewNotExists("wifi-button")
-        }
+        if let osVersionString = self.getOsVersion().split(separator: ".").first,
+          let osVersion = Int(osVersionString)
+        {
+          let toggle: XCUIElement
 
-        if toggle.value! as! String == "0" {
-          toggle.tap()
-        } else {
-          Logger.shared.i("wifi is already enabled")
+          if osVersion >= 18 {
+            let wifiOff = self.springboard.images["wifi.slash"].firstMatch
+            let wifiOn = self.springboard.images["wifi"].firstMatch
+            if wifiOn.exists {
+              Logger.shared.i("wifi is already enabled")
+              return
+            }
+            if !wifiOn.exists && !wifiOff.exists {
+              throw PatrolError.viewNotExists("wifi-button")
+            }
+            wifiOn.tap()
+            return
+          } else {
+            toggle = self.springboard.switches["wifi-button"]
+          }
+
+          let exists = toggle.waitForExistence(timeout: self.timeout)
+          guard exists else {
+            throw PatrolError.viewNotExists("wifi-button")
+          }
+
+          if toggle.value! as! String == "0" {
+            toggle.tap()
+            // Disabling wifi can cause a system alert to appear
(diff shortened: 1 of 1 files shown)


Kendru98:
https://github.com/leancodepl/patrol/issues/2506

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
```diff
@@ -529,34 +529,77 @@
...
-        let exists = toggle.waitForExistence(timeout: self.timeout)
-        guard exists else {
-          throw PatrolError.viewNotExists("wifi-button")
-        }
+        if let osVersionString = self.getOsVersion().split(separator: ".").first,
+          let osVersion = Int(osVersionString)
+        {
+          let toggle: XCUIElement
+
+          if osVersion >= 18 {
```
piotruela: I don't think we need to do that check this way. Look how I check for the name of notificationCell depending on which iOS we're running on:
```swift
    private let notificationCellIdentifier: String = {
      if #available(iOS 18, *) {
        return "ListCell"
      } else {
        return "NotificationCell"
      }
    }()
    ```

File: packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift [hunk 2]
```diff
@@ -529,34 +529,77 @@
...
+            let wifiOn = self.springboard.images["wifi"].firstMatch
+            if wifiOn.exists {
+              Logger.shared.i("wifi is already enabled")
+              return
+            }
+            if !wifiOn.exists && !wifiOff.exists {
+              throw PatrolError.viewNotExists("wifi-button")
+            }
+            wifiOn.tap()
+            return
```
piotruela: That's weird. First, we check if `wifiOn.exists` and if it does, we say that wifi is on and return (that sounds reasonable).
Then we check if none of the buttons(images) exists and if it true, we throw an error (also, fine).
But then to enable the wifi we click on `wifiOn.tap()`? But to click on it it has to exists right? So how is that possible that we will even get to that line and not return in line 542?

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
```diff
@@ -529,34 +529,77 @@
...
+            throw PatrolError.viewNotExists("wifi-button")
+          }
+
+          if toggle.value! as! String == "0" {
+            toggle.tap()
+            // Disabling wifi can cause a system alert to appear
+            try self.acceptSystemAlertIfVisible()
+          } else {
+            Logger.shared.i("wifi is already disabled")
+          }
```
piotruela, line 559: Should't that code be inside the same block of code as `toggle = self.springboard.switches["wifi-button"]` in line 550? I thought this is the part that gets executed when we are running on iOS < 18

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
Kendru98: Now im thinking why this works with wifi test in our repo. Im thinking if we need to cover case when wifi is enabled and we want to enable it. Cant we just click on wifi.slash icon to ON wifi, and if icon is not found method will throw?

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
Kendru98: should be done

packages/patrol/darwin/Classes/AutomatorServer/Automator/IOSAutomator.swift
Kendru98: Okay I think logic should be correct now @piotruela

Issue #2148, opened 2024-03-11
Choosing device of multiple ones during cli commands


jBorkowska:
Now user has to pass an argument to choose a device, on which to run eg. `patrol test`. We'd like to still have this opportunity, but if user haven't provided any device argument and there are multiple ones, we would ask them to interactively choose one (like Flutter does during `flutter run`), instead of choosing random one

jBorkowska:
size: S

PR #2722, opened 2025-08-11
Add interactive device selection to DeviceFinder

Kendru98:
- Added logic to determine if interactive selection should be shown based on the environment (not in CI).
- Enhanced documentation for the device selection behavior.

Resolve: https://github.com/leancodepl/patrol/issues/2148

### DIFF
--- a/packages/patrol_cli/lib/src/base/interactive_prompts.dart
+++ b/packages/patrol_cli/lib/src/base/interactive_prompts.dart
@@ -1,0 +1,63 @@
+import 'dart:io';
+
+import 'package:patrol_cli/src/base/logger.dart';
+
+/// Utility class for interactive command-line prompts.
+class InteractivePrompts {
+  InteractivePrompts({required Logger logger}) : _logger = logger;
+
+  final Logger _logger;
+
+  /// Prompts the user to select an option from a list.
+  ///
+  /// Returns the index of the selected option, or null if the user cancels.
+  /// Throws [UnsupportedError] if stdin doesn't have a terminal.
+  int? selectFromList<T>({
+    required String prompt,
+    required List<T> options,
+    required String Function(T) displayFunction,
+    String cancelMessage = 'Cancel',
+  }) {
+    if (!stdin.hasTerminal) {
+      throw UnsupportedError('Interactive prompts require a terminal');
+    }
+
+    if (options.isEmpty) {
+      throw ArgumentError('Options list cannot be empty');
+    }
+
+    _logger.info('$prompt\n');
+
+    // Display the options
+    for (var i = 0; i < options.length; i++) {
+      _logger.info('  [${i + 1}] ${displayFunction(options[i])}');
+    }
+    _logger.info('  [q] $cancelMessage\n');
+
+    while (true) {
+      stdout.write(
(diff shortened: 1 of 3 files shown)


packages/patrol_cli/lib/src/devices.dart
```diff
@@ -158,6 +175,49 @@ class DeviceFinder {
...
+  /// Determines if interactive device selection should be shown.
+  ///
+  /// Returns true if:
+  /// - We have a terminal (stdin.hasTerminal)
+  /// - Not running in CI environment
+  /// - Not running in test environment
+  bool _shouldShowInteractiveSelection() {
+    // Check if we're in a test environment
+    // PATROL_CLI_TEST: Set to 'true' when running Patrol CLI tests
+    final testEnv = io.Platform.environment['PATROL_CLI_TEST'] == 'true';
```
piotruela: What's that? I don't see this flag anywhere outside this PR. Isn't that a hallucination?

pdenert:
@Kendru98 could you add a demo to this PR?

Kendru98:
> @Kendru98 could you add a demo to this PR?

Is this enough or you want video? :D 

image

packages/patrol_cli/lib/src/base/interactive_prompts.dart
```diff
@@ -0,0 +1,63 @@
...
+    required String Function(T) displayFunction,
+    String cancelMessage = 'Cancel',
+  }) {
+    if (!stdin.hasTerminal) {
+      throw UnsupportedError('Interactive prompts require a terminal');
+    }
+
+    if (options.isEmpty) {
+      throw ArgumentError('Options list cannot be empty');
+    }
```
pdenert, line 27: Should this be shown, when there will be only one option?

Kendru98: This is check for empty list of devices not one, wdym?

pdenert: I mean - should we show interactive prompt, when we had one item one the list?

Kendru98: this is covered in devices.dart file,   
if (attachedDevices.length > 1 && _shouldShowInteractiveSelection()) {
        final selectedDevice = _promptForDeviceSelection(attachedDevices);
        if (selectedDevice == null) {
          throwToolExit('Device selection was cancelled');
        }
        return [selectedDevice];
      }

pdenert: I know it's covered for this case, I was thinking about general approach in this InteractivePrompts, but it can stay like this for now

Kendru98: It is tested well, not sure if CI check will for when someone want to run patrol test on CI, but it is not common case i would say

davidahlstroem:
Can i disable this feature? In the config or someplace else?
I think most developers use one emulator for developing new test and web is not that common.
So this feature add 5-10 seconds of waiting to choose the emulator (with options: emu, web edge, web chrome), instead of, like before, just choosing the first it finds automatically.
Yes, i know i can use -d to target the emulator.

Issue #2723, opened 2025-08-12
Android test report path is not correct sometimes


Kendru98:
### Steps to reproduce

Try to open android test report in app with flavors

### Actual results

There is missing /debug/ part of path often. When user using flavors

### Patrol version

-

### Patrol Doctor output

-

### Flutter Doctor output

-

ivan-horchakov:
Same. I've wrote about this issue in Discord DM for Julia Borowska 26.11.2024

Kendru98:
@ivan-horchakov Hi, next time please create an issue in our GitHub. Fix is on the way should be released next week, thanks for your patience

ivan-horchakov:
@Kendru98 I'm too ~~lazy~~ busy for creating issues here with this kind of priority & severity 🤓. For me is faster to add "debug" in broken path 😉.

StanislavTalanov:
I use command `open build/app/reports/androidTests/connected/debug/flavors/dev/index.html`

PR #2724, opened 2025-08-12
Fix path for android report; Add tests for path

Kendru98:
https://github.com/leancodepl/patrol/issues/2723


packages/patrol_cli/lib/src/android/android_test_backend.dart
```diff
@@ -236,12 +236,17 @@ class AndroidTestBackend {
             )
             ..disposedBy(scope);
 
-      var flavorPath = '';
+      var buildModeAndFlavorPath = '';
```
jBorkowska: maybe separating those into 2 variables would be more readable

packages/patrol_cli/test/android/android_test_backend_report_path_test.dart
```diff
@@ -0,0 +1,158 @@
...
+        flutter: flutterOptions,
+        appServerPort: 8081,
+        testServerPort: 8082,
+        uninstall: false,
+      );
+
+      // Test the internal path generation logic by simulating what happens
+      // in the execute method
+      const rootPath = '/example/project';
+      var buildModeAndFlavorPath = '';
```
jBorkowska: It's great you wrote those tests, but you have to refactor your code, so you can use the original code in the tests. Now it's a bit pointless - someone changes original code and test could still pass even when there's a bug

jBorkowska: Applies to all these tests

packages/patrol_cli/test/android/android_test_backend_report_path_test.dart
```diff
@@ -0,0 +1,158 @@
...
+          .toLowerCase();
+
+      // This mirrors the logic in android_test_backend.dart
+      if (flutterOptions.flavor != null) {
+        buildModeAndFlavorPath = '$buildMode/flavors/${flutterOptions.flavor}/';
+      } else {
+        buildModeAndFlavorPath = '$buildMode/';
+      }
+
+      final expectedPath =
```
jBorkowska: I think this should be actual, and other one is expected

packages/patrol_cli/test/android/android_test_backend_report_path_test.dart
```diff
@@ -0,0 +1,158 @@
+import 'package:patrol_cli/src/crossplatform/app_options.dart';
+import 'package:patrol_cli/src/ios/ios_test_backend.dart';
+import 'package:patrol_cli/src/runner/flutter_command.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('AndroidTestBackend report path generation', () {
```
jBorkowska: Those tests have a lot in common, please refactor so code is not repeated

jBorkowska (changes requested):

packages/patrol_cli/test/android/android_test_backend_report_path_test.dart
Kendru98: So I think we should remove tests in this case

Kendru98: not sure if this is worth to refactor whole path creation to have tests for this simple path

packages/patrol_cli/test/android/android_test_backend_report_path_test.dart
```diff
@@ -0,0 +1,65 @@
+import 'package:patrol_cli/src/android/android_test_backend.dart';
+import 'package:patrol_cli/src/ios/ios_test_backend.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('AndroidTestBackend.generateTestReportPath()', () {
```
jBorkowska, line 6: In my opinion this code is too complicated for a test, but let's hear what others have to say here

jBorkowska (commented):
One more thing - what if `build` folder is somewhere else? It's modifiable in one of the gradle's files (build.gradle I think)

packages/patrol_cli/test/android/android_test_backend_report_path_test.dart
Kendru98: don't get it, what code? generating report path method code?

pdenert (approved):
LGTM

PR #2738, opened 2025-09-23
Docs/vscode extension

jBorkowska:


docs/documentation/patrol-vs-code-extension.mdx
```diff
@@ -0,0 +1,135 @@
...
+patrol:
+  flavor: dev
+  test_file_suffix: _patrol.dart # add this line
+  android:
+    package_name: pl.leancode.patrol.example.dev
+  ios:
+    bundle_id: pl.leancode.patrol.Example
+```
+3. Rename your test files - change the suffix from `_test.dart` to `_patrol.dart`.
+4. If you have a method wrapping `patrolTest()`, you need to add an annotation to it:
```
zoskar, line 23: Maybe it's worth saying that if I don't have a wrapper I need to annotate all tests

docs/documentation/patrol-vs-code-extension.mdx
```diff
@@ -0,0 +1,135 @@
+# Guide for Patrol VS Code extension
+
+## How to setup
+
+1. Install the extension from VS Code Marketplace. It requires Dart extension -
```
zoskar: Do we already have a beta version in Marketplace? We could link it here, probably the redirect would open IDE with extension page

docs/documentation/patrol-vs-code-extension.mdx
jBorkowska: It's not true, if you don't have wrapper you don't have to annotate anything

docs/documentation/patrol-vs-code-extension.mdx
jBorkowska: Not yet, we'll add it later

PR #2741, opened 2025-09-28
Feature/patrol video recording

Kendru98:
Adds opt-in video recording of test runs, gated behind the `--record-video` flag (off by default).

Related: #2745

### Devices & commands
`patrol test --record-video` and `patrol develop --record-video`. Supported on **Android emulators** (`adb shell screenrecord`) and **iOS simulators** (`xcrun simctl io recordVideo`). **Physical Android devices may also work, depending on the vendor** — verified working on a Redmi (Android 15), blocked on a OnePlus (Android 16) where the ROM's SELinux policy denies `screenrecord` writes. iOS physical devices are not supported.

### Per test case or per file?
Per **test case**. Recording starts on each test's `start` and stops on `success`/`failure` (skipped tests are not recorded), producing one `.mp4` per test case.

### Effect on `patrol test` run time
None when the flag is off — recording is fully behind `--record-video`; without it no recording process is started. With it enabled there is minor overhead (a recording process per test + an `adb pull` on Android).

### How videos are saved
`.mp4` files written to `--video-output-dir` (default `<test-dir>/videos`), named `patrol_<test>_<device>_<timestamp>.mp4`. On Android the file is recorded on-device, finalized with SIGINT, then pulled to the output dir and removed from the device; on the iOS simulator it is written straight to the output dir.

### macOS / web
Excluded. `videoConfig` is only passed to the Android and iOS backends, so the flag is silently ignored on macOS and web (no recording).

### Flags
`--record-video`, `--video-output-dir`, `--video-size` (Android only, e.g. `1280x720`), `--video-bit-rate` (Android only, bits/s).

### Known limitations
- iOS physical devices: not supported (recorder logs a warning and skips; the test still runs).
- Some Android OEM ROMs block `adb shell screenrecord` from writing to storage via SELinux (observed on OnePlus/Android 16). On such devices recording is skipped with a clear warning; the test still runs.


packages/patrol_cli/lib/src/android/android_test_backend.dart
```diff
@@ -317,6 +352,11 @@ class AndroidTestBackend {
       patrolLogReader.stopTimer();
       processLogcat.kill();
 
+      // Cleanup video recording manager
+      if (videoRecordingManager != null) {
+        await videoRecordingManager.dispose();
+      }
```
PiotrRogulski:
```suggestion
      await videoRecordingManager?.dispose();
```

packages/patrol_cli/lib/src/android/patrol_log_reader_with_video.dart
```diff
@@ -0,0 +1,172 @@
...
+  final bool hideTestSteps;
+  final bool clearTestSteps;
+  final VideoRecordingManager? videoRecordingManager;
+  final void Function(Entry entry)? onLogEntry;
+  final StreamSubscription<void> Function(
+    void Function(String) onData, {
+    Function? onError,
+    void Function()? onDone,
+    bool? cancelOnError,
+  })
```
PiotrRogulski: This should probably be a typedef to reuse everywhere

packages/patrol_cli/lib/src/android/video_recording_config.dart
```diff
@@ -0,0 +1,67 @@
...
+    this.timeLimit = 180, // Default Android screenrecord limit
+  });
+
+  /// Creates a VideoRecordingConfig from command line arguments.
+  factory VideoRecordingConfig.fromArgs({
+    required bool recordVideo,
+    required String videoOutputDir,
+    String? videoSize,
+    String? videoBitRate,
+  }) {
```
PiotrRogulski: I'm not sure if this should be an entire "proper" constructor – this is just a rename-like mapping of params, which can be done in the test command itself.

This constructor implicitly duplicates the names of command line parameters like `--record-video` or `--video-output-dir` – if these names ever change, this config would need to change as well

packages/patrol_cli/lib/src/android/video_recording_config.dart
```diff
@@ -0,0 +1,67 @@
...
+    required String testName,
+  }) {
+    final timestamp = DateTime.now().millisecondsSinceEpoch;
+    final sanitizedTestName = testName.replaceAll(RegExp(r'[^\w\-_]'), '_');
+    final sanitizedDeviceId = deviceId.replaceAll(RegExp(r'[^\w\-_]'), '_');
+    return 'patrol_${sanitizedTestName}_${sanitizedDeviceId}_$timestamp.mp4';
+  }
+
+  /// Path on the Android device where the video will be temporarily stored.
+  String getDeviceVideoPath(String filename) {
```
PiotrRogulski: I'd encode the Android-only-ness in the method name itself

packages/patrol_cli/lib/src/android/patrol_log_reader_with_video.dart
```diff
@@ -0,0 +1,172 @@
...
+/// Interface for patrol log readers.
+abstract class PatrolLogReaderInterface {
+  void listen();
+  void startTimer();
+  void stopTimer();
+  String get summary;
+}
+
+/// Extended PatrolLogReader that supports video recording per test case.
+class PatrolLogReaderWithVideo implements PatrolLogReaderInterface {
```
PiotrRogulski: Instead of a wrapper, would it be possible to **extend** PatrolLogReader? We wouldn't need to define additional interfaces with similar names

packages/patrol_cli/lib/src/android/patrol_log_reader_with_video.dart
```diff
@@ -0,0 +1,172 @@
...
+import 'package:patrol_log/patrol_log.dart';
+import 'package:patrol_log/patrol_log_reader.dart';
+
+/// Interface for patrol log readers.
+abstract class PatrolLogReaderInterface {
+  void listen();
+  void startTimer();
+  void stopTimer();
+  String get summary;
+}
```
PiotrRogulski: If this interface is indeed needed (see my other comment below in this file), it should probably move out of this file (it's not Android-only)

packages/patrol_cli/lib/src/ios/ios_test_backend.dart
```diff
@@ -236,6 +272,11 @@ class IOSTestBackend {
       patrolLogReader.stopTimer();
       processLogs.kill();
 
+      // Cleanup video recording manager
+      if (videoRecordingManager != null) {
+        await videoRecordingManager.dispose();
+      }
```
PiotrRogulski:
```suggestion
      // Cleanup video recording manager
      await videoRecordingManager?.dispose();
```

packages/patrol_cli/lib/src/runner/patrol_command.dart
```diff
@@ -376,6 +376,31 @@ abstract class PatrolCommand extends Command<int> {
       );
   }
 
+  void usesVideoRecordingOptions() {
+    argParser
+      ..addFlag(
+        'record-video',
+        help:
+            'Record video of the test execution '
+            '(Android emulators and iOS simulators only).',
```
PiotrRogulski: Update the comment about Android (real devices?)

packages/patrol_cli/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - Fix `--clear-permissions` being ignored by `patrol build ios`. The flag was wired into `patrol test` but dropped from `build ios`, so prebuilt iOS test bundles (e.g. for BrowserStack/Firebase Test Lab) never had `CLEAR_PERMISSIONS` enabled.
 - Don't listen for `SIGTERM` on Windows, where it is not supported and throws an unhandled `SignalException`. (#3035)
 - Fix `--exclude` not working. (#2990)
+- Add `--record-video` flag to `patrol test` and `patrol develop` to record a video per test case on Android emulators and iOS simulators. (#2741)
```
PiotrRogulski: Ditto

packages/patrol_cli/lib/src/crossplatform/video_recording_config.dart
```diff
@@ -0,0 +1,52 @@
...
+  final int timeLimit;
+
+  /// Generates a unique filename for the video recording.
+  String generateVideoFilename({
+    required String deviceId,
+    required String testName,
+  }) {
+    final timestamp = DateTime.now().millisecondsSinceEpoch;
+    final sanitizedTestName = testName.replaceAll(RegExp(r'[^\w\-_]'), '_');
+    final sanitizedDeviceId = deviceId.replaceAll(RegExp(r'[^\w\-_]'), '_');
```
PiotrRogulski:
```suggestion
    final sanitizedTestName = testName.replaceAll(RegExp(r'[^\w\-]'), '_');
    final sanitizedDeviceId = deviceId.replaceAll(RegExp(r'[^\w\-]'), '_');
```
No need to replace underscores with underscores 😛

packages/patrol_cli/lib/src/ios/ios_video_recording_manager.dart
```diff
@@ -0,0 +1,305 @@
...
+      // Give the recording a moment to start and wait for the "Recording started" message
+      await Future<void>.delayed(const Duration(milliseconds: 2000));
+      _logger.detail('iOS video recording started successfully');
+
+      // Try to bring simulator to foreground to ensure content is visible
+      try {
+        await _processManager.run([
+          'osascript',
+          '-e',
+          'tell application "Simulator" to activate',
```
PiotrRogulski: FYI: once macOS 27 and Xcode 27 become stable and widely available, this will need to support the `DeviceHub` as well – Flutter had to do that as well https://github.com/flutter/flutter/pull/187910

zoskar:
offtop: whenever I see/hear "Record video":

image

packages/patrol_cli/lib/src/ios/ios_video_recording_manager.dart
Kendru98: Thanks we create an task in our board to adjust patrol for this

PiotrRogulski (approved):

packages/patrol_cli/lib/src/android/android_video_recording_manager.dart
```diff
@@ -0,0 +1,259 @@
...
+
+      // Detect an immediate exit (e.g. `screenrecord` denied access to the
+      // output path, which some OEM/SELinux policies enforce).
+      int? earlyExitCode;
+      unawaited(
+        _currentRecordingProcess!.exitCode.then((code) => earlyExitCode = code),
+      );
+
+      // Give the recording a moment to start.
+      await Future<void>.delayed(const Duration(milliseconds: 500));
```
pdenert, line 100: Does this extend time of the tests execution? Is this value (500ms) specific or some random? Can this be smaller?

packages/patrol_cli/lib/src/android/android_test_backend.dart
```diff
@@ -281,7 +297,13 @@ class AndroidTestBackend {
...
               hideTestSteps: hideTestSteps,
               clearTestSteps: clearTestSteps,
-              onLogEntry: onLogEntry,
+              onLogEntry: (entry) {
+                // Start/stop video recording on test lifecycle events.
+                if (entry is TestEntry) {
+                  videoRecordingManager?.handleTestEntry(entry);
+                }
+                onLogEntry?.call(entry);
+              },
```
pdenert: This looks ugly. I understand intensions but would prefer to have separate method for video

packages/patrol_cli/lib/src/android/android_video_recording_manager.dart
```diff
@@ -0,0 +1,259 @@
...
+      deviceId: _device.id,
+      testName: _sanitizeTestName(testName),
+    );
+    _currentDeviceVideoPath = _config.getAndroidDeviceVideoPath(
+      _currentVideoFilename!,
+    );
+
+    _logger
+      ..detail('Starting video recording for test: $testName')
+      ..detail('Video file: $_currentVideoFilename');
```
pdenert, line 61: Where this will be printed? Why don't we use patrol_log here?

packages/patrol_cli/lib/src/commands/develop_options.dart
```diff
@@ -75,6 +78,16 @@ class DevelopOptions {
       clearTestSteps: results['clear-test-steps'] as bool,
       checkCompatibility: results['check-compatibility'] as bool,
       iosVersion: results['ios'] as String?,
+      videoConfig: VideoRecordingConfig(
+        enabled: results['record-video'] as bool,
+        // Video output directory defaults to a `videos` directory next to
+        // the test target.
```
pdenert: This comments seems excessive

packages/patrol_cli/lib/src/commands/test.dart
```diff
@@ -458,10 +461,19 @@ See https://github.com/leancodepl/patrol/issues/1316 to learn more.
     required bool showFlutterLogs,
     required bool hideTestSteps,
     required bool clearTestSteps,
+    required String testDirectory,
   }) async {
     Future<void> Function() action;
     Future<void> Function()? finalizer;
 
+    // Video output directory always follows test directory
```
pdenert: same for this comment

packages/patrol_cli/lib/src/ios/ios_test_backend.dart
```diff
@@ -201,7 +217,13 @@ class IOSTestBackend {
...
               hideTestSteps: hideTestSteps,
               clearTestSteps: clearTestSteps,
-              onLogEntry: onLogEntry,
+              onLogEntry: (entry) {
+                // Start/stop video recording on test lifecycle events.
+                if (entry is TestEntry) {
+                  videoRecordingManager?.handleTestEntry(entry);
+                }
+                onLogEntry?.call(entry);
+              },
```
pdenert: same here

packages/patrol_cli/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased
+
+- Add `--record-video` flag to `patrol test` and `patrol develop` to record a video per test case on Android emulators and iOS simulators. May also work on physical Android devices, depending on the vendor (some OEM ROMs block `screenrecord`); iOS physical devices are not supported. (#2741)
```
pdenert: There's no need to be so detailed in changelog. Instead let's add something in the patrol docs

packages/patrol_cli/lib/src/ios/ios_video_recording_manager.dart
pdenert, line 1: We should have single VideoRecordingManager abstract class that will define interface, and then implement this for android and ios

packages/patrol_cli/lib/src/android/android_video_recording_manager.dart
Kendru98: No, doesn't extend test time - recording runs in parallel and the test never waits on it. The 500ms is just a small window to catch screenrecord failing to start (some devices block it from writing to storage) so we warn early instead of failing later on adb pull. Could be lower but then we'd miss that on slower devices.

packages/patrol_cli/lib/src/android/android_video_recording_manager.dart
Kendru98: Good point - moved it to the CLI summary. Per-test saves are now detail level (--verbose only) and we print one "Recorded N videos to <dir>" line next to the test summary. patrol_log itself is for the on-device test entries and report, so CLI-side output stays on Logger.

packages/patrol_cli/lib/src/ios/ios_video_recording_manager.dart
ecolab-neil: @pdenert  i'm excited about this, please let us know  when this is merged :)

packages/patrol_cli/lib/src/android/android_video_recording_manager.dart
pdenert, line 1: File name should start with "android_" to match the class name

pdenert (approved):
LGTM

PR #2746, opened 2025-10-03
Patrol web

konowrockis:


ecolab-neil:
@piotruela  @konowrockis  this seems exciting, is there a related issue or roadmap item or docs or ETA for this?

packages/patrol/example/patrol_test/example_test.dart
piotruela, line 1: Change in this file should be reverted

packages/patrol/example/patrol_test/main_test.dart
piotruela, line 1: Revert the change

packages/patrol/lib/src/native/contracts/contracts.dart
```diff
@@ -621,6 +621,68 @@ class IOSSelector with EquatableMixin {
   }
 }
 
+@JsonSerializable()
+class WebSelector with EquatableMixin {
```
piotruela, line 625: If this class wasn't generated from `schema.dart` (and it seems like it was not) it shouldn't be added to this file as it will be removed on next `./gen_from_schema` execution

packages/patrol/lib/src/native/native_automator2.dart
```diff
@@ -1067,4 +1068,101 @@ class NativeAutomator2 {
 
     return response.osVersion;
   }
+
+  // WEB:
+
+  Future<void> tapWeb({
```
piotruela, line 1074: Missing documentation for a public member.

packages/patrol/lib/src/native/native_automator2.dart
```diff
@@ -1067,4 +1068,101 @@ class NativeAutomator2 {
...
+
+  Future<void> enterTextWeb({
+    required WebSelector selector,
+    required String text,
+    WebSelector? iframeSelector,
+  }) async {
+    throw UnimplementedError('enterText is not implemented for web');
+  }
+
+  Future<void> grantPermissions({required List<String> permissions}) async {
```
piotruela, line 1089: Can we define an enum of permissions based on playwright documentation?

packages/patrol/lib/src/native/native_automator2.dart
```diff
@@ -1067,4 +1068,101 @@ class NativeAutomator2 {
...
+    throw UnimplementedError('goForward is not implemented for web');
+  }
+
+  Future<void> pressKey({required String key}) async {
+    throw UnimplementedError('pressKey is not implemented for web');
+  }
+
+  Future<void> pressKeyCombo({required List<String> keys}) async {
+    throw UnimplementedError('pressKeyCombo is not implemented for web');
+  }
```
piotruela, line 1119: Any chance we could reuse Flutter API on those? `List<String> keys` -> `LogicalKeySet keys` and `String key` -> `LogicalKeyboardKey.key`?

packages/patrol/lib/src/native/native_automator2.dart
```diff
@@ -1067,4 +1068,101 @@ class NativeAutomator2 {
...
+
+  Future<void> uploadFile({required List<Object> files}) async {
+    throw UnimplementedError('uploadFile is not implemented for web');
+  }
+
+  Future<List<String>> verifyFileDownloads() async {
+    throw UnimplementedError('verifyFileDownloads is not implemented for web');
+  }
+
+  Future<void> resizeWindow({required int width, required int height}) async {
```
piotruela, line 1165:
```suggestion
  Future<void> resizeWindow(Size size) async {
```
?

packages/patrol/lib/src/native/patrol_app_services/patrol_app_service_io.dart
piotruela, line 1: Let's move all the `patrol_app_service_*.dart` into separate directory

packages/patrol/web_runner/README.md
piotruela, line 1: This can be deleted.

packages/patrol_cli/lib/src/test_bundler.dart
```diff
@@ -74,6 +74,8 @@ Future<void> main() async {
 
   final nativeAutomator = NativeAutomator(config: NativeAutomatorConfig());
   await nativeAutomator.initialize();
+  final nativeAutomator2 = NativeAutomator2(config: NativeAutomatorConfig());
+  await nativeAutomator2.initialize();
```
piotruela, line 78: Creating both `NativeAutomator`s and initialising them is not required? How come it always worked without it?

packages/patrol_cli/lib/src/web/web_test_backend.dart
```diff
@@ -0,0 +1,740 @@
...
+
+  final ProcessManager _processManager;
+  final Logger _logger;
+  final DisposeScope _disposeScope;
+
+  Future<void> build(WebAppOptions options) async {
+    _logger.detail('Building web app for testing...');
+
+    // Warn if flavor is specified for web (not supported)
+    if (options.flutter.flavor != null) {
```
piotruela, line 34: Inside `test.dart` in line 180 we throw if user set flavor when running on web. So I guess we don't need to do this check here

packages/patrol_cli/lib/src/web/web_test_backend.dart
```diff
@@ -0,0 +1,740 @@
...
+  }
+
+  Future<Process> _startFlutterWebServer(
+    WebAppOptions options, {
+    required bool develop,
+  }) async {
+    _logger.detail('Starting Flutter web server...');
+
+    // Warn if flavor is specified for web (not supported)
+    if (options.flutter.flavor != null) {
```
piotruela, line 181: ditto

packages/patrol_cli/lib/src/web/web_test_backend.dart
```diff
@@ -0,0 +1,740 @@
...
+        'Failed to parse package_config.json: $e\n'
+        'Please ensure your project dependencies are properly resolved by running "dart pub get".',
+      );
+    }
+
+    throw Exception(
+      'patrol package not found in package configuration.\n'
+      'Please ensure patrol is added as a dependency and run "dart pub get".',
+    );
+  }
```
piotruela, line 641: This method requires some extra care.
We should use https://pub.dev/packages/package_config to get to the config.
Or **maybe** the whole `web_runner/` should be placed under package `patrol_cli` and not `patrol`?

packages/patrol/lib/src/native/native_automator2.dart
alszczep: Ofc technically we can, but based on what they say about permissions in the docs I am not sure if it is a good idea.

image

packages/patrol/lib/src/native/contracts/contracts.dart
konowrockis: This most probably doesn't matter because in platform automator it's [here](https://github.com/leancodepl/patrol/pull/2789/files#diff-ce8cf435469d61a57ad747bf2aca7d1da4d909d03d44829fd8804ac8ced0b5ef)

packages/patrol/lib/src/native/native_automator2.dart
konowrockis: I;d focus on adding documentation in platform automator, because this file is already migrated

packages/patrol_cli/lib/src/test_bundler.dart
konowrockis: Previously there was single "intialize" in `native` and `native2` - they did the same thing so only one was sufficient. Now we only implemented `native2` for web so we needed to add `initialize` call from `native2`. However, again, it doesn't matter, because platform automator rewrites all of this.

Ortes:
Hi why is this PR closed ?

alszczep:
> Hi why is this PR closed ?

No worries, it was merged into another branch, just not directly through github.

PR #2765, opened 2025-10-31
introduce openPlatformApp, add apps enum

### LINKED ISSUE #2766: Add to native2 openPlatformApp and enums for Apple and Google
### New feature

Add openPlatformApp({String? androidAppId, String? iosAppId}).
Add AppleApp enum with closed app Ids list: https://support.apple.com/en-gb/guide/deployment/depece748c41/web 
Add GoogleApp enum with app Ids

zoskar:
Closes #2766

### Summary
- Adds `openPlatformApp()` method to `NativeAutomator2` with `GoogleApp` and `AppleApp` enums for common system apps.
- Command also supports custom Id

- Added `GoogleApp`, `AppleApp` enums
- Updated code generator to support enhanced enums with value fields
- Updated some dependencies to make `gen_from_schema` work (`leancode_lint`, `analyzer`, `dart_style`)

### Usage

```dart
await $.native2.openPlatformApp(
  androidAppId: GoogleApp.chrome,
  iosAppId: AppleApp.safari,
);
```

```dart
await $.native.openPlatformApp(
  androidAppId: 'com.mycompany.myapp',
  iosAppId: 'com.mycompany.myapp',
);
```

### DIFF
--- a/packages/patrol/lib/src/native/contracts/contracts.dart
+++ b/packages/patrol/lib/src/native/contracts/contracts.dart
@@ -240,22 +240,6 @@ class DartGroupEntry with EquatableMixin {
 
   @override
   List<Object?> get props => [name, type, entries, skip, tags];
-
-  DartGroupEntry copyWith({
-    String? name,
-    GroupEntryType? type,
-    List<DartGroupEntry>? entries,
-    bool? skip,
-    List<String>? tags,
-  }) {
-    return DartGroupEntry(
-      name: name ?? this.name,
-      type: type ?? this.type,
-      entries: entries ?? this.entries,
-      skip: skip ?? this.skip,
-      tags: tags ?? this.tags,
-    );
-  }
 }
 
 @JsonSerializable()
@@ -271,10 +255,6 @@ class ListDartTestsResponse with EquatableMixin {
 
   @override
   List<Object?> get props => [group];
-
-  ListDartTestsResponse copyWith({DartGroupEntry? group}) {
-    return ListDartTestsResponse(group: group ?? this.group);
-  }
 }
 
 @JsonSerializable()
@@ -290,10 +270,6 @@ class RunDartTestRequest with EquatableMixin {
 
   @override
   List<Object?> get props => [name];
-
-  RunDartTestRequest copyWith({String? name}) {
-    return RunDartTestRequest(name: name ?? this.name);
-  }
 }
 
 @JsonSerializable()
@@ -310,16 +286,6 @@ class RunDartTestResponse with EquatableMixin {
 
   @override
   List<Object?> get props => [result, details];
-
-  RunDartTestResponse copyWith({
-    RunDartTestResponseResult? result,
-    String? details,
-  }) {
-    return RunDartTestResponse(
-      result: result ?? this.result,
-      details: details ?? this.details,
-    );
-  }
 }
 
 @JsonSerializable()
@@ -335,12 +301,6 @@ class ConfigureRequest with EquatableMixin {
 
   @override
   List<Object?> get props => [findTimeoutMillis];
-
-  ConfigureRequest copyWith({int? findTimeoutMillis}) {
-    return ConfigureRequest(
-      findTimeoutMillis: findTimeoutMillis ?? this.findTimeoutMillis,
-    );
-  }
 }
 
 @JsonSerializable()
@@ -356,10 +316,6 @@ class OpenAppRequest with EquatableMixin {
 
   @override
   List<Object?> get props => [appId];
-
-  OpenAppRequest copyWith({String? appId}) {
-    return OpenAppRequest(appId: appId ?? this.appId);
-  }
 }
 
 @JsonSerializable()
@@ -376,13 +332,6 @@ class OpenPlatformAppRequest with EquatableMixin {
 
   @override
   List<Object?> get props => [androidAppId, iosAppId];
-
-  OpenPlatformAppRequest copyWith({String? androidAppId, String? iosAppId}) {
-    return OpenPlatformAppRequest(
-      androidAppId: androidAppId ?? this.androidAppId,
-      iosAppId: iosAppId ?? this.iosAppId,
-    );
-  }
 }
 
 @JsonSerializable()
@@ -396,10 +345,6 @@ class OpenQuickSettingsRequest with EquatableMixin {
 
   @override
   List<Object?> get props => const [];
-
-  OpenQuickSettingsRequest copyWith() {
-    return OpenQuickSettingsRequest();
-  }
 }
 
 @JsonSerializable()
@@ -415,10 +360,6 @@ class OpenUrlRequest with EquatableMixin {
 
   @override
   List<Object?> get props => [url];
-
-  OpenUrlRequest copyWith({String? url}) {
-    return OpenUrlRequest(url: url ?? this.url);
-  }
 }
 
 @JsonSerializable()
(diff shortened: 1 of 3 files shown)


packages/patrol_gen/lib/src/generators/dart/dart_contracts_generator.dart
```diff
@@ -15,7 +15,9 @@ class DartContractsGenerator {
       buffer.writeln(_createMessage(messageDefintion));
     }
 
-    final content = DartFormatter().format(buffer.toString());
+    final content = DartFormatter(
+      languageVersion: DartFormatter.latestLanguageVersion,
```
zoskar, line 19: Need confirmation from mobile positive person that this is fine

packages/patrol_gen/lib/src/generators/dart/dart_contracts_generator.dart
```diff
@@ -15,7 +15,9 @@ class DartContractsGenerator {
       buffer.writeln(_createMessage(messageDefintion));
```
piotruela: Could you fix this funny typo along the way? `Defintion`

schema.dart
```diff
@@ -489,3 +495,80 @@ enum IOSElementType {
...
+  calculator('com.google.android.calculator'),
+  calendar('com.google.android.calendar'),
+  chrome('com.android.chrome'),
+  drive('com.google.android.apps.docs'),
+  gmail('com.google.android.gm'),
+  maps('com.google.android.apps.maps'),
+  photos('com.google.android.apps.photos'),
+  playStore('com.android.vending'),
+  settings('com.android.settings'),
+  youtube('com.google.android.youtube');
```
piotruela, line 510: Let’s add a comment about the source from which we took these strings.

schema.dart
```diff
@@ -489,3 +495,80 @@ enum IOSElementType {
...
+
+  const GoogleApp(this.value);
+  final String value;
+}
+
+enum AppleApp {
```
piotruela, line 517: ditto

packages/patrol_gen/pubspec.yaml
```diff
@@ -10,11 +10,11 @@ environment:
   sdk: '>=3.5.0 <4.0.0'
```
piotruela: Imo we can bump it to `>=3.8.0 <4.0.0` (and go higher with the dependencies as well). Patrol is already migrated to sdk >3.8.0

packages/patrol_devtools_extension/lib/api/contracts.dart
```diff
@@ -361,6 +361,29 @@ class OpenAppRequest with EquatableMixin {
...
+
+  @override
+  List<Object?> get props => [androidAppId, iosAppId];
+
+  OpenPlatformAppRequest copyWith({String? androidAppId, String? iosAppId}) {
+    return OpenPlatformAppRequest(
+      androidAppId: androidAppId ?? this.androidAppId,
+      iosAppId: iosAppId ?? this.iosAppId,
+    );
+  }
```
piotruela: @Kendru98 I see you've added copyWith to generator some time ago but I don't see it's used for any of the generated models. What was the idea behind it?

packages/patrol/lib/src/native/contracts/contracts.dart
```diff
@@ -10,211 +10,211 @@ import 'package:json_annotation/json_annotation.dart';
...
+  grid('grid'),
+  levelIndicator('levelIndicator'),
+  cell('cell'),
+  layoutArea('layoutArea'),
+  layoutItem('layoutItem'),
+  handle('handle'),
+  stepper('stepper'),
+  tab('tab'),
+  touchBar('touchBar'),
+  statusItem('statusItem');
```
piotruela, line 137: All the `value`s in that enum match the defined enum key. In such case we can get the enum key as `String` using  `IOSElementType.statusItem.name` so we don't need to define the redundant `value` for such sets of enums. 

On the other hand, maybe it's better to keep the consistency of defining/getting the `value` with the more complex cases (like in `GoogleApp`)

piotruela: I leave it for you to decide, just wanted to share some thoughts

packages/patrol_gen/lib/src/resolve_schema.dart
```diff
@@ -15,10 +15,23 @@ Future<Schema> resolveSchema(String schemaPath) async {
...
+        final enumFields = declaration.constants.map((e) {
+          final name = e.name.lexeme;
+          // Extract the value from the enum constant arguments
+          final arguments = e.arguments?.argumentList.arguments;
+          final value = arguments != null && arguments.isNotEmpty
+              ? arguments.first
+                  .toString()
+                  .replaceAll("'", '')
+                  .replaceAll('"', '')
+              : name;
```
piotruela:
```suggestion
          final value =
              switch (e.arguments?.argumentList.arguments.firstOrNull) {
                final expression? =>
                  expression.toString().replaceAll("'", '').replaceAll('"', ''),
                _ => name,
              };
```

packages/patrol_gen/lib/src/generators/dart/dart_contracts_generator.dart
piotruela: Shouldn't it be based on dart formatter version used in `packages/patrol` (as this is the repo where generated code ends up therefore it needs to be compliant with the formatting used there)?

piotruela: Ok I see now that we've introduced `dart format "$patrol_dart_out_dir"` line to `./gen_from_schema` some time ago so my final answer is:
It doesn't matter what we use here because the output will be formatted inside `package/patrol` anyway

schema.dart
zoskar: Android ones were fetched manually since I found no official doc from Google with those package names. Do you think linking the tool makes sense?
https://play.google.com/store/apps/details?id=com.csdroid.pkg&pcampaignid=web_share

packages/patrol_devtools_extension/lib/api/contracts.dart
Kendru98: I changed approach and  finally didn't use it. But maybe it can be useful in the future

packages/patrol/lib/src/native/contracts/contracts.dart
zoskar: @pdenert any opinion?

schema.dart
piotruela: Definitely

packages/patrol_devtools_extension/lib/api/contracts.dart
piotruela: I’m invoking YAGNI and believe we should remove it until we actually need it

Kendru98: If @pdenert does not have objections, let's remove, that was his idea, but as I said, finally it is not used anywhere

pdenert: If we don't use it, please remove it

packages/patrol/lib/src/native/contracts/contracts.dart
pdenert: I understand that it comes from adding value to contract generator. We could expand it to add value only if this is in schema, but this doesn't break anything, so it could stay

packages/patrol_devtools_extension/lib/api/contracts.dart
zoskar: Removed ✅

PR #2775, opened 2025-11-06
Add handling patrol version as path

jBorkowska:


packages/patrol_cli/lib/src/pubspec_reader.dart
```diff
@@ -126,6 +126,8 @@ class PubspecReader {
           if (git['ref'] != null) {
             return git['ref'].toString();
           }
+        } else if (patrol['path'] != null) {
+          return patrol['path'].toString();
```
pdenert, line 130: Could you add a demo how this works? Also please add the tests and update the changelog

jBorkowska: I looked into it more and I think this is not a fix - there was something else that made it work for me locally. I'll do a rework of this PR

PR #2778, opened 2025-11-13
Patrol web playwright config

alszczep:


packages/patrol/web_runner/playwright.config.ts
```diff
@@ -1,18 +1,50 @@
...
+    viewport,
   },
   globalSetup: require.resolve("./tests/setup"),
   // Output test results to the tested app directory
   outputDir,
-  reporter: [["html", { outputFolder, open: "never" }], ["json"], ["list"]],
-  timeout: 300000,
+  reporter: reporter,
+  retries,
+  timeout: timeout ?? 300000,
```
piotruela: The timeout we set here means the timeout for the whole set of tests. From my experience with using Patrol on Web, single e2e test takes >60 seconds on CI. 
I feel like the default timeout value should be `unlimited` or maybe some value multiplied by number of tests (if possible).

packages/patrol_cli/lib/src/runner/patrol_command.dart
```diff
@@ -229,6 +229,76 @@ abstract class PatrolCommand extends Command<int> {
     );
   }
 
+  void usesPlaywright() {
+    argParser
+      ..addOption(
+        'playwright-retries',
```
piotruela: Isn't naming those options **playwright**-sth a bit too "provider-specific"? I'd say playwright as a test runner for web it's just an implementation detail. Maybe "web-option" would be more accurate? Not sure tho

packages/patrol_cli/lib/src/runner/patrol_command.dart
```diff
@@ -229,6 +229,76 @@ abstract class PatrolCommand extends Command<int> {
...
+        help: 'https://playwright.dev/docs/test-retries',
+        valueHelp: 'number',
+      )
+      ..addOption(
+        'playwright-video',
+        help: 'https://playwright.dev/docs/videos',
+        valueHelp: 'off | on | retain-on-failure | on-first-retry',
+      )
```
piotruela, line 243: Should it defaults to `off`?

packages/patrol_cli/lib/src/web/web_test_backend.dart
```diff
@@ -427,6 +427,48 @@ class WebTestBackend {
                 'PATROL_TEST_REPORT_DIR': testReportDir,
                 'PLAYWRIGHT_JSON_OUTPUT_NAME': 'results.json',
                 'PLAYWRIGHT_JSON_OUTPUT_DIR': testReportDir,
+                ...(options.retries != null
+                    ? {'PLAYWRIGHT_RETRIES': options.retries.toString()}
+                    : {}),
```
piotruela:
```suggestion
if(options.retries != null) 'PLAYWRIGHT_RETRIES': options.retries.toString(),
```

piotruela: Same for the rest below

packages/patrol/web_runner/playwright.config.ts
alszczep: `Playwright Test enforces a timeout for each test, 30 seconds by default. Time spent by the test function, fixture setups, and beforeEach hooks is included in the test timeout.`
That is what they say in playwright docs, I'll make it clear that it is for a single test

packages/patrol_cli/lib/src/runner/patrol_command.dart
alszczep: I'd say it should, unless the patrols default is different

piotruela (approved):

PR #2730, opened 2025-09-05
Replace clearing permissions with full app uninstallation

KrzysztofMamak:
Fixes #1317.

# Summary

Replace clearing permissions with full app uninstallation for better test isolation on iOS and add Android's `clearPackageData` management.

# Changes
- Replaced clearing permissions with full app uninstallation on iOS
- Changed flag from `--clear-permissions` to `--full-isolation`
- Allow overriding Android's `clearPackageData` via `--full-isolation` flag in `patrol test`
- Bumped minimum iOS version (`12.0`->`13.0`)
- Unified `IPHONEOS_DEPLOYMENT_TARGET` across all build configurations
- Fix typo in `SymbolName` extension

PR #2803, opened 2025-11-30
`full-isolation` & `clear-permissions`

piotruela:
After testing on device farms, it seems like this trick with uninstalling the app between the test runs does not work when running on real iOS devices with iOS newer that 17. Due to that, I decided to block using this flag when not running the tests on iOS Simulator. Also brings back `clear-permissions` flag as for now `full-isolation` cannot fully replace that one.


pdenert:
Related #2730

PR #2805, opened 2025-12-01
Add MobileSelector and PlatformSelector

fylyppo:
Users should have a successors of `NativeSelector` that handles platform-specific selectors using the `PlatformAutomator`.


packages/patrol/lib/src/platform/selector.dart
```diff
@@ -96,6 +96,37 @@ class Selector implements CompoundSelector {
       web_selector.WebSelector(text: text, cssOrXpath: className);
 }
 
+/// A compound selector that behaves like NativeSelector, allowing platform-specific
```
konowrockis: I think we can forget about `Native*`. This is only important today, but it won't mean anything tomorrow that something behaves like `NativeSelector`.

packages/patrol/lib/src/platform/selector.dart
```diff
@@ -96,6 +96,37 @@ class Selector implements CompoundSelector {
...
+/// iOS and Android selectors to be specified separately.
```
konowrockis: >iOS and Android **and Web**

:) Technically we could future proof it and say just `allowing platform-specific selectors to be specified separately`

packages/patrol/lib/src/platform/selector.dart
```diff
@@ -96,6 +96,37 @@ class Selector implements CompoundSelector {
...
+///
+/// This is useful when iOS and Android require different selector values or types.
+class PlatformCompoundSelector implements CompoundSelector {
```
konowrockis: Not sure I like this name. Maybe just `PlatformSelector`? Or something else?

fylyppo: I don't like it either :) it's just a draft name. Will wait for more ideas from the others

piotruela:
> Should we add the usage to the docs? If so, is it enough?

Let's add an information about existence of `PlatformCompoundSelector` above `tap()` method from `PlatformAutomator`. This is currently the only method that will allow using this selector. And also we can add information that `PlatformCompoundSelector` is one of the implementation of `CompoundSelector` to docs above `CompoundSelector` class. `CompoundSelector` is a param of `tap` so people will naturally look at this class to see what's going on

konowrockis:
> > Should we add the usage to the docs? If so, is it enough?
> 
> Let's add an information about existence of `PlatformCompoundSelector` above `tap()` method from `PlatformAutomator`. This is currently the only method that will allow using this selector. And also we can add information that `PlatformCompoundSelector` is one of the implementation of `CompoundSelector` to docs above `CompoundSelector` class. `CompoundSelector` is a param of `tap` so people will naturally look at this class to see what's going on

Not true, all the `$.platform.mobile` methods also use `CompoundSelector` so it will also work there.

piotruela:
> > > Should we add the usage to the docs? If so, is it enough?
> > 
> > 
> > Let's add an information about existence of `PlatformCompoundSelector` above `tap()` method from `PlatformAutomator`. This is currently the only method that will allow using this selector. And also we can add information that `PlatformCompoundSelector` is one of the implementation of `CompoundSelector` to docs above `CompoundSelector` class. `CompoundSelector` is a param of `tap` so people will naturally look at this class to see what's going on
> 
> Not true, all the `$.platform.mobile` methods also use `CompoundSelector` so it will also work there.

Ok, then first part is wrong but we can still do this:
add information that `PlatformCompoundSelector` is one of the implementation of `CompoundSelector` to docs above `CompoundSelector` class. `CompoundSelector` is a param of all grouped methods so people will naturally look at this class to see what's going on

zoskar (commented):
LGTM, I'm fine with `PlatformSelector` name

PR #2808, opened 2025-12-01
Prepare patrol_finders 3.0.0-dev.1 to release

piotruela:
In patrol_log 0.6.0 we introduced this change:
```dart
  String get nameWithPath {
    const testDirectory = String.fromEnvironment('PATROL_TEST_DIRECTORY');
    return '$_testName ${AnsiCodes.gray}($testDirectory/$_filePath.dart)${AnsiCodes.reset}';
  }
```
If we would release version 2.10.0 of `patrol_finders`, users of `patrol` 3.20.0 would be affected by that change. We need to bump major version of `patrol_finders` to make sure that this will only be used in patrol/cli 4.0.0


pdenert (approved):
LGTM

Issue #2827, opened 2025-12-09
Patrol web test doesn't start: playwright timeout


JaredEzz:
### Steps to reproduce

1. Clone project https://github.com/Ostrich-Cyber-Risk/patrol_testing.git (bare-bones sample starter Flutter App with patrol added to pubspec.yaml)
2. Run `patrol test --device chrome --target patrol_test/app_test.dart --verbose`

### Actual results

upon reaching the step for `$ npx playwright test tests/test.spec.ts`

Error logs read: 
```
Stopping Flutter web server...
Playwright process exited unexpectedly with code 1

See the logs above to learn what happened. Also consider running with --verbose. If the logs still aren't useful, then it's a bug - please report it.
```
error in `playwright-report/index.html` reads:
```
TimeoutError: page.waitForFunction: Timeout 30000ms exceeded.

   at tests\initialise.ts:8

   6 |   })
   7 |
>  8 |   await page.waitForFunction(
     |              ^
   9 |     () => {
  10 |       if (!window.__patrol__onInitialised) return false
  11 |
    at initialise (C:\Users\JaredHasson\AppData\Local\Pub\Cache\hosted\pub.dev\patrol-4.0.0\web_runner\tests\initialise.ts:8:14)
    at setup (C:\Users\JaredHasson\AppData\Local\Pub\Cache\hosted\pub.dev\patrol-4.0.0\web_runner\tests\setup.ts:16:3)
```

Could be a problem with local setup, but nowhere in the docs could I find troubleshooting for this. Also, the [Write your first test](https://patrol.leancode.co/documentation/write-your-first-test) section seems to be written entirely for Android/iOS/native, and not for web. 

### Logs

<details>
<summary>Logs</summary>

```
PS C:\Users\JaredHasson\IdeaProjects\patrol_testing> patrol test --device chrome --target patrol_test/app_test.dart --verbose    
Verbose mode enabled. More logs will be printed.
Received 1 test target(s)
Received test target: C:\Users\JaredHasson\IdeaProjects\patrol_testing\patrol_test/app_test.dart
Generated entrypoint C:\Users\JaredHasson\IdeaProjects\patrol_testing\patrol_test\test_bundle.dart with 1 bundled test(s)
$ flutter --no-version-check --suppress-analytics devices --machine
Received 1 device(s) to run on
Received device: Chrome (chrome)
$ flutter --suppress-analytics --no-version-check pub deps --style=list
[...]

PR #2828, opened 2025-12-09
Do not cancel subscriptions on stdout/stderr of flutter web server

piotruela:
Fixes issue #2827.
For some weird reason, cancelling subscription on stdout/stderr of running Flutter web server, kills the server(?).
I might need to report that to Flutter github, but for now let's fix it in Patrol


pdenert (approved):
LGTM

PR #2829, opened 2025-12-10
Update packages readme files

### LINKED ISSUE #2760: [patrol_finders]pub.dev listing does not indicate web support
https://pub.dev/packages/patrol_finders 
does not show "web"

patrol_finders is a "fairly thin" wrapper around flutter_test that is a dev dependency only; it certainly supports flutter web right?

pdenert:
Update packages readme files.

Define platforms for patrol_finders. Fix #2760

### DIFF
--- a/packages/patrol_cli/README.md
+++ b/packages/patrol_cli/README.md
@@ -23,7 +23,7 @@ Learn more about Patrol:
 
 ## Patrol CLI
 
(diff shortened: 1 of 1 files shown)


jBorkowska:
Good, but I have some small suggestions for the future:
- in other packages than patrol, the overview section about Patrol should be at the end, not at the start
- if we want to show whole ecosystem of packages, would be better to show it more explicitly (eg. in a table)

PR #2833, opened 2025-12-11
add platform_automator to exports

### LINKED ISSUE #2834: Patrol should export `PlatformAutomator`
### Use case

I want to extend `PlatformAutomator` class as I used to do with `NativeAutomator` but it's not possible right now with platform_automator not being exported

### Proposal

Export `platform_automator.dart`

zoskar:
Closes #2834
Add platform_automator to exports so that users can use `PlatformnAutomator` class - for example to extend it


pdenert (approved):
LGTM

zoskar:
@pdenert One more change was needed. Since I exposed platform_automator via patrol, the additional imports in test bundler is no longer needed and it generated warnings in autogenerated files

image

pdenert (approved):
Good job

Issue #2853, opened 2025-12-24
PATROL_FLUTTER_COMMAND not working for the web


tpouriel-twipi:
### Steps to reproduce

1. Add an env variable like so in your shell : `PATROL_FLUTTER_COMMAND="fvm flutter"`
2. Run any minimal patrol test with the command : ` patrol test -t patrol_test/patrol_test.dart -d chrome -v`

### Actual results

PATROL_FLUTTER_COMMAND becomes only `fvm` instead of `fvm flutter` when patrol tries to `fvm flutter run -d web-server ...`

### Logs

<details>
<summary>Logs</summary>

```
Received test target: /Users/toug/workspace/twipi/applications/app/flutter/web/patrol_test/introduction_test.dart
Generated entrypoint /Users/toug/workspace/twipi/applications/app/flutter/web/patrol_test/test_bundle.dart with 1 bundled test(s)
$ fvm flutter --no-version-check --suppress-analytics devices --machine
Received 1 device(s) to run on
Received device: Chrome (chrome)
$ fvm flutter --suppress-analytics --no-version-check pub deps --style=list
Received 6 --dart-define(s) (0 custom, 6 internal)
Received internal --dart-define: PATROL_ANDROID_APP_NAME=Keyclic
Received internal --dart-define: PATROL_IOS_APP_NAME=Keyclic
Received internal --dart-define: INTEGRATION_TEST_SHOULD_REPORT_RESULTS_TO_NATIVE=false
Received internal --dart-define: PATROL_TEST_LABEL_ENABLED=true
Received internal --dart-define: PATROL_TEST_DIRECTORY=patrol_test
Received internal --dart-define: COVERAGE_ENABLED=false
Will uninstall apps before running tests
Starting web test execution...
Building Flutter web app...
Starting Flutter web server...
$ fvm run -d web-server --target=/Users/toug/workspace/twipi/applications/app/flutter/web/patrol_test/test_bundle.dart --debug --dart-define=PATROL_ANDROID_APP_NAME=Keyclic --dart-define=PATROL_IOS_APP_NAME=Keyclic --dart-define=INTEGRATION_TEST_SHOULD_REPORT_RESULTS_TO_NATIVE=false --dart-define=PATROL_TEST_LABEL_ENABLED=true --dart-define=PATROL_TEST_DIRECTORY=patrol_test --dart-define=COVERAGE_ENABLED=false
Waiting for web server to start...
Flutter stderr: Could not find an option or flag "-d".
[...]

PR #2857, opened 2025-12-29
Fix/2853 passing flutter command not work on the web

Kendru98:
Fix for https://github.com/leancodepl/patrol/issues/2853

### DIFF
--- a/packages/patrol_cli/test/crossplatform/app_options_test.dart
+++ b/packages/patrol_cli/test/crossplatform/app_options_test.dart
@@ -543,9 +543,7 @@ void main() {
           dartDefineFromFilePaths: [],
         );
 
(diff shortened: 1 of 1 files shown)


packages/patrol_cli/lib/src/web/web_test_backend.dart
```diff
@@ -33,6 +33,7 @@ class WebTestBackend {
 
     final result = await _processManager.run([
```
pdenert: We should extract content of this run method call to separate function `toFlutterBuildInvocation` like it is already donect for ios and macos. This way we'll be able to test this

pdenert (approved):
LGMT

pdenert (approved):
LGTM

PR #2863, opened 2026-01-04
fix: Catch TestFailure and exceptions in testBody to prevent test hang

akaboshinit:
## Problem

When `fail()` or an exception is thrown inside `patrolTest`, the test hangs indefinitely instead of failing properly.

## Cause

In `_wrapTestBodyWithExceptionGatherer`, only `FlutterError` was caught via `FlutterError.onError`. Regular exceptions and `TestFailure` (thrown by `fail()`) were not caught and therefore not recorded in `_testResults`. This prevented `tearDown` from properly reporting the test result to the native side.

## Solution

Wrap `testBody()` in a `try-catch-finally` block to:
1. Catch all exceptions, including `TestFailure`
2. Record them in `_testResults` for proper reporting
3. Rethrow to let the Flutter test framework handle the failure
4. Ensure `FlutterError.onError` is restored in `finally`

### DIFF
--- a/packages/patrol/lib/src/binding.dart
+++ b/packages/patrol/lib/src/binding.dart
@@ -247,9 +247,26 @@ class PatrolBinding extends LiveTestWidgetsFlutterBinding {
       }
     };
 
-    await testBody();
+    try {
+      await testBody();
+    } catch (err, st) {
+      // Catch TestFailure and other exceptions thrown in testBody
+      if (_currentDartTest case final testName?) {
+        final previousDetails = switch (_testResults[testName]) {
+          Failure(:final details?) => details,
+          _ => null,
+        };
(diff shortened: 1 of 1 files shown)


mateuszwojtczak:
@pdenert @piotruela someone with permissions has to run the pipeline checks

akaboshinit:
After investigating the implementation, I found that this was actually my mistake, so no action is needed on your side.

Specifically, the issue was that when initializing the SentryFlutter plugin (used for Sentry integration), errors were not being properly propagated to Patrol. I haven't fully investigated the details yet, but I believe this is likely because SentryFlutter internally overrides FlutterError.onError.

Since this is a compatibility issue with Sentry's side, adding a workaround in Patrol itself could have unintended side effects. Therefore, no changes are necessary for this.

PR #2898, opened 2026-01-26
fix(ios): support text parameter in tapOnNotificationBySelector

zoskar:
Previously, tapOnNotificationBySelector only checked textContains and titleContains properties, causing Selector(text: ...) to default to an empty string and fail. Now also checks the text property as a fallback.

error
```
    await system.mobile.tapOnNotificationBySelector(
      Selector(text: 'Meeting has started'),
    );
```

👇 
```
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞═════════════════
The following PatrolActionException was thrown running a test:
Patrol action failed: IosAutomatorClientException:
tapOnNotificationBySelector() failed with Invalid response: 400
notification containing text "" doesn't exist
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,35 @@ cd packages/patrol_cli
 flutter pub global activate -s path .
 ```
 
+## 🛠️ Maintained by LeanCode
+
+<div align="center">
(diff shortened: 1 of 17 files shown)


pdenert:
@zoskar what are we missing there?

zoskar:
@pdenert I forgot to undraft PR

PR #2904, opened 2026-01-29
Fix/ios develop test framework output

zoskar:
This PR aligns `patrol develop` logs on iOS with Android in case of failed test. This is important, because output "Some tests failed" at the end of failed develop test is the only consistent way to determine that test run in develop has finished. It's needed in case any other external dependency relied on it 🤭

Before
```
Hot Restart for entrypoint test_bundle.dart...
🧪 counter state is the same after going to home and switching apps
        ❌   1. tap widgets with text "I do not exist!".

```
After:
```
Hot Restart for entrypoint test_bundle.dart...
        00:00 +0: example_test counter state is the same after going to home and switching apps
🧪 counter state is the same after going to home and switching apps
        ❌   1. tap widgets with text "I do not exist!".
        00:10 +0: example_test counter state is the same after going to home and switching apps [E]
        00:10 +0 -1: Some tests failed.
```

I've put this change in patrol log changelog, not sure if that's enough

### DIFF
--- a/packages/patrol_log/CHANGELOG.md
+++ b/packages/patrol_log/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+- Show Flutter test framework output on iOS in `patrol develop`. (#2904)
+
 ## 0.6.0
 
 - Update `TestEntry.nameWithPath` to use actual test directory. (#2728)


packages/patrol_log/lib/src/patrol_log_reader.dart
```diff
@@ -83,11 +83,21 @@ class PatrolLogReader {
...
     try {
       if (line.contains('PATROL_LOG')) {
         _parsePatrolLog(line);
+      } else if (_testFrameworkPattern.hasMatch(line)) {
+        // Print test framework output to the console. On iOS, this output
+        // is captured by the log stream but not by flutter logs.
+        final match = _testFrameworkPattern.firstMatch(line);
+        if (match != null) {
+          log('\t${match.group(1)}');
+        }
```
pdenert, line 100: Did you test how this will behave with enabled clearing lines in patrol test?

zoskar: It does not show up in `patrol test` at all. In develop it's still there regardless of the flag but the only new logs are at the start of the test
` 00:00 +0: autofocus_textfield_test enterText() on autofocused text fields`
and in case if fail at the end
```
00:48 +0: autofocus_textfield_test enterText() on autofocused text fields [E]
          Test failed. See exception logs above.
        00:48 +0 -1: Some tests failed.
```

PR #2905, opened 2026-01-30
fix: listen to stdout/stderr streams in buildApkConfigOnly to prevent hang

latygindenis2:
On Windows 11, `buildApkConfigOnly` hangs indefinitely at `await process.exitCode`. 

Added `listenStdOut` and `listenStdErr` calls with dispose scope to properly consume streams before awaiting `exitCode`. This also provides useful debug logging via `_logger`.

Tested on Windows 11 with:
- Flutter 3.35.6, 3.38.8
- patrol 3.20.0, 4.1.0
- patrol_cli 3.11.0, 4.0.2

Before fix: `patrol build android` hangs indefinitely on `flutter build apk --config-only`
After fix: Build completes successfully

### DIFF
--- a/packages/patrol_cli/lib/src/android/android_test_backend.dart
+++ b/packages/patrol_cli/lib/src/android/android_test_backend.dart
@@ -170,26 +170,26 @@ class AndroidTestBackend {
   /// This fix issue: https://github.com/leancodepl/patrol/issues/1668
   Future<void> buildApkConfigOnly(FlutterAppOptions options) async {
     await _disposeScope.run((scope) async {
-      final process = await _processManager
-          .start([
-            options.command.executable,
-            ...options.command.arguments,
-            'build',
-            'apk',
-            '--config-only',
-            if (options.buildName case final buildName?) ...[
-              '--build-name',
-              buildName,
-            ],
-            if (options.buildNumber case final buildNumber?) ...[
-              '--build-number',
-              buildNumber,
-            ],
-            if (options.noTreeShakeIcons) '--no-tree-shake-icons',
-            '-t',
-            options.target,
-          ], runInShell: true)
-          .disposedBy(scope);
+      final process =
+          await _processManager.start([
+              options.command.executable,
+              ...options.command.arguments,
+              'build',
+              'apk',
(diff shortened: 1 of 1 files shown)


pdenert:
@latygindenis2 thanks for the contribution. I would like to do a one step back. Can you create an issue with reproducible example, so we can confirm the issue and the fix?

packages/patrol_cli/lib/src/android/android_test_backend.dart
```diff
@@ -169,29 +169,34 @@ class AndroidTestBackend {
...
+          buildName,
+        ],
+        if (options.buildNumber case final buildNumber?) ...[
+          '--build-number',
+          buildNumber,
+        ],
+        if (options.noTreeShakeIcons) '--no-tree-shake-icons',
+        '-t',
+        options.target,
+      ], runInShell: true);
```
pdenert: Add `diposedBy(scope)`

latygindenis2: @pdenert, fixed

latygindenis2:
> Looks good, one thing is missing. Also please update the changelog. Keep in mind that there's another PR with same fix #2927. So whoever will be first, will be merged. Apart from that - again thanks for contribution 🙇

Updated the changelog

packages/patrol_cli/lib/src/android/android_test_backend.dart
```diff
@@ -169,29 +169,36 @@ class AndroidTestBackend {
...
+            ],
+            if (options.buildNumber case final buildNumber?) ...[
+              '--build-number',
+              buildNumber,
+            ],
+            if (options.noTreeShakeIcons) '--no-tree-shake-icons',
+            '-t',
+            options.target,
+          ], runInShell: true)
+          .disposedBy(scope);
```
pdenert: It should be double dot, so we return to the process value returned from the `start` method
```suggestion
          ..disposedBy(scope);
```

latygindenis2: Fixed a typo, sorry about that

Issue #2894, opened 2026-01-23
[Web] Relative imports from patrol_test/ to other directories fail during web build


danielgomezriconeon:
## Description

When running Patrol tests on web (Chrome), relative imports from `patrol_test/` directory to files in other directories (e.g., `integration_test/`) fail during the Flutter web build phase. The same imports work correctly on Android.

## Environment

- **Patrol CLI version**: 4.0.2
- **patrol package version**: 4.1.0
- **Flutter version**: 3.35.5 (stable)
- **Dart version**: 3.9.2
- **Platform**: macOS (darwin)
- **Target**: Web (Chrome)

## Steps to Reproduce

1. Create a patrol test in `patrol_test/web/my_test.dart`
2. Create a helper file in `integration_test/test_utils/helper.dart`
3. Import the helper using relative path:
   ```dart
   import '../../integration_test/test_utils/helper.dart';
   ```
4. Run patrol test on web:
   ```bash
   patrol test --targets patrol_test/web/my_test.dart --device chrome
   ```

## Expected Behavior

The relative import should resolve correctly and the test should compile and run.

## Actual Behavior

The build fails with file not found errors:

```
Flutter stderr: org-dartlang-app:/web/my_test.dart:5:8: Error: Error when reading 'org-dartlang-app:/integration_test/test_utils/helper.dart': File not found
Flutter stderr: import '../../integration_test/test_utils/helper.dart';
Flutter stderr:        ^
```

## Analysis

The issue appears to be related to how Patrol builds the test bundle for web. When running `patrol test` with `--verbose`, I can see:

```
$ flutter run -d web-server --target=/path/to/project/patrol_test/test_bundle.dart ...
```

The web build resolves paths differently than the Android build. The `org-dartlang-app:` URI scheme doesn't correctly resolve the relative paths that go outside the `patrol_test/` directory.

Note the path in the error: `org-dartlang-app:/web/my_test.dart` - it's missing the `patrol_test` prefix, which explains why `../../integration_test/` doesn't resolve correctly.

## Workaround

Move patrol tests to `integration_test/patrol/` and configure `test_dir` in `pubspec.yaml`:

```yaml
patrol:
  test_dir: integration_test/patrol
  # ... other config
```

This allows relative imports to work correctly because the tests are now inside the `integration_test/` tree.

## Additional Context

- The same test files with the same imports work correctly on Android
- Regular `flutter build web` succeeds - the issue is specific to how Patrol runs web tests
[...]

PR #2906, opened 2026-01-30
Fix relative imports in tests, web

Kendru98:
Fix for https://github.com/leancodepl/patrol/issues/2894

### DIFF
--- a/.github/workflows/test-ios-simulator-webview.yaml
+++ b/.github/workflows/test-ios-simulator-webview.yaml
@@ -61,8 +61,9 @@ jobs:
 
       - name: Set tests to include
         run: |
-          TESTS_TO_INCLUDE="webview_hackernews_test,\
-          # webview_leancode_test"  # See https://github.com/leancodepl/patrol/issues/2353
+          TESTS_TO_INCLUDE="webview_hackernews_test"
+          # webview_leancode_test  
+          # See https://github.com/leancodepl/patrol/issues/2353
         
 
           target_paths=""


packages/patrol_cli/lib/src/commands/test.dart
```diff
@@ -183,6 +179,21 @@ See https://github.com/leancodepl/patrol/issues/1316 to learn more.
       return 1;
     }
 
+    final isWeb = device.targetPlatform == TargetPlatform.web;
```
piotruela: Move this line higher (above line 178) and reuse `isWeb` in line 178

packages/patrol_cli/lib/src/commands/test.dart
```diff
@@ -183,6 +179,21 @@ See https://github.com/leancodepl/patrol/issues/1316 to learn more.
...
+    final entrypoint = _testBundler.getBundledTestFile(
+      testDirectory,
+      web: isWeb,
+    );
+    if (boolArg('generate-bundle') || isWeb) {
```
piotruela: We still want to be able to not regenerate the `test_bundle.dart` file (even on web). So `|| isWeb` should be removed.

packages/patrol_cli/lib/src/test_bundler.dart
```diff
@@ -222,15 +236,25 @@ ${generateGroupsCode(testDirectory, [testFilePath]).split('\n').map((e) => '  $e
...
         testFilePath,
       );
       final testName = _createTestName(testDirectory, relativeTestFilePath);
-      final groupName = testName.replaceAll('__', '.');
+
+      // For web tests, include the test directory in the group name to match
+      // the import path structure
+      final groupName = web
+          ? '$testDirectory.${testName.replaceAll('__', '.')}'
+          : testName.replaceAll('__', '.');
```
piotruela: Is that part necessary?

PR #2924, opened 2026-02-16
Add CI/CD workflows documentation and guidelines

Kendru98:


.github/workflows/RULE.md
```diff
@@ -0,0 +1,40 @@
...
+1. **File location**: `.github/WORKFLOWS.md`
+2. **Update the following information**:
+   - Add new workflows to the appropriate section (Testing, Package Preparation, Publishing, etc.)
+   - Update workflow descriptions if behavior changes
+   - Update trigger conditions (PR paths, schedules, manual triggers)
+   - Update Flutter/Dart versions if changed
+   - Update the "Schedule Summary" section for scheduled workflows
+   - Remove entries for deleted workflows
+
+### Why this matters:
```
zoskar, line 18: I'm not sure if this section is needed.

.github/workflows/RULE.md
```diff
@@ -0,0 +1,40 @@
...
+
+- Helps team members understand the CI/CD pipeline
+- Documents which workflows run when and why
+- Makes it easier to debug and maintain workflows
+- Provides visibility into testing coverage and deployment processes
+
+### Example workflow table entry:
```
zoskar, line 25: I wonder if this makes sense: agent should figure out entry format based on contents of WORKFLOWS.md.

.github/workflows/RULE.md
Kendru98: I'm not sure either. It's something that Cursor added, not me, so I assume it's okay to leave it there. But i have no strong opinion we can also remove :D

.github/workflows/RULE.md
Kendru98: ditto

.github/workflows/RULE.md
zoskar: if it's 50/50 I'd remove it, but I leave it up to you.

PR #2925, opened 2026-02-17
Task/migrate workflows to tags instead of excludes

Kendru98:


.github/WORKFLOWS.md
```diff
@@ -105,3 +105,18 @@ These workflows verify the user has write access before running. If you don't ha
...
+
+### Tag-Based Test Selection
+
+Test workflows use a tag-based system to select which tests to run. Tests are tagged in their source files using the `tags` parameter.
+
+**Common tags:**
+- **Platform**: `android`, `ios`, `web`, `macos`
+- **Environment**: `physical_device`, `emulator`, `simulator`
+- **Features**: `webview`, `locale_testing_android`, `locale_testing_ios`
+- **Exclusions**: `android_only`, `ios_only`, `not_on_android_emulator`, `not_on_ios_simulator`, `not_on_ios_physical_device`
```
zoskar: do we need exclude tags`android_only`, `ios_only` ? Don't we use `android` or `ios` in all mobile workflows?

.github/WORKFLOWS.md
pdenert, line 1: I think that it will be cleaner if you add another column (or two) - tags and exclude tags. It can be before the description

.github/workflows/test-ios-simulator.yaml
```diff
@@ -104,7 +81,11 @@ jobs:
           sleep 10 # See https://github.com/leancodepl/patrol/issues/1282
 
           TESTS_EXIT_CODE=0
-          patrol test --exclude ${{ env.EXCLUDED_TESTS }} --verbose --full-isolation || TESTS_EXIT_CODE=$?
+          patrol test \
+          --tags='ios && simulator' \
+          --exclude-tags='webview || web || macos || android_only || not_on_ios_simulator' \
```
pdenert: Do we need that excludes? Are there any additional tests that will be run if we remove them?

pdenert (approved):
Looks good

PR #2931, opened 2026-02-22
fix: Web failures when sharding

piotruela:


packages/patrol_cli/lib/src/web/web_test_backend.dart
pdenert, line 1: It will be useful to also add comment for 'hide-test-steps' that it will be disabled when web-shard is used

Issue #2910, opened 2026-02-04
Web tests fail in headless mode due to playwright timeout


eric-odonoghue:
### Steps to reproduce

1. Fork https://github.com/JaredEzz/patrol_testing
2. Create a new branch
3. Add something and push the branch
4. Create a PR and look at the github action run results

You can also look at an existing runs failure at https://github.com/JaredEzz/patrol_testing/actions/runs/21644104235/job/62391467292

### Actual results

We are trying to setup Patrol to run flutter web tests in a [github actions script](https://github.com/JaredEzz/patrol_testing/blob/master/.github/workflows/patrol-web-test.yml). I have Patrol setup locally and it works great for the test project (on windows 11), but every time I run it via github actions it seems that Playwright times out.

```
Playwright stderr: 2026-02-03T19:18:04.147Z pw:api <= page.waitForFunction failed
Stopping Flutter web server...
Playwright process exited unexpectedly with code 1
```

### Logs

<details>
<summary>Logs</summary>

```
patrol test \
    --device chrome \
    --target patrol_test/app_test.dart \
    --web-headless=true \
    --verbose \
    --show-flutter-logs \
    --web-global-timeout=100000 \
    --web-timeout=60000
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    FLUTTER_ROOT: /opt/hostedtoolcache/flutter/stable-3.38.1-x64
    PUB_CACHE: /home/runner/.pub-cache
Verbose mode enabled. More logs will be printed.

Completion files installed. To enable completion, run the following command in your shell:
source /home/runner/.bash_profile

Received 1 test target(s)
Received test target: /home/runner/work/patrol_testing/patrol_testing/patrol_test/app_test.dart
Generated entrypoint /home/runner/work/patrol_testing/patrol_testing/patrol_test/test_bundle.dart with 1 bundled test(s)
$ flutter --no-version-check --suppress-analytics devices --machine
Received 1 device(s) to run on
Received device: Chrome (chrome)
$ flutter --suppress-analytics --no-version-check pub deps --style=list
Received 4 --dart-define(s) (0 custom, 4 internal)
Received internal --dart-define: INTEGRATION_TEST_SHOULD_REPORT_RESULTS_TO_NATIVE=false
Received internal --dart-define: PATROL_TEST_LABEL_ENABLED=true
Received internal --dart-define: PATROL_TEST_DIRECTORY=patrol_test
[...]

Issue #2832, opened 2025-12-10
Add support for custom Chrome/browser flags in patrol test command


ccrambert-ship-it:
### Use case

When running Flutter web integration tests with Patrol, developers often need to customize the browser environment to match their testing requirements. Currently, there's no way to pass custom flags to the Chrome browser launched by Playwright.

### Problem:
I'm developing a Flutter web application that connects to local backend services during integration tests. I frequently encounter these issues:

1. CORS restrictions: When testing against local Spring Boot backends running on different ports, browser CORS policies block requests even though this is a legitimate testing scenario
2. Port conflicts: The default web server port may conflict with other services running locally
3. Browser configuration: Need to disable certain security features (like web security, same-origin policy) in test environments that are safe in CI/CD but blocked by default Chrome settings

### Proposal

Add command-line options to patrol test that allow passing custom flags to the underlying browser:

# Single flag
patrol test -d chrome --web-browser-flag="--disable-web-security"

# Custom web port
patrol test -d chrome --web-port=8080

piotruela:
Sound very reasonable to me and quite easy to achieve.
We should extend `WebAppOptions` to take flags proposed in the issue description and pass command invocation inside `WebTestBackend` `build()` method.

baslogic:
We've done an audit of all Playwright options vs what Patrol currently exposes. Here's the full gap analysis:

**Patrol already supports 17 `--web-*` flags:** `headless`, `viewport`, `locale`, `timezone`, `color-scheme`, `geolocation`, `permissions`, `user-agent`, `video`, `retries`, `timeout`, `global-timeout`, `workers`, `reporter`, `shard`, `results-dir`, `report-dir`.

### Missing: Browser Launch Options ([`browserType.launch()`](https://playwright.dev/docs/api/class-browsertype))

| Proposed Flag | Playwright Option | Type | Description |
|---|---|---|---|
| `--web-browser-flag` | `args` | `string` (multi) | Raw Chromium flags (e.g. `--disable-web-security`, `--no-sandbox`) |
| `--web-channel` | `channel` | `string` | Browser channel: `chrome`, `chrome-beta`, `msedge`, etc. |
| `--web-executable-path` | `executablePath` | `string` | Path to custom browser binary |
| `--web-slow-mo` | `slowMo` | `number` | Delay (ms) between operations for debugging |
| `--web-chromium-sandbox` | `chromiumSandbox` | `bool` | Enable Chromium sandbox |
[...]

PR #2932, opened 2026-02-22
Allow passing additional args to web run

piotruela:
Fixes: #2910

Based on [this fix](https://github.com/JaredEzz/patrol_testing/pull/3#issue-3910138082) it seems that sometimes there is a need to pass some additional args to browser (above those supported by default).
This PR adds `web-browser-args` flag that will allow to do exactly that.

### DIFF
--- a/packages/patrol/web_runner/tests/setup.ts
+++ b/packages/patrol/web_runner/tests/setup.ts
@@ -4,10 +4,12 @@ import { DartTestEntry, PatrolTestEntry } from "./types"
 
 async function setup(config: FullConfig) {
   const { baseURL } = config.projects[0].use
-  const launchArgs = parseBrowserArgs()
+  const browserArgs: string[] | undefined = process.env.PATROL_WEB_BROWSER_ARGS
+    ? JSON.parse(process.env.PATROL_WEB_BROWSER_ARGS)
+    : undefined
 
   const browser = await chromium.launch({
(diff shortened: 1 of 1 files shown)


pdenert:
@piotruela maybe also add description for that in: https://patrol.leancode.co/cli-commands/test

pdenert:
Does it solve #2832?

baslogic:
The CLI to environment variable to Playwright approach here is correct and follows the established pattern. One gap though:

`parseBrowserArgs()` is only added to `setup.ts`, which handles test discovery (global setup). During actual test execution, Playwright launches its own browser instance configured by `playwright.config.ts` — and that config doesn't receive the browser args.

To complete this, `playwright.config.ts` should also read `PATROL_WEB_BROWSER_ARGS`, parse the JSON array, and pass the result into `defineConfig()` under `use.launchOptions.args` — the same way `PATROL_WEB_HEADLESS`, `PATROL_WEB_LOCALE`, etc. are already handled in that file.

This way the args apply during both discovery and test execution.

Also — re @pdenert's question, this partially addresses #2832. The full gap analysis is in [this comment](https://github.com/leancodepl/patrol/issues/2832#issuecomment-3977063860). Browser args is the most requested piece; the remaining launch/context options (channel, proxy, bypass CSP, etc.) could follow in separate PRs.

PR #2933, opened 2026-02-23
Fix typo in BrowserStack setup docs

MatejLNCD:


fylyppo:
In my project it works without "="

MatejLNCD:
@fylyppo that's weird because even the default [build.gradle.kts](https://github.com/leancodepl/patrol/blob/master/packages/patrol/example/android/app/build.gradle.kts) has
`testInstrumentationRunner = "pl.leancode.patrol.PatrolJUnitRunner"`
with the "="

Otherwise this error occurs during `patrol build`
```
Script compilation error:
	
	  Line 44:         testInstrumentationRunner "pl.leancode.patrol.BrowserstackPatrolJUnitRunner"
	                                             ^ Unexpected tokens (use ';' to separate expressions on the same line)
```

pdenert:
It depends whether we use old format (build.gradle) or new (build.gradle.kts). Since most of the projects should use new format, we can change it. But then we also have to file name in a snippet from:
`groovy title="android/app/build.gradle"`
to
 `groovy title="android/app/build.gradle.kts"`
 @MatejLNCD

PR #2935, opened 2026-02-24
Update misleading error logging for Android automator

serbandin:
The PR just fixes a misleading error log label in the Android native automator wrapper so that thrown/logged exceptions correctly reference Android rather than iOS.


pdenert:
Thanks for the contribution @serbandin. Unfortunately I didn't notice this PR and today I quick fixed same thing in #2939. So I will close this PR, but do not hesitate to do next contributions in the future. I promise to improve 🙏

PR #2942, opened 2026-02-26
Fix tests, update tags

Kendru98:


dev/e2e_app/patrol_test/camera_features/pick_image_from_gallery_test.dart
```diff
@@ -16,6 +16,8 @@ void main() {
     } else if (Platform.isIOS && !isVirtualDevice) {
       await cameraHelpers.takePhotosAcceptDialogsAndOpenAppOnRealDeviceIOS();
     }
+    // Wait for photos to be saved in gallery.
+    await Future<void>.delayed(const Duration(seconds: 10));
```
pdenert: Does it really have to be 10 seconds? Won't be 5 seconds enough?

Kendru98: Maybe 5 will be enough I decided to make 10 because this is sharded anyway, so this is quite fast.

PR #2943, opened 2026-02-26
BrowserStack image injection support for iOS

MatejLNCD:
This PR adds
 - `injectCameraPhoto` method to inject images into the iOS camera via BrowserStack's
  BrowserStackTestHelper framework
- `feedInjectedImageToViewfinder` method to continuously feed the injected image into the camera
  viewfinder, enabling QR scanner flows that rely on AVCaptureSession rather than single-shot photo
  capture
- `CameraViewfinderInjector` to support continuous image injection for QR scanner flows that rely
  on AVCaptureSession, not just single-shot photo capture
- `--browserstack` CLI flag to gate BS-specific configuration code
- QR scanner example page and test to the example app


pdenert:
@MatejLNCD please update branch

PR #2951, opened 2026-03-01
fix: disable Chrome throttling options in Playwright configuration

ahmednfwela:
chrome places some restrictions on background tabs, and this can hurt testing performance and loading time a lot, especially with large apps

https://developer.chrome.com/blog/background_tabs


piotruela:
Closing this PR as #2932 will allow passing those three flags via `--web-browser-flags`

PR #2968, opened 2026-03-07
feat(web): add multi-tab browser support for web tests

### LINKED ISSUE #2871: Multi-Windows
Hi!

Is there an some plan to support multi windows tests?

I would need to create and layout some and then launch some operations on different ones to validate a complex ui

Best wishes for 2026 patrol team

baslogic:
## Summary

Adds multi-tab browser support for Patrol web tests, enabling interaction with content across multiple browser tabs during test execution.

Resolves #2871

### New API methods on `$.platform.web`:
- `openNewTab(url:)` — open a URL in a new browser tab
- `switchToTab(tabId:)` — switch active tab by stable ID
- `closeTab(tabId:)` — close a tab and revert to the Flutter tab
- `getTabs()` — list all open tabs with IDs and URLs
- `getCurrentTab()` — get the active tab's ID
- `waitForPopup(triggerAction:, ...)` — capture a popup triggered by a DOM action

### Architecture
- `PageManager` class manages tab lifecycle with stable string IDs (`tab_0`, `tab_1`, ...)
- Uses `context.exposeBinding` (once, covers all pages) instead of per-page binding
- Existing actions (`tap`, `enterText`, `scrollTo`, etc.) automatically route to the active tab

### Test coverage
- 76 Playwright unit tests (PageManager, contracts, tab actions, handler dispatch)
- 32 Playwright integration tests (multi-tab flows, existing action compatibility)
- 4 Flutter E2E tests on real Chrome (form interaction in new tabs, cross-tab cookies, navigation coexistence)

### Also includes
- Changelog entry (`## Unreleased`) referencing #2871
- Documentation update in `docs/documentation/web.mdx`

## Test plan

- [x] Playwright unit tests pass (`npx playwright test --config playwright.unit.config.ts`)
- [x] Playwright integration tests pass (`npx playwright test --config playwright.integration.config.ts`)
- [x] Flutter E2E tests pass (`patrol test --target patrol_test/web/ -d chrome`)
- [ ] Android emulator CI — expected to fail for contributors without write access (per CONTRIBUTING.md)

### DIFF
--- a/dev/e2e_app/patrol_test/web/web_example_multi_tab_navigation_test.dart
+++ b/dev/e2e_app/patrol_test/web/web_example_multi_tab_navigation_test.dart
@@ -1,0 +1,53 @@
+import '../common.dart';
+
+import 'web_example_app.dart';
+
+void main() {
+  patrol('interact with form in new tab while navigating Flutter app',
+      ($) async {
+    await $.pumpWidgetAndSettle(const WebExampleApp());
+
+    final formUrl = '${Uri.base.origin}/assets/assets/iframe_content.html';
+    final newTabId = await $.platform.web.openNewTab(url: formUrl);
+    await $.pumpAndSettle();
+    await Future<void>.delayed(const Duration(seconds: 2));
+
+    final tabsResult = await $.platform.web.getTabs();
+    final tabs = tabsResult['tabs'] as List<Object?>;
+    expect(tabs.length, 2);
+
+    // Fill form in the new tab
+    await $.platform.web.switchToTab(tabId: newTabId);
+    await $.pumpAndSettle();
+    await Future<void>.delayed(const Duration(seconds: 1));
+
+    await $.platform.web.enterText(
+      WebSelector(cssOrXpath: '#test-input'),
+      text: 'Patrol multi-tab test',
+    );
+    await $.platform.web.tap(WebSelector(cssOrXpath: '#submit-button'));
+    await Future<void>.delayed(const Duration(seconds: 1));
+
+    // Switch back and navigate the Flutter app
+    await $.platform.web.switchToTab(tabId: 'tab_0');
+    await $.pumpAndSettle();
+
+    await $('Go to Page 1').scrollTo().tap();
+    await $.pumpAndSettle();
+    await Future<void>.delayed(const Duration(seconds: 2));
+
+    expect($('This is Page 1'), findsOneWidget);
+
+    await $.platform.web.goBack();
+    await $.pumpAndSettle();
+    await Future<void>.delayed(const Duration(seconds: 2));
+
+    expect($('This is the home page'), findsOneWidget);
+
+    await $.platform.web.closeTab(tabId: newTabId);
+
+    final finalTabs = await $.platform.web.getTabs();
+    final remaining = finalTabs['tabs'] as List<Object?>;
+    expect(remaining.length, 1);
+  });
+}
--- a/dev/e2e_app/patrol_test/web/web_example_multi_tab_test.dart
+++ b/dev/e2e_app/patrol_test/web/web_example_multi_tab_test.dart
@@ -1,0 +1,101 @@
+import 'dart:collection';
+
+import '../common.dart';
+
+import 'web_example_app.dart';
+
+void main() {
+  patrol('open external page, fill form, and return to Flutter app', ($) async {
+    await $.pumpWidgetAndSettle(const WebExampleApp());
+
+    final formUrl = '${Uri.base.origin}/assets/assets/iframe_content.html';
+    final newTabId = await $.platform.web.openNewTab(url: formUrl);
+    await $.pumpAndSettle();
+    await Future<void>.delayed(const Duration(seconds: 2));
+
+    await $.platform.web.switchToTab(tabId: newTabId);
+    await $.pumpAndSettle();
+    await Future<void>.delayed(const Duration(seconds: 1));
+
+    final currentTab = await $.platform.web.getCurrentTab();
+    expect(currentTab, newTabId);
+
+    await $.platform.web.enterText(
+      WebSelector(cssOrXpath: '#test-input'),
+      text: 'Hello from new tab',
+    );
+    await $.platform.web.tap(WebSelector(cssOrXpath: '#submit-button'));
+    await Future<void>.delayed(const Duration(seconds: 1));
+
+    await $.platform.web.switchToTab(tabId: 'tab_0');
+    await $.pumpAndSettle();
+
+    expect(await $.platform.web.getCurrentTab(), 'tab_0');
+    expect($('This is the home page'), findsOneWidget);
+
+    await $.platform.web.closeTab(tabId: newTabId);
+  });
+
+  patrol('open leancode.co, accept cookies, and close tab', ($) async {
+    await $.pumpWidgetAndSettle(const WebExampleApp());
+
+    final newTabId = await $.platform.web.openNewTab(
+      url: 'https://leancode.co/',
+    );
+    await $.pumpAndSettle();
+    await Future<void>.delayed(const Duration(seconds: 3));
+
+    await $.platform.web.switchToTab(tabId: newTabId);
+    await $.pumpAndSettle();
+    await Future<void>.delayed(const Duration(seconds: 2));
+
+    await $.platform.web.tap(WebSelector(text: 'Accept All'));
+    await Future<void>.delayed(const Duration(seconds: 1));
+
+    await $.platform.web.switchToTab(tabId: 'tab_0');
+    await $.pumpAndSettle();
+
+    expect($('This is the home page'), findsOneWidget);
+
+    await $.platform.web.closeTab(tabId: newTabId);
+  });
+
+  patrol('cross-tab cookies persist across browser context', ($) async {
+    await $.pumpWidgetAndSettle(const WebExampleApp());
+
+    await $.platform.web.addCookie(
+      name: 'session',
+      value: 'abc123',
+      url: Uri.base.origin,
+    );
+
+    var cookies = await $.platform.web.getCookies();
+    var sessionCookie = cookies.firstWhere(
+      (c) => c['name'] == 'session',
+      orElse: LinkedHashMap<Object?, Object?>.new,
+    );
+    expect(sessionCookie['value'], 'abc123');
+
+    final formUrl = '${Uri.base.origin}/assets/assets/iframe_content.html';
+    final newTabId = await $.platform.web.openNewTab(url: formUrl);
+    await $.pumpAndSettle();
+    await Future<void>.delayed(const Duration(seconds: 2));
+
+    await $.platform.web.switchToTab(tabId: newTabId);
+    await $.pumpAndSettle();
+    await Future<void>.delayed(const Duration(seconds: 1));
+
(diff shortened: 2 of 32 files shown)


packages/patrol/web_runner/tests/patrolPlatformHandler.ts
```diff
@@ -1,15 +1,16 @@
...
+import { PageManager } from "./pageManager"
 
-export async function exposePatrolPlatformHandler(page: Page) {
-  await page.exposeBinding("__patrol__platformHandler", async ({ page }, request) =>
-    handlePatrolPlatformAction(page, request),
+export async function exposePatrolPlatformHandler(context: BrowserContext, pageManager: PageManager) {
+  await context.exposeBinding("__patrol__platformHandler", async (_source, request) =>
+    handlePatrolPlatformAction(pageManager, request),
   )
 }
```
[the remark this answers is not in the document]
baslogic: Fixed — added a `tab_0` guard on the binding. Only the main Flutter test page can invoke the platform handler now. External pages opened during tests will get an error if they attempt to call it.

packages/patrol/web_runner/tests/patrolPlatformHandler.ts
```diff
@@ -1,15 +1,16 @@
...
 
-async function handlePatrolPlatformAction(page: Page, { action, params }: PatrolNativeRequest) {
+export async function handlePatrolPlatformAction(pageManager: PageManager, { action, params }: PatrolNativeRequest) {
   logger.info(params, `Received action: ${action}`)
```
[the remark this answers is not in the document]
baslogic: This is an existing pattern — the handler already logged params before this PR. Changing the logging behavior across the codebase is outside the scope of this feature. Worth a separate issue if the maintainers want to address it.

dev/e2e_app/patrol_test/web/web_example_multi_tab_navigation_test.dart
```diff
@@ -0,0 +1,55 @@
...
+
+void main() {
+  patrol('interact with form in new tab while navigating Flutter app',
+      ($) async {
+    await $.pumpWidgetAndSettle(const WebExampleApp());
+
+    final formUrl = '${Uri.base.origin}/assets/assets/iframe_content.html';
+    final newTabId = await $.platform.web.openNewTab(url: formUrl);
+    await $.pumpAndSettle();
+    await Future<void>.delayed(const Duration(seconds: 2));
```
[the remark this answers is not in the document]
baslogic: The `Future.delayed` pattern is consistent with existing Patrol web E2E tests (e.g., `web_example_iframe_interactions_test.dart`, `webview_hackernews_test.dart`). Replacing these with condition-based waits would be a broader improvement across the test suite, not specific to this PR.

dev/e2e_app/patrol_test/web/web_example_multi_tab_test.dart
```diff
@@ -0,0 +1,101 @@
...
+import 'web_example_app.dart';
+
+void main() {
+  patrol('open external page, fill form, and return to Flutter app', ($) async {
+    await $.pumpWidgetAndSettle(const WebExampleApp());
+
+    final formUrl = '${Uri.base.origin}/assets/assets/iframe_content.html';
+    final newTabId = await $.platform.web.openNewTab(url: formUrl);
+    await $.pumpAndSettle();
+    await Future<void>.delayed(const Duration(seconds: 2));
```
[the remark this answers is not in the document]
baslogic: Same as above — following the established pattern used across Patrol's existing E2E tests.

packages/patrol/lib/src/platform/web/web_automator.dart
```diff
@@ -88,4 +88,28 @@ abstract interface class WebAutomator {
...
+
+  /// Closes the tab with the given [tabId].
+  Future<void> closeTab({required String tabId});
+
+  /// Switches the active tab to [tabId].
+  /// All subsequent actions will target this tab until switched again.
+  Future<void> switchToTab({required String tabId});
+
+  /// Returns information about all open tabs.
+  Future<LinkedHashMap<Object?, Object?>> getTabs();
```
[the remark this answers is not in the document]
baslogic: Good suggestion. The `LinkedHashMap<Object?, Object?>` return type matches the existing pattern used by `getCookies()` in the same interface. Introducing a typed `TabsInfo` class would be cleaner but would diverge from the current convention. Happy to change this if the maintainers prefer a typed approach.

packages/patrol/web_runner/tests/actions/waitForPopup.ts
```diff
@@ -0,0 +1,23 @@
...
+  context: BrowserContext,
+): Promise<string> {
+  const actionFn = actions[params.triggerAction as keyof typeof actions]
+  if (!actionFn) {
+    throw new Error(`Unknown trigger action: "${params.triggerAction}"`)
+  }
+
+  const [newPage] = await Promise.all([
+    context.waitForEvent("page"),
+    (actionFn as Function)(page, params.triggerParams, pageManager, context).catch(() => {}),
```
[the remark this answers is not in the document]
baslogic: Fixed — the `.catch(() => {})` has been replaced with error logging via `logger.warn()`. Trigger action failures are now visible in the output instead of being silently swallowed. Also replaced the `!` non-null assertion with a defensive check that throws a descriptive error.

packages/patrol/lib/src/platform/web/web_automator_native.dart
```diff
@@ -283,4 +283,71 @@ class WebAutomator implements web_automator.WebAutomator {
...
+      {},
+      logger: _config.logger,
+      patrolLog: _patrolLog,
+    );
+    return result as String;
+  }
+
+  @override
+  Future<String> waitForPopup({
+    required String triggerAction,
```
piotruela, line 342: As much as possible, I think it’s better to take inspiration from how the dismissNextDialog method is implemented.
We call it before the action that is supposed to show the dialog, then we perform the action that triggers the dialog to appear, and in the next step we await it. I think a similar approach should work in this case.

What seems particularly suboptimal to me in this approach is passing the action that triggers opening the popup as a `String`.

packages/patrol/lib/src/platform/web/web_automator.dart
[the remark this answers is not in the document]
piotruela: I like what Gemini suggested. Would also be great to get some more info about the tab so that it will be easier to identify them. Maybe `url` is something we can get?

piotruela (commented):
Amazing job 👏

packages/patrol/web_runner/tests/actions/closeTab.ts
```diff
@@ -0,0 +1,15 @@
+import type { Page } from "playwright"
+import type { CloseTabRequest } from "../contracts"
+import type { PageManager } from "../pageManager"
+
+export async function closeTab(
+  _page: Page,
+  params: CloseTabRequest["params"],
+  pageManager: PageManager,
+): Promise<void> {
+  if (params.tabId === "tab_0") {
```
konowrockis, line 10: If `tab_0` requires special treatment the logic should be moved to `PageManager`.

Maybe it should be like `pageManager.close(params.tabId)` or at least `pageManager.isProtected(params.tabId)`

packages/patrol/web_runner/tests/actions/closeTab.ts
```diff
@@ -0,0 +1,15 @@
+import type { Page } from "playwright"
+import type { CloseTabRequest } from "../contracts"
+import type { PageManager } from "../pageManager"
+
+export async function closeTab(
+  _page: Page,
+  params: CloseTabRequest["params"],
+  pageManager: PageManager,
+): Promise<void> {
```
konowrockis, line 9: No need to explicitly type return value

packages/patrol/web_runner/tests/actions/openNewTab.ts
```diff
@@ -0,0 +1,14 @@
+import type { BrowserContext, Page } from "playwright"
+import type { OpenNewTabRequest } from "../contracts"
+import type { PageManager } from "../pageManager"
+
+export async function openNewTab(
+  _page: Page,
+  params: OpenNewTabRequest["params"],
+  pageManager: PageManager,
+  context: BrowserContext,
```
konowrockis, line 9: Hmm, I'm wondering whether doing just that is good idea because we're extending on already implicit assumption 🤔 
Contract for action was already implicit that it received `(page: Page, params: TParams)` but up until now every action had the same assumptions. Now that we are changing that maybe we should structure that better and make it explicit.
What do you think about having something like
```ts
type Action<TParams, TResult> = (params: { page: Page, params: TParams, pageManager: PageManager, context: BrowserContext }) => Promise<TResult>

export const openNewTab: Action<OpenNewTabRequest["params"], string> = async ({ params, pageManager, context}) => { ... }
```
This would require refactor of all the other actions as well but I think now is the time for that.

packages/patrol/web_runner/tests/pageManager.ts
```diff
@@ -0,0 +1,70 @@
+import type { Page, BrowserContext } from "playwright"
+
+export class PageManager {
+  private _registry = new Map<string, Page>()
```
konowrockis, line 4: nitpick: prefixing private things with `_` feels very _`dart`y_. In typescript [more idiomatic](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements) would be `#registry` (or just `private registry` is fine IMO)

packages/patrol/web_runner/tests/pageManager.ts
```diff
@@ -0,0 +1,70 @@
...
+  private _reverse = new Map<Page, string>()
+  private _activeId: string
+  private _nextIndex = 0
+  readonly context: BrowserContext
+
+  constructor(context: BrowserContext, initialPage: Page) {
+    this.context = context
```
konowrockis, line 11:
```suggestion
  constructor(readonly context: BrowserContext, initialPage: Page) {
```

packages/patrol/web_runner/tests/pageManager.ts
```diff
@@ -0,0 +1,70 @@
...
+    this._register(initialPage)
+    this._activeId = "tab_0"
```
konowrockis, line 13: `tab_0` should be extracted to some const

packages/patrol/web_runner/tests/contracts.ts
```diff
@@ -1,6 +1,6 @@
 type PatrolNativeRequestBase<TAction extends string, TParams> = {
   action: TAction
-  params: TParams
+  params: TParams & { _routeToTab?: string }
```
konowrockis, line 3: Hmm... I'm not a fan of this approach. IMO this should be explicitly added to params which only can make use of this. For example cookies related things shouldn't care for page, maybe clipboard as well.

konowrockis (changes requested):
In this PR we're mixing `tab` and `page`. For example we have `PageManager` but `tabId` or `routeToTab`. I think we should stick to pages as it's more native to playwright.

Great job overall! Thanks! Great reason to stop procrastinating 😅

pdenert:
@baslogic thanks for contribution. We appreciate it 👏 Will you plan to continue working on this PR?

pdenert:
Merged after changes in #3065

PR #2970, opened 2026-03-09
fix: Reflect failed tests in Playwight report

piotruela:
Failed tests were not marked as failed in Playwright report (only in Patrol log summary at the end of the execution)


pdenert (commented):
Update the changelog

PR #2976, opened 2026-03-13
Swift Package Manager support

MatejLNCD:
## Summary                                                                                           
                                
- Add a `Package.swift` manifest for the Patrol plugin, enabling Flutter's native SPM integration as 
an alternative to CocoaPods     
- Restructure source files into SPM-compatible layout (`Sources/patrol/`, `Sources/HTTPParserC/`), since SPM doesn't support mixing C and Swift, while maintaining CocoaPods compatibility by updating podspec paths
- Configure XCTest weak framework linking with dynamic Xcode developer directory resolution, so the
plugin builds correctly under SPM without hard-coded paths
- Update the example app's Xcode project to use `FlutterGeneratedPluginSwiftPackage` and add
`HEADER_SEARCH_PATHS` for the patrol ObjC headers
- Add Flutter SPM pre-action build scripts to both `dev` and `prod` schemes
- Handle SPM/CocoaPods divergence in `RunnerUITests.m` with a conditional `#import` fallback for the
`PATROL_INTEGRATION_TEST_IOS_RUNNER` macro

## Details

The plugin now supports both CocoaPods and SPM side-by-side:

- **CocoaPods path**: `patrol.podspec` updated to point at the new source locations under
`patrol/Sources/`
- **SPM path**: New `Package.swift` defines two targets (`HTTPParserC` for the C HTTP parser,
`patrol` for all Swift sources) with `CocoaAsyncSocket` as an external dependency
- Swift files that previously relied on implicit imports now explicitly `import Foundation`
- Example app analysis options updated to exclude `.build/` directories


packages/patrol/darwin/patrol/Package.swift
```diff
@@ -0,0 +1,74 @@
...
+    platforms: [
+        .iOS("13.0"),
+        .macOS("10.14")
+    ],
+    products: [
+        .library(name: "patrol", targets: ["patrol"])
+    ],
+    dependencies: [
+        // TODO: Add this when Patrol targets Flutter 3.41.0 or higher
+        // .package(name: "FlutterFramework", path: "../FlutterFramework")
```
PiotrRogulski: Why will this be needed? 🤔

mafreud: @MatejLNCD Could you reply this question? A lot of users, including myself are waiting for SPM support!

Peetee06: @PiotrRogulski do you have any other channel over which you can reach @MatejLNCD to ping him? I saw Matej at the Flutter Tech Summit, so I assume LeanCode has a way to reach him. 

Would be great if we can prevent any issues once Swift Package Manager becomes required

olekeke999: @Peetee06 this is a new requirement since flutter 3.41 https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors#:~:text=New%20in%20Flutter%203.41!%20Add%20the%20FlutterFramework please look for ‘ FlutterFramework’ on this page.
Currently works without it well on the new versions (I’m on 3.44.1).
But it displays in the build logs message about adding it in Package.swift

pdenert: @mafreud @Peetee06 @olekeke999 this will be handled in the upcoming week

MatejLNCD: Looking at this now, no worries guys, I was busy with other projects and overlooked this conversation

MatejLNCD: The previous version of the docs linked by @olekeke999 contained this text

> **New in Flutter 3.41!** Add the FlutterFramework as a dependency and update Dart/Flutter version.

but now that line was changed and it says

> Add the `FlutterFramework` as a dependency and update Dart and Flutter versions.

So either way, this dependency needs to be here.

packages/patrol_gen/lib/src/generators/darwin/darwin_url_session_client_generator.dart
```diff
@@ -110,8 +112,9 @@ $endpoints
   }
 
   String _createEndpoint(Endpoint endpoint) {
-    final requestDef =
-        endpoint.request != null ? 'request: ${endpoint.request!.name}' : '';
+    final requestDef = endpoint.request != null
+        ? 'request: ${endpoint.request!.name}'
+        : '';
```
piotruela: nitpick (optional): Get rid of bang operator by using switch expression

packages/patrol_gen/lib/src/generators/darwin/darwin_telegraph_server_generator.dart
```diff
@@ -43,10 +45,12 @@ $endpoints
...
-    final request =
-        endpoint.request != null ? 'request: ${endpoint.request!.name}' : '';
-    final response =
-        endpoint.response != null ? ' -> ${endpoint.response!.name}' : '';
+    final request = endpoint.request != null
+        ? 'request: ${endpoint.request!.name}'
+        : '';
+    final response = endpoint.response != null
+        ? ' -> ${endpoint.response!.name}'
+        : '';
```
piotruela: nitpick (optional): Get rid of bang operator by using switch expression

packages/patrol_gen/lib/src/resolve_schema.dart
```diff
@@ -112,38 +109,40 @@ Message _createMessage(ClassDeclaration declaration) {
         .map((e) => e.fields)
         .whereType<VariableDeclarationList>()
         .map((e) {
```
piotruela, line 111:
```suggestion
        .map((e) {
          final type = e.type;
          if (type is! NamedType) {
            throw UnsupportedError('unsupported type $type');
          }

          final arguments = type.typeArguments?.arguments;

          final fieldType = switch (type.type) {
            final t? when t.isDartCoreMap => switch (arguments) {
              [final NamedType key, final NamedType value] => MapFieldType(
                keyType: key.name2.lexeme,
                valueType: value.name2.lexeme,
              ),
              _ => throw UnsupportedError('unsupported map type $type'),
            },
            final t? when t.isDartCoreList => switch (arguments) {
              [final NamedType element, ...] => ListFieldType(
                type: element.name2.lexeme,
              ),
              _ => throw UnsupportedError('unsupported list type $type'),
            },
            _ => OrdinaryFieldType(type: type.name2.lexeme),
          };

          return MessageField(
            isOptional: type.question != null,
            name: e.variables.first.name.lexeme,
            type: fieldType,
          );
        })

```

packages/patrol_gen/lib/src/generators/android/android_http4k_server_generator.dart
```diff
@@ -53,36 +53,41 @@ $routes
...
-      return '    fun ${endpoint.name}($request)$response';
-    }).join('\n');
+    return service.endpoints
+        .map((endpoint) {
+          final response = endpoint.response != null
+              ? ': Contracts.${endpoint.response!.name}'
+              : '';
+          final request = endpoint.request != null
+              ? 'request: Contracts.${endpoint.request!.name}'
+              : '';
```
piotruela: ditto

packages/patrol/example/ios/RunnerUITests/RunnerUITests.m
```diff
@@ -2,4 +2,10 @@
 @import patrol;
 @import ObjectiveC.runtime;
 
+// With CocoaPods, @import patrol exports ObjC headers including this macro.
+// With SPM, @import patrol only exports Swift APIs, so we need explicit #import.
```
piotruela: Is this comment still true with our Clang umbrella approach?
Will there be a migration step for projects that are currently on CocoaPods and already have patrol?

piotruela (approved):

pdenert (approved):

PR #2992, opened 2026-03-20
fix: pass appId

yurtemre7:
Hey Patrol Team,

thanks for your time and effort put into this cool testing framework.
While working today with Patrol I've encountered a _small bug_ with the `MobileAutomator` class' passing of the `String? appId` parameter. In some methods, it was simply forgotten to pass that parameter down the underlying iOS method.

We encountered the following issue in that way:

1. Be on a iOS App
2. Have some logic invoking the iOS local auth (Pin Code / Passcode).
3. Try entering in the special TextField outside of the Flutter app via:

```dart
$.platform.mobile.enterTextByIndex(x, index: 0, appId: "com.apple.springboard"); 
```
You will see the test throwing an error 400 no TextField found and it will display your own apps appId instead of the provided one.

4. If you call the underlying iOS method instead:

```dart
$.platform.ios.enterTextByIndex(x, index: 0, appId: "com.apple.springboard"); 
``` 
it will work because the iOS method does pass the appId down to the next method.

Out of completion reasons I've also updated that inside of the NativeAutomator2 class, as it was lacking there too, but I do understand that this class is marked deprecated and will be removed in the future. I can of course also remove that fix there if it is not needed.

I hope I do not break any rules, I've read the Contributing guideline and did not saw another PR/issue declared.

Thanks again for this package and the framework to make writing flutter tests easier for everyone.

### DIFF
--- a/packages/patrol/CHANGELOG.md
+++ b/packages/patrol/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.4.1
+
+- Fix `appId` not being passed down on `$.platform.mobile.enterText` and `$.platform.mobile.enterTextByIndex` (#2992)
+
+
 ## 4.4.0
 - Add support for emulators with API lvl 36 for pickMultipleImagesFromGallery method. (#2977)
 - Fix `tapOnNotification` on iOS physical devices to Open button when tapping on notification. (#2972)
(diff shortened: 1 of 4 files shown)


yurtemre7:
@pdenert done!

pdenert:
In future PRs (which I hope there will be), don’t change the version in the pubspec; instead, use “Unreleased” in the changelog

PR #2996, opened 2026-03-24
chore: Please pana for patrol_cli

MatejLNCD:


packages/patrol_cli/pubspec.yaml
pdenert, line 1: You can add to changelog:
Updating dependencies

Issue #2769, opened 2025-11-03
Run patrol build once and execute individual tests separately


leonardmatasel-dinova:
### Use case

Currently, running `patrol test` rebuilds the app. This makes it harder to isolate integration tests, because resetting the emulator state between tests is cumbersome when each test triggers a new build.

I would like to:

1. Run the `patrol build` once for all tests.
2. Launch individual tests with `patrol test` separately without rebuilding the app each time.

This would allow:

- Faster test execution, as the app is built only once.
- Simpler isolation of integration tests, since individual tests can be launched using tags, and scripts in CI/CD pipelines can restore a default emulator snapshot before each test.

Alternative solution:
- Ability to run scripts before or after each test execution.

### Proposal

Add a feature to patrol test that allows:

- Specifying a single build to use for multiple test runs.
- Selecting which tests to run via a tag

Example CLI usage:

```
# Build once
patrol build

# Run individual tests using the prebuilt app
patrol test --use-existing-build --tags login

# Execute same scripts
load_snapshot default_root

patrol test --use-existing-build --tags signup
```

zoskar:
Hi @leonardmatasel-dinova
That feature would be great, but there are many technical blockers ATM that prevent us from easily developing this feature. Hopefully we'll work them out one by one.

And to not leave you empty handed here's a sneak peek for feature that gets us one step closer to your request
https://github.com/leancodepl/patrol/pull/2734
We're planning to release is soon :)

PR #3017, opened 2026-03-30
Add option to skip build step in test command

Mcluky:
This pull request adds support for skipping the build step when running patrol test by introducing a new --no-build flag. This allows users to separate the build and test steps, which is especially useful in CI pipelines or iterative workflows where the app is built once and tested multiple times.

Implements what is mentioned in https://github.com/leancodepl/patrol/issues/2769


piotruela:
#3197 added a way to skip build step for already compiled app + tests.
See more about how to use it in docs: https://patrol.leancode.co/documentation/ci/build-time-test-discovery

PR #3034, opened 2026-04-21
Note additional setup requirement in patrol README

### LINKED ISSUE #3031: Missing Android Test Orchestrator silently breaks multi-test runs with an unhelpful 500 error
# Missing Android Test Orchestrator silently breaks multi-test runs with an unhelpful 500 error

## Summary

Patrol requires Android Test Orchestrator to run more than one test per invocation. That requirement isn't in the README, isn't added by `patrol bootstrap`, and isn't checked at runtime. When it's missing, the first test passes, every subsequent test fails with `PatrolAppServiceClientException: Invalid response 500, Internal Server Error`, and `patrol test`'s own stdout reports "All tests passed!" — so the failure isn't visible unless you read the JUnit XML directly.

I've already worked around this locally. Filing so the next person who hits it gets a clearer signal.

## Repro

1. `flutter create` an Android app and add `patrol` to it.
2. Follow the README quickstart. It tells you to set `testInstrumentationRunner = "pl.leancode.patrol.PatrolJUnitRunner"` in `build.gradle.kts`. It does not tell you to configure Android Test Orchestrator.
3. Add two tests in two separate files and run `patrol test`.

Observed output:

```
Test summary:
📝 Total: 1
✅ Successful: 1
❌ Failed: 0
⏱️  Duration: 20s

✗ Failed to execute tests of apk ... (Gradle test execution failed with code 1)
```

Patrol says one test ran and passed. Gradle exits 1. If you dismiss the exit code as the known `gradle/gradle#11195` teardown flake (which is common advice), you can ship the broken run.

The truth is in `build/app/outputs/androidTest-results/connected/debug/TEST-*.xml`:

```xml
<testsuite name="com.example.app.MainActivityTest" tests="5" failures="4" errors="0">
  <testcase name="runPatrolTest[first_test ...]" time="3.407" />
  <testcase name="runPatrolTest[second_test ...]" time="0.031">
    <failure>java.lang.RuntimeException: PatrolAppServiceClientException: Invalid response 500, Internal Server Error
      at pl.leancode.patrol.PatrolJUnitRunner.runDartTest(PatrolJUnitRunner.java:164)
      ...
    </failure>
  </testcase>
  <!-- three more identical failures -->
</testsuite>
```

Five tests ran. Four silently failed. Patrol's own summary showed zero failures.

## Why it happens

`patrol/lib/src/platform/mobile/patrol_app_service_io.dart` (4.3.0) holds two `final` `Completer`s that are single-use:

```dart
final _testExecutionRequested = Completer<String>();
final _testExecutionCompleted = Completer<_TestExecutionResult>();
```

On the second `runDartTest` call, both are already completed, the pipeline returns HTTP 500, and the native side surfaces it as `PatrolAppServiceClientException`.

Android Test Orchestrator masks this because it runs each `@Test` method in a fresh instrumentation process, so `PatrolAppService` is rebuilt per test. Without orchestrator, everything shares one process and the single-use completers guarantee failure starting at test 2.

So orchestrator isn't an optional optimization — it's load-bearing. The example app does configure it: `patrol-4.3.0/example/android/app/build.gradle.kts:44-50,65` sets `execution = "ANDROIDX_TEST_ORCHESTRATOR"`, `clearPackageData = "true"`, and `androidTestUtil("androidx.test:orchestrator:1.5.1")`. Consumers who don't diff their gradle config against the example app have no clear signal that these lines are required rather than exemplary.

## Why it's hard to notice

1. **Patrol's stdout summary reports `Total: 1, Successful: 1, Failed: 0`** even when AGP ran five tests and four crashed. I think it's counting Dart `test()` invocations it observed complete, not AGP test methods. Skimming the log suggests the run was green.
2. **The exit code is 1**, which matches the known gradle/gradle#11195 teardown flake. A common workaround for that flake is to grep the stdout summary for `Failed: 0` and return success — which, combined with point 1, can mask real failures entirely. We shipped exactly that workaround for a while and it silently let broken runs through our CI before we switched to parsing the JUnit XML.
3. **`patrol test` prints no warning** when the host project is missing orchestrator config.
4. **#2879** described this exact symptom ("Disabling Orchestrator causes 500 Internal Server Error for all tests except the first one") and was closed as a duplicate of #1839, but #1839 is a different bug (forward slash in test descriptions). Might be worth reopening.
5. **#2698** is an open report of the same root cause in an `adb am instrument` context. The maintainer reply ("This command will throw if there are multiple tests in the apk") is accurate for that user's script, but the underlying cause is the same as the plain `patrol test`-without-orchestrator case.

## What might help

Any one of these would close the gap; together they'd close it completely. Happy to defer to whatever fits Patrol's roadmap best.

**1. Fail early at runtime.** In `PatrolJUnitRunner.setUp(...)`, check whether the instrumentation is running under Android Test Orchestrator. If not, throw with a clear message:

```
PatrolJUnitRunner: Android Test Orchestrator is not configured. Patrol requires
the orchestrator to run more than one test per invocation. Add the following to
your app's build.gradle(.kts):

  android {
    defaultConfig {
      testInstrumentationRunnerArguments["clearPackageData"] = "true"
    }
    testOptions {
      execution = "ANDROIDX_TEST_ORCHESTRATOR"
    }
  }
  dependencies {
    androidTestUtil("androidx.test:orchestrator:1.5.1")
  }

See https://github.com/leancodepl/patrol/blob/main/example/android/app/build.gradle.kts
for the reference configuration.
```

The check itself can be a one-liner — `InstrumentationRegistry.getArguments().getString("orchestratorService") != null` is a reasonable signal.

**2. Add the orchestrator config to `patrol bootstrap`.** Those three lines could land in the consumer's `build.gradle.kts` the same way the test runner class does today. New projects would work out of the box; existing ones would get the missing config when they rerun `patrol bootstrap`.

**3. Mention the requirement in the README.** A couple of sentences next to the quickstart's `testInstrumentationRunner = PatrolJUnitRunner` instruction, pointing at the example app's gradle config, would make the requirement discoverable.

## Environment

- `patrol`: 4.3.0
- `patrol_cli`: 4.3.0
- Flutter: 3.41.4 (Dart 3.11.1)
- AGP: 8.11.1
- Device: Pixel 7 API 34 AVD (headless)

## Related

- #2879 — same symptom, closed as duplicate of #1839; the two appear unrelated.
- #2698 — open report of the same root cause via `adb am instrument`.

Happy to help verify a fix against our setup if useful. Thanks for maintaining Patrol — it's been a good tool to work with overall.

zoskar:
Closes https://github.com/leancodepl/patrol/issues/3031

User reported issue which could be avoided if patrol readme on installation step would point to setup in docs. I've also added CLI mention as this was reported by one mobile dev (🌲) that the relation was not clear

### DIFF
--- a/packages/patrol/README.md
+++ b/packages/patrol/README.md
@@ -49,7 +49,7 @@ to only use custom finders, check out
 $ dart pub add patrol --dev
 ```
 
-> ⚠️ Additional setup is required to run tests — install [`patrol_cli`][patrol_cli_link] and complete native setup. See the [Patrol documentation][docs_setup].
+> ⚠️ Additional setup is required to run tests – install [`patrol_cli`][patrol_cli_link] and complete [native setup][docs_setup].
 
 ## Usage
 
--- a/packages/patrol/README.md
+++ b/packages/patrol/README.md
@@ -49,6 +49,8 @@ to only use custom finders, check out
 $ dart pub add patrol --dev
 ```
 
+> ⚠️ Additional setup is required to run tests — install [`patrol_cli`][patrol_cli_link] and complete native setup. See the [Patrol documentation][docs_setup].
+
 ## Usage
 
 Patrol has 2 main features – [native automation] and [custom finders].
@@ -168,6 +170,7 @@ We are **top-tier experts** focused on Flutter Enterprise solutions.
 [patrol_github_link]: https://github.com/leancodepl/patrol
 [patrol_discord_link]: https://discord.gg/ukBK5t4EZg
 [docs]: https://patrol.leancode.co/?utm_source=github.com&utm_medium=referral&utm_campaign=patrol-readme
+[docs_setup]: https://patrol.leancode.co/documentation?utm_source=github.com&utm_medium=referral&utm_campaign=patrol-readme
 [docs_finders]: https://patrol.leancode.co/finders/overview
 [promo_graphics]: ../../assets/promo_banner.png
 [article_web]: https://leancode.co/blog/patrol-web-support?utm_source=github.com&utm_medium=referral&utm_campaign=patrol-readme


packages/patrol/README.md
```diff
@@ -49,6 +49,8 @@ to only use custom finders, check out
 $ dart pub add patrol --dev
 ```
 
+> ⚠️ Additional setup is required to run tests – install [`patrol_cli`][patrol_cli_link] and complete [native setup][docs_setup].
+
```
pdenert, line 53: Let's just redirect for whole setup to the docs, so we should remove:

```console
$ dart pub add patrol --dev
```

pdenert (approved):

PR #3040, opened 2026-04-27
feat: enable patrol develop on web and harden MCP session

piotruela:
`patrol develop -d chrome` now hot restarts the app in place instead of relaunching the browser on every `r`, and the Patrol DevTools extension works on web. Shipped as a `-dev.1` prerelease.

Web Hot Restart needs Flutter 3.47 ([flutter/flutter#183838](https://github.com/flutter/flutter/pull/183838)). Declaring that in `patrol`'s pubspec would force 3.47 on every Android and iOS user of this line, and break `pub get` for this repo's own workflows, which are pinned to 3.38.x. So the floor stays out of the pubspec: `patrol_cli` reads the SDK version when a web develop session starts and exits with a message naming it, everything else keeps working on 3.32. Releasing as a prerelease means people who want web develop opt in with `^4.11.0-dev.1` and pick up Flutter 3.47 themselves, while the stable line is untouched.

- One `flutter run` and one Playwright driver stay alive for the whole session, so a restart lands in roughly 130 to 320 ms instead of a full rebuild.
- The DevTools inspector renders the page's DOM, where the `flt-semantics` elements carry the roles and labels finders match; `--open-devtools` serves it through patrol's own Dart Tooling Daemon, because DevTools can't discover extensions for a web app ([flutter/devtools#9667](https://github.com/flutter/devtools/pull/9667)).
- MCP treats web as a develop platform: `devices` lists browsers and `run` accepts `device: "chrome"`. Web is never auto-selected.
- A session whose `flutter run` dies during startup now fails with a message, instead of exiting 255 and leaving the terminal in raw mode with an orphaned Playwright driver.
- `patrol` 4.11.0-dev.1, `patrol_cli` 4.9.0-dev.1, `patrol_mcp` 0.3.0-dev.1, `patrol_devtools_extension` 0.5.0-dev.1. Web develop is listed under opt-in features in the compatibility table.
- A `test-web` job runs `patrol_develop_test.dart --web` on 3.47.x stable; the rest of CI stays on 3.38.x.

Red until the packages are published, the same way the 4.10.0 / 4.8.0 / 0.2.1 release went: the
semver check, because this bumps versions, and the `patrol_mcp` smoke and pana jobs, which
resolve from pub.dev on purpose and so can't see `patrol_cli` 4.9.0-dev.1 yet.

To see the DevTools extension locally you need to run `./publish_to_patrol_extension` first, since the built assets aren't committed.


zoskar (commented):
🐐 
You're the first victim of MCP pipelines, LMK if you have any feedback

zoskar:
Changelogs 🙏

packages/patrol_cli/lib/src/crossplatform/flutter_tool.dart
```diff
@@ -337,8 +337,16 @@ class FlutterTool {
   }
 
   void revertInteractiveMode(StdinModes stdinModes) {
-    io.stdin.echoMode = stdinModes.echoMode;
-    io.stdin.lineMode = stdinModes.lineMode;
+    // stdin may already be closed (e.g. after Ctrl+C), in which case setting
+    // echo/line mode throws a StdinException with "Bad file descriptor".
+    // That's harmless — the terminal is going away anyway — so swallow it.
```
pdenert: Do we need so long comments?

packages/patrol_cli/lib/src/web/web_test_backend.dart
pdenert, line 1: Do we need all this comments in code?

Issue #2874, opened 2026-01-09
Battery optimization notification allowance


danielgomezrico:
### Use case

Some apps requires special battery permission to be accepted by the user, it shows a dialog like this:

<img width="1076" height="372" alt="Image" src="https://github.com/user-attachments/assets/61b5b8a1-0675-4a8d-a9b0-eb2ed9e668ff" />

The current methods to allow permissions do not work with this one

The android permission is: `android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS`

### Proposal

The current ones supports this kind of permission dialog too

fylyppo:
Hi, thanks for the proposal. Currently, we can't put this into our roadmap, but if you want, you can raise a PR to cover this.
If you're interested, we can guide you on how an implementation should look. Wdyt?

danielgomezrico:
Sure! I'll be glad to try it

PR #3052, opened 2026-05-07
Feature/battery optimization allowance

dworik:
Issue #2874

Implements allowPermission as a new endpoint for android only.
Adds new resourceId for denyPermission for android.


pdenert (approved):
LGTM

Issue #2202, opened 2024-05-21
[Android] Keyboard does NOT close by any action after beeing opened by autofocus on inputfield widget.


DimitarStoyanov91:
### Steps to reproduce

1. Create emulator Pixel XL API 33 in Android Studio
2. Clone 'git@github.com:Prime-Holding/widget_toolkit.git'
3. Open in Android Studio widget_toolkit->packages->widget_toolkit->example
4. Checkout branch 'fix-patrol-debug'
5. Check test 'edit_fields_test.dart'
6. Run command `patrol test --target integration_test/tests/edit_fields_test.dart`

### Actual results

We have a widget with autofocus on the input field, which causes the keyboard to appear.
The normal flow would be that the keyboard hides after clicking a "Save" button.
Unfortunately when the test clicks "Save" the keyboard stays on the screen and hides other
widgets underneath it. We tried $.native.tap() on some elements on the screen, but that did
not hide the keyboard.

Update:
Tried also this approach:
Future<void> nativeEnterText(var widget, String inputText) async { await $(widget).tap(): await checkWidgetVisibility(widget); await $.native.enterText(Selector(focused: true), text: inputText, keyboardBehavior: KeyboardBehavior.showAndDismiss); }
This works fine for positive cases, but fails to enter the text when there are multiple invalid inputs
for that field and a final step with valid input that is accepted.

### Logs

There are no log errors.

### Patrol version

Patrol version: 3.6.1
patrol_cli version: 2.7.0
Device: Android 13.0 Pixel XL API 33

### Patrol Doctor output

Patrol doctor:
Patrol CLI version: 2.7.0
Flutter command: flutter 
  Flutter 3.16.3 • channel stable
Android: 
• Program adb found in /opt/homebrew/bin/adb
• Env var $ANDROID_HOME set to /Users/dimitar.stoyanov/Library/Android/sdk
iOS / macOS: 
• Program xcodebuild found in /usr/bin/xcodebuild
• Program ideviceinstaller found in /opt/homebrew/bin/ideviceinstaller

### Flutter Doctor output

[✓] Flutter (Channel stable, 3.16.3, on macOS 14.1.2 23B92 darwin-arm64, locale en-GB)
[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2023.1)
[✓] IntelliJ IDEA Community Edition (version 2022.3.1)
[✓] Connected device (2 available)
[...]

PR #3059, opened 2026-05-11
Fix/enter text issues

dworik:
## Summary

Fixes `enterText()` handling for already-focused text fields, especially fields focused via `autofocus`.
Previously, `PatrolTester.enterText()` registered `TestTextInput` before checking whether the target `EditableText` already had a real input connection. On live tests this could desynchronize Flutter focus, `TestTextInput`, and the native keyboard, causing the keyboard to stay open or native keyboard actions to stop working.
Changes:
### #2202
- Detect the target `EditableText` before registering `TestTextInput`.
- If the field is already focused, we intercept the connection to the system keyboard.

### https://github.com/leancodepl/patrol/pull/2111#issuecomment-2079825193
1) Add `hideKeyboard` to `enterText()` so live tests can keep/restorе the real keyboard connection when they need follow-up native keyboard actions.
2) Register and clean up `TestTextInput` around `WidgetTester.enterText()`.

## CI
- `patrol_finders check semver` expected to fail


packages/patrol_finders/lib/src/custom_finders/patrol_finder.dart
```diff
@@ -323,6 +323,7 @@ class PatrolFinder implements MatchFinder {
     Duration? visibleTimeout,
     Duration? settleTimeout,
     Alignment alignment = Alignment.center,
+    bool hideKeyboard = true,
```
piotruela, line 326: I think this change should be mentioned in the changelog

packages/patrol_finders/lib/src/custom_finders/patrol_tester.dart
```diff
@@ -431,33 +424,68 @@ class PatrolTester {
...
+          );
+          var editableTextState = tester.state<EditableTextState>(
+            editableTextFinder,
+          );
+          final wasFocused = editableTextState.widget.focusNode.hasFocus;
+          final usesLiveBinding =
+              tester.binding is LiveTestWidgetsFlutterBinding;
+
+          if (!kIsWeb && wasFocused) {
+            editableTextState.widget.focusNode.unfocus();
```
piotruela: Can this trigger `unfocus` action on the field? What if someone has validation triggered on unfocus? Will it be invoked?

dev/e2e_app/pubspec.yaml
```diff
@@ -29,6 +29,8 @@ dependencies:
 
 # Due to: https://github.com/Baseflow/flutter-permission-handler/issues/1450#issuecomment-2673836556
```
piotruela, line 30: This comment should be placed directly above line 34

packages/patrol_finders/lib/src/custom_finders/patrol_tester.dart
dworik: In the current version it does. I don't know if we can restore connection to the system keyboard without losing the focus first (this is only necessary with autofocus). I will try some other approaches but if nothing works I think its an acceptable side effect.

dworik: The new version seems to work without losing focus

piotruela (approved):

pdenert (approved):

Issue #1938, opened 2023-11-21
Wait for `PatrolFinder` to be visible while using `at()`


fylyppo:
### Steps to reproduce

1. Clone [bug/at_reproduce_example](https://github.com/leancodepl/patrol/tree/bug/at_reproduce_example)
2. Go to `packages/patrol/example`
3. Run `patrol test -t integration_test/example_test.dart`

### Actual results

Test is failed right after `await $(ListView).$(ListTile).at(1).tap();` is invoked.

### Logs

<details>
<summary>Logs</summary>

```
java.lang.AssertionError: Dart test failed: example_test aye
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞═════════════════
The following IndexError was thrown running a test:
RangeError (index): Index out of range: index should be less than
1: 1

When the exception was thrown, this was the stack:
#0      Iterable.elementAt (dart:core/iterable.dart:807:5)
#1      _IndexFinderMixin.filter (package:flutter_test/src/finders.dart:1103:28)
#3      _HitTestableWidgetFinder.filter (package:flutter_test/src/finders.dart:1143:37)
#5      Iterable.isEmpty (dart:core/iterable.dart:540:33)
#6      PatrolTester.waitUntilVisible.<anonymous closure> (package:patrol_finders/src/custom_finders/patrol_tester.dart:401:43)
#9      TestAsyncUtils.guard (package:flutter_test/src/test_async_utils.dart:71:41)
#10     PatrolTester.waitUntilVisible (package:patrol_finders/src/custom_finders/patrol_tester.dart:397:27)
#11     PatrolTester.tap.<anonymous closure> (package:patrol_finders/src/custom_finders/patrol_tester.dart:249:36)
#14     TestAsyncUtils.guard (package:flutter_test/src/test_async_utils.dart:71:41)
#15     PatrolTester.tap (package:patrol_finders/src/custom_finders/patrol_tester.dart:248:27)
#16     PatrolFinder.tap (package:patrol_finders/src/custom_finders/patrol_finder.dart:207:18)
#17     main.<anonymous closure> (file:///Users/fylyppo/Documents/GitHub/patrol/packages/patrol/example/integration_test/example_test.dart:12:43)
<asynchronous suspension>
#18     patrolTest.<anonymous closure> (package:patrol/src/common.dart:110:7)
<asynchronous suspension>
[...]

PR #3062, opened 2026-05-12
Fix/wait for at

dworik:
Fixes issue #1938
## Problem
`first`, `last`, and `at()` delegated directly to Flutter finders, which throw immediately for empty results (`StateError`/`RangeError`). That prevented Patrol’s normal waiting behavior in `tap()`/`waitUntilVisible()` from kicking in when a widget appeared later.
## Fix
Replaced those selectors with a Patrol wrapper that returns an empty result instead of throwing when the selected element is not available yet. This lets waitUntilVisible() keep retrying until the widget appears or the regular timeout is reached.
## Test plan
- [x] dart analyze in `packages/patrol_finders` clean
- [x] flutter test in `packages/patrol_finders/test` - 127/127 passed, including 7 new tests covering: 
   - negative numbers
   - widgets appearing after delay
   - non-existing widgets
- [x] patrol test in `dev/e2e_app/patrol_test/at_finder_test.dart` - successfully waits for a second widget to appear without throwing.

https://github.com/user-attachments/assets/c8b766e7-5968-45d7-84c0-0535f9b08134

## CI
- `test android emulator webview...` expected to fail

### DIFF
--- a/dev/e2e_app/patrol_test/at_finder_test.dart
+++ b/dev/e2e_app/patrol_test/at_finder_test.dart
@@ -1,4 +1,4 @@
-import 'package:flutter/material.dart';
+import 'package:e2e_app/keys.dart';
 
 import 'common.dart';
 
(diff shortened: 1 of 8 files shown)


packages/patrol_finders/lib/src/custom_finders/patrol_finder.dart
```diff
@@ -554,7 +554,10 @@ class PatrolFinder implements MatchFinder {
   @override
   PatrolFinder at(int index) {
     // TODO: Throw a better error (https://github.com/leancodepl/patrol/issues/548)
```
pdenert: You can remove this comment. The issue has been closed

packages/patrol_finders/lib/src/custom_finders/patrol_finder.dart
pdenert, line 1: Can we apply same logic for first and last?

packages/patrol_finders/test/patrol_finder_test.dart
pdenert, line 1: Please add test with negative index, to check if that behave correctly

pdenert (commented):
Please also add video of the demo run of this tests

dev/e2e_app/patrol_test/at_finder_test.dart
```diff
@@ -0,0 +1,121 @@
...
+      key: _AtFinderItemKey(index),
+      title: Text('Item $index'),
+      onTap: onTap,
+    );
+  }
+}
+
+class _AtFinderItemKey extends ValueKey<int> {
+  const _AtFinderItemKey(super.value);
+}
```
pdenert: This shouldn't be added in test, but directly in the e2e_app. Or you can find and reuse some screen that already exist

dev/e2e_app/pubspec.yaml
```diff
@@ -29,6 +29,8 @@ dependencies:
 
 # Due to: https://github.com/Baseflow/flutter-permission-handler/issues/1450#issuecomment-2673836556
 dependency_overrides:
+  patrol_finders:
+    path: ../../packages/patrol_finders
```
pdenert: Revert that. We want to test there real packages

pdenert (approved):

PR #3073, opened 2026-05-17
fix: rename vendored HTTPParserC enums to avoid macOS http.h conflicts

### LINKED ISSUE #1976: Telegraph is not working with macOS
[Telegraph](https://github.com/Building42/Telegraph) depends on a [HTTPParserC](https://github.com/Building42/HTTPParserC) the current version of which does not work on macOS. For now, we have applied a workaround by copying Telegraph's code and fixed HTTPParserC code into our library. We should aim to merge [this fix](https://github.com/Building42/HTTPParserC/pull/4). So that we can go back to importing Telegraph as the normal dependency.

brucearctor:
## Problem

The vendored `http_parser.h` (inside `darwin/Classes/Telegraph/HTTPParserC/`) defines enum types (`http_method`, `http_status`, `http_errno`) that conflict with macOS system header `<http.h>`, causing compilation failures on macOS targets.

A previous fix already prefixed the **enum values** (`HTTP_GET` → `PARSER_HTTP_GET`), but the **enum type names** and **function signatures** were not updated.

## Solution

Rename the enum types to avoid namespace collisions:
- `enum http_method` → `enum http_parser_method`
- `enum http_status` → `enum http_parser_status`
- `enum http_errno` → `enum http_parser_errno`

Also updates the Swift typealias and usage in `HTTPParser+Raw.swift` to reference the renamed types.

This matches the approach from the now-closed [HTTPParserC PR #4](https://github.com/Building42/HTTPParserC/pull/4) (upstream has since moved to `llhttp`).

## Files Changed (3 files, 15 line swaps)
- `http_parser.h` — renamed enum type declarations and function signatures
- `http_parser.c` — renamed enum type references in function implementations
- `HTTPParser+Raw.swift` — updated typealias and constructor call

Fixes #1976

### DIFF
--- a/packages/patrol/darwin/Classes/Telegraph/HTTPParserC/http_parser.c
+++ b/packages/patrol/darwin/Classes/Telegraph/HTTPParserC/http_parser.c
@@ -937,7 +937,7 @@ reexecute:
           goto error;
         }
 
-        parser->method = (enum http_method) 0;
+        parser->method = (enum http_parser_method) 0;
         parser->index = 1;
         switch (ch) {
           case 'A': parser->method = PARSER_HTTP_ACL; break;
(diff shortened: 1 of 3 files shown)


brucearctor:
looks like failures are due to permissions?

Issue #3080, opened 2026-05-21
When using $.platform.tap() appId defaults to null on Android but defaults to app being tested on iPhone


MarkOSullivan94:
### Steps to reproduce

1. Clone https://github.com/MarkOSullivan94/patrol_example
2. Open Android emulator 
3. Run the integration test `patrol test -t patrol_test/app_test.dart`
4. Open iPhone simulator
5. Run the integration test `patrol test -t patrol_test/app_test.dart`

### Actual results

The Android integration test will pass and the iPhone integration test will fail because for the tap to work on the website within the iPhone you need to add `appId: com.apple.mobilesafari` to the tap.

The behaviour is not consistent on both platforms which feels like a bug.

### Logs

<details>
<summary>Logs</summary>

```
➜  patrol_example git:(main) ✗ patrol test -t patrol_test/app_test.dart --verbose                 
Verbose mode enabled. More logs will be printed.

Update available! 4.3.1 → 4.4.0

Check the compatibility table at: https://patrol.leancode.co/documentation/compatibility-table


Received 1 test target(s)
Received test target: /Users/mos/IdeaProjects/patrol_example/patrol_test/app_test.dart
$ flutter --no-version-check --suppress-analytics devices --machine
Multiple devices found. Please choose one:

  [1] iPhone (00008030-0008253C349B802E)
  [2] iPhone 17 Pro Max (0AB540DE-939E-4CA9-AFC6-AD1DDAAB87C1)
  [3] macOS (macos)
  [4] Chrome (chrome)
  [q] Cancel and exit

Please select an option (1-4, q to quit): 2
Selected device: iPhone 17 Pro Max (0AB540DE-939E-4CA9-AFC6-AD1DDAAB87C1)
Received 1 device(s) to run on
Received device: iPhone 17 Pro Max (0AB540DE-939E-4CA9-AFC6-AD1DDAAB87C1)
Generated entrypoint /Users/mos/IdeaProjects/patrol_example/patrol_test/test_bundle.dart with 1 bundled test(s)
$ flutter --suppress-analytics --no-version-check pub deps --style=list
Received 10 --dart-define(s) (0 custom, 10 internal)
Received internal --dart-define: PATROL_APP_PACKAGE_NAME=com.example.patrol_example
Received internal --dart-define: PATROL_APP_BUNDLE_ID=com.example.patrolExample
Received internal --dart-define: PATROL_ANDROID_APP_NAME=Patrol Example
[...]

PR #3088, opened 2026-05-27
Document that appId is ignored on Android in PlatformAutomator

zoskar:
Methods on PlatformAutomator and MobileAutomator that take an optional appId silently drop it on Android (only the iOS branch passes it through). Note this in the dartdoc of each affected method

Refs #3080.

### DIFF
--- a/packages/patrol/lib/src/platform/platform_automator.dart
+++ b/packages/patrol/lib/src/platform/platform_automator.dart
@@ -186,7 +186,9 @@ class PlatformAutomator {
   /// It waits for the view to become visible for [timeout] duration.
   /// If the native view is not found, an exception is thrown.
   ///
-  /// On Android, [appId] is ignored.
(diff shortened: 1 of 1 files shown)


packages/patrol/lib/src/platform/platform_automator.dart
```diff
@@ -185,6 +185,8 @@ class PlatformAutomator {
   ///
   /// It waits for the view to become visible for [timeout] duration.
   /// If the native view is not found, an exception is thrown.
+  ///
+  /// On Android, [appId] is ignored.
```
pdenert: Here we have also web call, maybe we should describe that it's used in iOS (and why)

zoskar: Fixed

pdenert (approved):

Issue #1428, opened 2023-06-29
Generated test bundle is broken on Patrol 2.0 on Windows


GentiShtjefni:
### Steps to reproduce

Create a simple Flutter project and setup patrol native automation.
main_test.dart: 

```dart
import 'package:flutter/material.dart';
import 'package:flutter_test/flutter_test.dart';
import 'package:patrol/patrol.dart';

void main() {
  patrolTest(
    'counter state is the same after going to home and switching apps',
    nativeAutomation: true,
        ($) async {
      // Replace later with your app's main widget
      await $.pumpWidgetAndSettle(
        MaterialApp(
          home: Scaffold(
            appBar: AppBar(title: const Text('app')),
            backgroundColor: Colors.blue,
          ),
        ),
      );

      expect($('app'), findsOneWidget);
      await $.native.pressHome();
    },
  );
}
```

MainActivityTest.java:

```java
package com.ecample.test; // replace "pl.leancode.patrol.example" with your app's package

import androidx.test.platform.app.InstrumentationRegistry;
import org.junit.Test;
import org.junit.runner.RunWith;
import org.junit.runners.Parameterized;
import org.junit.runners.Parameterized.Parameters;
import pl.leancode.patrol.PatrolJUnitRunner;

@RunWith(Parameterized.class)
public class MainActivityTest {
    @Parameters(name = "{0}")
    public static Object[] testCases() {
        PatrolJUnitRunner instrumentation = (PatrolJUnitRunner) InstrumentationRegistry.getInstrumentation();
        instrumentation.setUp(MainActivity.class);
        instrumentation.waitForPatrolAppService();
        return instrumentation.listDartTests();
    }

    public MainActivityTest(String dartTestName) {
        this.dartTestName = dartTestName;
    }

    private final String dartTestName;

    @Test
    public void runDartTest() {
        PatrolJUnitRunner instrumentation = (PatrolJUnitRunner) InstrumentationRegistry.getInstrumentation();
        instrumentation.runDartTest(dartTestName);
    }
}
```

build.gradle: 

```gradle
def localProperties = new Properties()
def localPropertiesFile = rootProject.file('local.properties')
if (localPropertiesFile.exists()) {
    localPropertiesFile.withReader('UTF-8') { reader ->
        localProperties.load(reader)
    }
}

def flutterRoot = localProperties.getProperty('flutter.sdk')
if (flutterRoot == null) {
[...]

PR #3094, opened 2026-06-04
Fix/patrol cli windows hang bundle

### LINKED ISSUE #2565: Can't launch test, stuck on gradlew dependencies
### Steps to reproduce

Using Windows env (not WSL)

- Follow the getting started for Android only
- Execute the following command : `patrol test -t integration_test/example_test.dart -v`
- Stuck at line `.\gradlew.bat :app:dependencies`

### Actual results

It stuck forever... I tried to go into android and execute manually the command `.\gradlew.bat :app:dependencies` and it build perfectly.

### Logs

<details>
<summary>Logs</summary>

```
C:\Users\dimit\Documents\Dev\MyApp>patrol test -t integration_test/example_test.dart -v
Verbose mode enabled. More logs will be printed.
Received 1 test target(s)
Received test target: C:\Users\dimit\Documents\Dev\MyApp\integration_test/example_test.dart
Generated entrypoint C:\Users\dimit\Documents\Dev\MyApp\integration_test\test_bundle.dart with 1 bundled test(s)
$ flutter --no-version-check --suppress-analytics devices --machine
No device specified, using the first one (ce051715db73da0601)
Received 1 device(s) to run on
Received device: ce051715db73da0601
$ flutter doctor --verbose
$ flutter --suppress-analytics --no-version-check pub deps --style=list
Received 10 --dart-define(s) (0 custom, 10 internal)
Received internal --dart-define: PATROL_WAIT=0
Received internal --dart-define: PATROL_APP_PACKAGE_NAME=com.example.myapp
Received internal --dart-define: PATROL_APP_BUNDLE_ID=com.example.myapp
Received internal --dart-define: PATROL_ANDROID_APP_NAME=MyApp
Received internal --dart-define: PATROL_IOS_APP_NAME=MyApp
Received internal --dart-define: INTEGRATION_TEST_SHOULD_REPORT_RESULTS_TO_NATIVE=false
Received internal --dart-define: PATROL_TEST_LABEL_ENABLED=true
Received internal --dart-define: PATROL_TEST_SERVER_PORT=8081
Received internal --dart-define: PATROL_APP_SERVER_PORT=8082
Received internal --dart-define: COVERAGE_ENABLED=false
$ flutter build apk --config-only -t integration_test/test_bundle.dart
$ flutter doctor --verbose
$ .\gradlew.bat :app:dependencies
```

</details>

### Patrol version

patrol: ^3.14.1

### Patrol Doctor output

<details>
<summary>Patrol Doctor output</summary>

```
Patrol doctor:
Patrol CLI version: 3.5.1
Flutter command: flutter 
  Flutter 3.29.2 • channel stable
Android: 
• Program adb found in C:\Users\dimit\AppData\Local\Android\Sdk\platform-tools\adb.exe
• Env var $ANDROID_HOME set to C:\Users\dimit\AppData\Local\Android\Sdk
```

</details>

### Flutter Doctor output

<details>
<summary>Flutter Doctor output</summary>

```
Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel stable, 3.29.2, on Microsoft Windows [version 10.0.26100.3476], locale fr-FR)
[√] Windows Version (Windows 11 or higher, 24H2, 2009)
[√] Android toolchain - develop for Android devices (Android SDK version 35.0.0)
[X] Chrome - develop for the web (Cannot find Chrome executable at .\Google\Chrome\Application\chrome.exe)
    ! Cannot find Chrome. Try setting CHROME_EXECUTABLE to a Chrome executable.
[X] Visual Studio - develop Windows apps
    X Visual Studio not installed; this is necessary to develop Windows apps.
      Download at https://visualstudio.microsoft.com/downloads/.
      Please install the "Desktop development with C++" workload, including all of its default components
[√] Android Studio (version 2024.3)
[√] VS Code (version 1.98.2)
[√] Connected device (3 available)
[√] Network resources

! Doctor found issues in 2 categories.
```

</details>

zoskar:
## Problem

`patrol test`/`build`/`develop` hang on Windows at `gradlew :app:dependencies`.
`detectOrchestratorVersion()` only drains the process's stdout, so once Gradle
fills the stderr pipe buffer the child blocks and `exitCode` never completes.
`buildApkConfigOnly()` already drains both streams (#2905); this method was missed.

## Fix

Drain stderr too. No change on macOS/Linux.

#2565 was closed after an unrelated NDK-warning fix that just cut stderr below
the buffer threshold and masked the deadlock — the root cause still reproduces
on `master`.

---

This PR also dropped its original `test_bundler.dart` change for #1428: `master`
now resolves those paths with `path.relative` (#3104), so it's obsolete. Kept a
regression test for the forward-slash / `.\` cases instead.

Fixes #2565.

### DIFF
--- a/packages/patrol_cli/lib/src/android/android_test_backend.dart
+++ b/packages/patrol_cli/lib/src/android/android_test_backend.dart
@@ -230,6 +230,11 @@ class AndroidTestBackend {
             }
           })
           .disposedBy(scope);
+      // `:app:dependencies` writes a lot to stderr (Gradle warnings, daemon
+      // and progress output). The stderr stream must be drained or the child
+      // process blocks once the OS pipe buffer fills, hanging the command
+      // forever on Windows. Same failure mode fixed in buildApkConfigOnly.
+      process.listenStdErr((l) => _logger.detail('\t$l')).disposedBy(scope);
 
       await process.exitCode;
     });
--- a/packages/patrol_cli/lib/src/android/android_test_backend.dart
+++ b/packages/patrol_cli/lib/src/android/android_test_backend.dart
@@ -230,10 +230,7 @@ class AndroidTestBackend {
             }
           })
           .disposedBy(scope);
-      // `:app:dependencies` writes a lot to stderr (Gradle warnings, daemon
-      // and progress output). The stderr stream must be drained or the child
-      // process blocks once the OS pipe buffer fills, hanging the command
-      // forever on Windows. Same failure mode fixed in buildApkConfigOnly.
+      // Drain stderr, or the process hangs on Windows when the pipe fills.
       process.listenStdErr((l) => _logger.detail('\t$l')).disposedBy(scope);
 
       await process.exitCode;
--- a/packages/patrol_cli/test/test_bundler_test.dart
+++ b/packages/patrol_cli/test/test_bundler_test.dart
@@ -114,12 +114,9 @@ import 'example/example_test.dart' as example__example_test;''');
     });
 
     test('generates clean imports from non-native path forms (#1428)', () {
-      // The target is not always passed in the host-native form: `patrol
-      // develop`/MCP hand over forward-slash absolute paths, and PowerShell
-      // tab-completion injects a leading `.\` segment. Both must still resolve
-      // to a valid, relative import rather than leaking a raw absolute path
-      // into the generated bundle. Regression test for:
-      // https://github.com/leancodepl/patrol/issues/1428
+      // Non-native target forms must still resolve to a relative import:
+      // forward-slash absolute paths (`patrol develop`/MCP) and `.\`-prefixed
+      // paths (PowerShell tab-completion).
       final absoluteNative = fs.path.join(
         platform.home,
         'awesome_app',
--- a/packages/patrol_cli/test/test_bundler_test.dart
+++ b/packages/patrol_cli/test/test_bundler_test.dart
@@ -113,6 +113,37 @@ import 'example_test.dart' as example_test;
 import 'example/example_test.dart' as example__example_test;''');
     });
 
+    test('generates clean imports from non-native path forms (#1428)', () {
+      // The target is not always passed in the host-native form: `patrol
+      // develop`/MCP hand over forward-slash absolute paths, and PowerShell
+      // tab-completion injects a leading `.\` segment. Both must still resolve
+      // to a valid, relative import rather than leaking a raw absolute path
+      // into the generated bundle. Regression test for:
+      // https://github.com/leancodepl/patrol/issues/1428
+      final absoluteNative = fs.path.join(
+        platform.home,
+        'awesome_app',
+        'patrol_test',
+        'example_test.dart',
+      );
+
+      // Forward-slash absolute path (e.g. `c:/Users/.../example_test.dart`).
+      final forwardSlashAbsolute = absoluteNative.replaceAll(r'\', '/');
+      expect(
+        testBundler.generateImports(testDirectory, [forwardSlashAbsolute]),
+        "import 'example_test.dart' as example_test;",
+      );
+
+      // `.\`-prefixed relative path.
+      final dotSlashRelative =
+          '.${fs.path.separator}'
+          '${fs.path.join('patrol_test', 'example_test.dart')}';
+      expect(
+        testBundler.generateImports(testDirectory, [dotSlashRelative]),
+        "import 'example_test.dart' as example_test;",
+      );
+    });
+
     test(
       'generates relative import when target is outside the test directory',
       () {
--- a/packages/patrol_cli/CHANGELOG.md
+++ b/packages/patrol_cli/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Don't listen for `SIGTERM` on Windows, where it is not supported and throws an unhandled `SignalException`. (#3035)
 - Fix `--exclude` not working. (#2990)
 - Fix `test_bundle.dart` generating a broken absolute import (and an invalid import alias containing characters such as `-`) when the test target lives outside the configured `test_directory`. The import is now computed relative to the bundle and the alias is sanitized. (#3104)
+- Fix `patrol test`/`patrol develop` hanging on Windows at `gradlew :app:dependencies` by also draining the gradle process stderr stream during orchestrator-version detection. (#2565)
 
 ## 4.4.0
 


JanGarstecki:
# PR #3094 (`fix/patrol-cli-windows-hang-bundle`) — Local Verification Report

**Date:** 2026-07-07
**Machine/OS:** Windows 11 Pro (build 26200)
**Device:** `emulator-5554` (sdk gphone64 x86_64, Android emulator)
**Toolchain:** Flutter 3.41.2 / Dart 3.11.0, Gradle 8.12, Java 21 (JBR bundled with Android Studio)
**Test project:** `dev/e2e_app` in the `patrol` repo — a real, multi-plugin Flutter app (Google Maps, image_picker, geolocator, permission_handler, webview_flutter, etc.), already built before (warm Gradle caches).
**Repo state:** `patrol_cli` was globally activated on this machine via `dart pub global activate --source path` pointing directly at `packages/patrol_cli` inside this repo. This means `patrol <command>` on the command line always executes whatever code is currently checked out on the local `patrol` git repo — switching branches with `git checkout` and re-running `patrol` exercises that branch's code directly, with no reinstall step.

Two independent bugs are bundled in this PR. Each was tested separately.

---

## Bug 1 — Gradle `:app:dependencies` hang on Windows (issue #2565)

### What the PR claims
`AndroidTestBackend.detectOrchestratorVersion()` starts `gradlew :app:dependencies` and only drains the child process's **stdout**. If Gradle writes enough to **stderr** to fill the OS pipe buffer before the process exits, the child blocks on the write and `await process.exitCode` never returns — a classic Windows anonymous-pipe deadlock. The sibling method `buildApkConfigOnly()` already drains both streams and does not have this problem.

### What we verified by reading the code
On `master` (commit `76aaeb4b9`), in `packages/patrol_cli/lib/src/android/android_test_backend.dart`:
- `buildApkConfigOnly()` (existing code, unmodified by this PR): calls `process.listenStdOut(...)` **and** `process.listenStdErr(...)` before `await process.exitCode`.
- `detectOrchestratorVersion()` (the function this PR patches): calls `process.listenStdOut(...)` only, then `await process.exitCode`, with no stderr listener at all.

This confirms the code asymmetry described in the PR is real and still present on current `master` — no other patch has fixed it in the meantime.

### Test A — full build, branch A/B comparison on the real emulator
Command run from `dev/e2e_app`, identical on both branches:
```
dart run ../../packages/patrol_cli/bin/main.dart build android -t patrol_test/example_test.dart --no-check-compatibility
```
This command exercises the exact affected code path (`build()` calls `buildApkConfigOnly()` → `loadJavaPathFromFlutterDoctor()` → `detectOrchestratorVersion()` → `:app:assembleDebug`).

| Branch | Result | Total wall time | `:app:assembleDebug` step time |
|---|---|---|---|
| `pr-3094-test` (PR head, fix present) | succeeded | 3m03.7s | 90.0s |
| `master` (no fix) | succeeded | 8m04.9s | 304.2s |

Neither run produced a full hang. The pre-build phase (which includes the buggy `gradlew :app:dependencies` call) was roughly 2x slower on `master`, but it completed. We measured the raw output of `:app:dependencies` directly (`./gradlew.bat :app:dependencies --configuration debugRuntimeClasspath > stdout.log 2> stderr.log` after stopping the Gradle daemon with `gradlew --stop` to force a cold run): it produced **43,606 bytes of stdout and 0 bytes of stderr** in this specific project/toolchain state right now. With zero stderr, the described deadlock condition (an unread stderr pipe filling up) cannot be triggered — this matches the PR's own explanation for why the original bug reporters saw #2565 "go away" after fixing an unrelated NDK warning: it didn't fix the bug, it just reduced stderr volume below the pipe-buffer threshold, making the deadlock non-reproducible in that environment. The same appears to be true for this specific project on this machine today.

**Conclusion of Test A:** could not force a real Gradle-triggered hang in this specific project/toolchain combination, because it currently emits ~0 bytes of stderr during `:app:dependencies`. This does not disprove the bug — it confirms the bug is real but environment/output-volume-dependent, exactly as the PR describes.

### Test B — isolated minimal reproduction of the actual deadlock mechanism
Since Test A's result depends on how much stderr Gradle happens to produce (which the PR itself says is unreliable), we wrote a standalone Dart script (`repro_stderr_deadlock.dart`) that reproduces the exact code pattern from `detectOrchestratorVersion()`, independent of Gradle:
- Spawns a PowerShell child process that writes ~4,000,000 bytes to stderr (`[Console]::Error.WriteLine` in a loop) and a small amount to stdout.
- **Buggy variant** (`drainStderr=false`): listens to stdout only, then `await process.exitCode`. Matches current `detectOrchestratorVersion()`.
- **Fixed variant** (`drainStderr=true`): listens to both stdout and stderr, then `await process.exitCode`. Matches the PR's patch.

Results (each run capped at a 15-second timeout):
```
drainStderr=false → TIMEOUT after 0:00:15.019636: process.exitCode never completed -> DEADLOCKED
drainStderr=true  → process.exitCode completed after 0:00:00.459808 with exitCode=0 -> no deadlock
```

**Conclusion of Test B:** the deadlock mechanism described in the PR is 100% real and reproducible on this Windows machine. The buggy pattern (stdout drained, stderr not) hangs indefinitely; the fixed pattern (both drained) completes in well under half a second. This isolates and confirms the root cause independent of Gradle's specific output volume on any given day.

### Bug 1 verdict
**Real bug, correctly diagnosed, fix is correct and minimal.** The patch (`process.listenStdErr((l) => _logger.detail('\t$l')).disposedBy(scope);` added to `detectOrchestratorVersion()`) applies cleanly to current `master` with no conflicts (confirmed via `git merge-tree`).

---

## Bug 2 — garbled `test_bundle.dart` imports for forward-slash absolute paths (issue #1428)

### What the PR claims
`TestBundler._normalizeTestPath()` strips a literal backslash-prefixed project-root path from the test file path. When the caller passes a forward-slash absolute path instead (e.g. from `patrol develop`/MCP, or PowerShell tab-completion inserting a `.\` segment), the literal string strip fails to match, and the raw path leaks into the generated import statement (e.g. `import 'c:/…/foo_test.dart' as c:__…;`), which is invalid Dart and breaks the build.

### What we found by inspecting git history
The PR branch's merge-base with `master` is commit `f4f2526da` (2026-06-16). Since that point, two unrelated commits landed on `master` that rewrote the exact function this PR patches:
- `75120cf2d` — "fix(patrol_cli): generate relative, valid imports in test_bundle.dart for targets outside test_directory" (2026-06-15, fixes issue #3104)
- `9ceff8c4e` — "fix(patrol_cli): resolve relative test paths against project root in test bundler" (2026-06-16)

We confirmed neither commit is an ancestor of the PR branch (`git merge-base --is-ancestor <commit> pr-3094-test` returned false for both). On current `master`, `TestBundler._normalizeTestPath()` no longer exists — it has been replaced by `_toAbsoluteTestPath()` + `_createImportPath()`/`_createTestName()`, which resolve paths using `package:path`'s `path.relative()` instead of literal string manipulation.

### Test A — merge conflict check
We ran:
```
git merge-tree $(git merge-base master pr-3094-test) master pr-3094-test
```
Result: **`packages/patrol_cli/lib/src/test_bundler.dart` has a real, unresolved conflict** — the PR's edit to `_normalizeTestPath()` collides with master's replacement of that function. (There is also a trivial conflict in `CHANGELOG.md`.) The `android_test_backend.dart` change (Bug 1's fix) merges cleanly with no conflict.

### Test B — regression test run on both branches
We added a temporary unit test to `packages/patrol_cli/test/test_bundler_test.dart` reproducing the exact scenario described in the PR: a forward-slash absolute path on a simulated Windows filesystem (`c:/Users/charlie/awesome_app/patrol_test/example_test.dart`), asserting that `generateImports()` produces a clean relative import instead of a garbled one.

```
dart test test/test_bundler_test.dart --plain-name "regression #1428"
```

| Branch | Result |
|---|---|
| `pr-3094-test` (PR head, fix present) | PASS |
| `master` (current, PR not applied) | **PASS** |

The test passes on current `master` even without this PR's patch. Master's newer `path.relative()`-based implementation (from commits `75120cf2d`/`9ceff8c4e`) already normalizes forward-slash absolute paths and `.\`-prefixed paths correctly as a side effect — this was not the primary goal of those commits (they were fixing a different symptom, #3101/#3104), but it happens to also resolve the symptom described in #1428.

### Bug 2 verdict
**The underlying symptom (garbled/invalid imports from forward-slash absolute paths) no longer reproduces on current `master`.** It appears to have already been fixed as a side effect of unrelated work (`75120cf2d`, `9ceff8c4e`) that landed after this PR's branch point. This part of PR #3094 is stale: it edits a function (`_normalizeTestPath`) that no longer exists on `master`, and its diff does not merge cleanly.

---

## Summary

| Item | Status | Evidence |
|---|---|---|
| Bug 1 — Gradle stderr deadlock (#2565) | **Confirmed real, fix confirmed correct** | Code inspection (asymmetry vs. `buildApkConfigOnly`) + isolated minimal repro (deadlock reproduced and resolved deterministically) + full A/B build comparison (master path measurably slower, though it happened not to fully hang this run because current stderr output is ~0 bytes) |
| Bug 2 — garbled test_bundle.dart imports (#1428) | **Symptom no longer reproduces on current `master`; PR is stale here** | `git merge-tree` shows a real conflict against the function this PR edits; a regression test reproducing the exact reported scenario passes on `master` without this PR's patch |

**What was actually fixed by this PR, as verified on this machine:** the Gradle `:app:dependencies` stderr-drain fix in `android_test_backend.dart` (Bug 1). This is a genuine, still-needed, correctly-implemented fix and merges cleanly onto current `master`.

**What needs follow-up:** PR #3094 needs to be rebased onto current `master`. During the rebase, the `test_bundler.dart` hunk (Bug 2) will conflict and, based on the regression test above, is very likely no longer necessary — it should be re-evaluated (and probably dropped, possibly keeping just a regression test for the forward-slash-path case) rather than reapplied as-is.

packages/patrol_cli/lib/src/android/android_test_backend.dart
```diff
@@ -230,6 +230,11 @@ class AndroidTestBackend {
             }
           })
           .disposedBy(scope);
+      // `:app:dependencies` writes a lot to stderr (Gradle warnings, daemon
+      // and progress output). The stderr stream must be drained or the child
+      // process blocks once the OS pipe buffer fills, hanging the command
+      // forever on Windows. Same failure mode fixed in buildApkConfigOnly.
```
pdenert: Does this comment need to be so long?

packages/patrol_cli/test/test_bundler_test.dart
```diff
@@ -113,6 +113,37 @@ import 'example_test.dart' as example_test;
 import 'example/example_test.dart' as example__example_test;''');
     });
 
+    test('generates clean imports from non-native path forms (#1428)', () {
+      // The target is not always passed in the host-native form: `patrol
+      // develop`/MCP hand over forward-slash absolute paths, and PowerShell
+      // tab-completion injects a leading `.\` segment. Both must still resolve
+      // to a valid, relative import rather than leaking a raw absolute path
+      // into the generated bundle. Regression test for:
+      // https://github.com/leancodepl/patrol/issues/1428
```
pdenert: Same here

zoskar: @pdenert fixed both

pdenert (approved):

PR #3109, opened 2026-06-22
fix: support BrowserStack (-only-testing) sharding on iOS

fylyppo:
## What
Make Patrol's iOS test runner work with xcodebuild `-only-testing`, so a
selected subset of tests can run in one invocation. This is what BrowserStack
XCUITest sharding uses, and it previously could not run Patrol tests.

## Why it didn't work
The runner adds one XCTest method per Dart test dynamically inside
`+testInvocations`, and the `-only-testing` path was incomplete:
- a fresh `PatrolServer` was created per selected test → port re-bind broke the
  app-service connection when a shard held >1 test;
- the path skipped the initial `listDartTests` handshake → the app's
  `PatrolAppService` wasn't ready → "connection refused";
- the `-only-testing` selector carries only the dotted group name, but the app
  matches `runDartTest` on the full `group + test` name.

## Changes (`PatrolIntegrationTestIosRunner.h`)
- accumulate selected tests (dedup, dotted-name guard) instead of overwriting;
- create the server / app-service client once and reuse them;
- always launch + `listDartTests` once (cached) to warm the service and get full names;
- map each selected group selector → full Dart test name(s);
- skip-placeholder for a selector with no match (e.g. platform-excluded) so the
  shard doesn't crash on a nil method signature.

## Testing
Validated on BrowserStack App Automate (iPhone 16 Pro Max, 2 shards): each shard
ran its assigned tests and passed; a platform-excluded test in a shard was
skipped gracefully. Non-sharded runs are unaffected (`selectedTests` stays empty).


fylyppo:
Solved by https://github.com/leancodepl/patrol/pull/3197

PR #3114, opened 2026-06-24
CI: fix iOS 26.2 simulator E2E job (Xcode 26.2 + GoogleMaps 10) + other workflows

Kendru98:
## What

Fixes the `test ios simulator` E2E job (`Flutter 3.38.x on iPhone 17 (26.2) simulator`), which was failing on `master` and on PRs.

## Root causes & fixes

1. **Toolchain/runtime skew → `Failed to terminate app:0` flake.** The workflow targets the iOS 26.2 runtime but never selected an Xcode, so it ran `xcodebuild`/XCUITest from the runner default (Xcode 16.4) against an iOS 26.2 simulator. That mismatch made `XCUIApplication.launch()`'s implicit app termination unreliable. → Added a **Select Xcode (26.2)** step so the toolchain matches the runtime.

2. **GoogleMaps link failure on Xcode 26.** Once aligned to Xcode 26.x, the build failed to link: GoogleMaps 8.4.0 / Google-Maps-iOS-Utils 5.x reference the `swiftCompatibility56` Swift shim libraries, which Xcode 26 removed. → Bumped `google_maps_flutter` to `^2.17.1` (resolves GoogleMaps **10.14.0** + Utils **6.1.3**, modern Swift, no shims) and raised the e2e app's iOS deployment target to **16.0** (required by GoogleMaps 10).

## Result

The iOS simulator job now builds and runs `pull_to_refresh_test` to a clean pass on Xcode 26.2 (`✅ Successful: 1`), with no terminate flake.

## Notes

This will not fix 100% of our flakiness, to do this we should use better macos runner like _l or _xl


dev/e2e_app/pubspec.lock
```diff
@@ -764,10 +764,10 @@ packages:
     dependency: "direct main"
     description:
       name: meta
-      sha256: "1741988757a65eb6b36abe716829688cf01910bbf91c34354ff7ec1c3de2b349"
+      sha256: "23f08335362185a5ea2ad3a4e597f1375e78bce8a040df5c600c8d3552ef2394"
       url: "https://pub.dev"
     source: hosted
-    version: "1.18.0"
+    version: "1.17.0"
```
PiotrRogulski, line 770: Several packages are now downgraded? 👀

PiotrRogulski (dismissed):

PiotrRogulski (approved):

piotruela (approved):

.github/workflows/test-ios-simulator.yaml
```diff
@@ -50,6 +50,11 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@v4
 
+      - name: Select Xcode
+        uses: maxim-lobanov/setup-xcode@v1
```
Komoszek: Pin to SHA. IMHO we should also pin all other actions to the SHAs instead of versions

dev/e2e_app/android/gradle.properties
```diff
@@ -1,2 +1,6 @@
 org.gradle.jvmargs=-Xmx3512M
 android.useAndroidX=true
+# Kotlin 2.3.x incremental compilation fails on Windows CI with
+# "Could not close incremental caches ... *.tab" file-locking errors.
+# Incremental compilation gives no benefit on clean CI builds, so disable it.
+kotlin.incremental=false
```
pdenert: How this will affect local testing?

Kendru98: Hmm need to check, but maybe then we can just set this option on CI/CD

pdenert: Yeah, that will be the best solution

.github/workflows/test-ios-simulator.yaml
Kendru98: should be fixed, waiting for workflows

pdenert (approved):

PR #3126, opened 2026-06-29
Workflows/clean up not needed test

Kendru98:
These tests are not supposed to be run on Ci/CD I accidentialy add tags last time to them. But currently we dont have any test to be run with PlatformSelector so this can be considered as something to be add. The second test is removed, because test that open patrol docs, and interacte with it in webview, already cover this.

### DIFF
--- a/dev/e2e_app/patrol_test/selector_test.dart
+++ b/dev/e2e_app/patrol_test/selector_test.dart
@@ -1,40 +1,18 @@
 import 'common.dart';
 
 void main() {
-  patrol(
-    'PlatformSelector allows different selectors per platform',
-    ($) async {
-      await createApp($);
+  patrol('PlatformSelector allows different selectors per platform', ($) async {
+    await createApp($);
 
-      await $('Open webview (Hacker News)').scrollTo().tap();
-      await $.pump(Duration(seconds: 5));
+    await $('Open webview (Hacker News)').scrollTo().tap();
+    await $.pump(Duration(seconds: 5));
 
-      await $.platform.tap(
-        PlatformSelector(
-          android: AndroidSelector(text: 'login'),
-          ios: IOSSelector(label: 'login'),
-          web: WebSelector(text: 'login'),
-        ),
-      );
-    },
-    tags: ['android', 'emulator', 'ios', 'simulator'],
-  );
-
-  patrol(
-    'MobileSelector allows different selectors for mobile platforms',
-    ($) async {
-      await createApp($);
-
-      await $('Open webview (Hacker News)').scrollTo().tap();
-      await $.pump(Duration(seconds: 5));
-
-      await $.platform.mobile.tap(
-        MobileSelector(
-          android: AndroidSelector(text: 'login'),
-          ios: IOSSelector(label: 'login'),
-        ),
-      );
-    },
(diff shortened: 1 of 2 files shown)


dev/e2e_app/patrol_test/selector_test.dart
PiotrRogulski, line 1: No tags on this test anymore?

dev/e2e_app/patrol_test/selector_test.dart
```diff
@@ -1,40 +1,31 @@
...
+        timeout: const Duration(seconds: 10),
       );
-    },
-    tags: ['android', 'emulator', 'ios', 'simulator'],
-  );
+    } on PatrolActionException catch (_) {
+      // nothing
+    }
+    await Future<void>.delayed(const Duration(seconds: 5));
+    // Open the search modal and type a query into its input field.
```
jBorkowska: where's the typing?

dev/e2e_app/patrol_test/selector_test.dart
Kendru98: I added them by mistake earlier (tags) and this test was not run on CI/CD, so I will keep it off CI for now. However, we can probably consider it in the future. It will require time for debugging though, which I don't have right now.

dev/e2e_app/patrol_test/selector_test.dart
Kendru98: It was copied from another test. There is not need to type anything in this test. - Removed

dev/e2e_app/patrol_test/selector_test.dart
```diff
@@ -1,40 +1,31 @@
...
+    await $.platform.mobile.tap(
```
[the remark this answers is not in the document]
jBorkowska: How is it resolved? Why do we have web selector here?

dev/e2e_app/patrol_test/selector_test.dart
```diff
@@ -1,40 +1,30 @@
...
-    },
-    tags: ['android', 'emulator', 'ios', 'simulator'],
-  );
-
-  patrol(
-    'MobileSelector allows different selectors for mobile platforms',
-    ($) async {
-      await createApp($);
+    // Dismiss the cookie consent popup if it appears. The exact button label
+    // can vary, so try the common variants and ignore if none are present.
```
PiotrRogulski: This test only tries one variant (`Selector(text: 'ACCEPT ALL COOKIES'),`

dev/e2e_app/patrol_test/selector_test.dart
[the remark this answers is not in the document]
Kendru98: This test purpose is (if im understand it correctly, it is not mine test I accidentially add tags to these tests earlier). To check if this test will work with PlatformSelector on Web/iOs/Android. But this is excluded from Ci/Cd for now, i have no time to debug it, we want to fix workflows for now, by this PR. I just adjust it a little be to cover idea of testing it on 3 platforms.

jBorkowska (approved):

dev/e2e_app/patrol_test/selector_test.dart
Kendru98: fixed

PiotrRogulski (approved):

PR #3133, opened 2026-07-03
fix(patrol): stop web tests hanging between bundled tests

### LINKED ISSUE #3132: Web: page.waitForFunction hangs ~10min between bundled tests (DDC reload / 2nd test)
# Patrol web: hang between tests (`page.waitForFunction` timeout)

Bug report draft for upstream: [leancodepl/patrol](https://github.com/leancodepl/patrol/issues/new).

## Summary

Running a multi-scenario Patrol web suite on Chrome hangs between tests. Playwright
`page.waitForFunction` blocks for **~10 minutes**, then fails. The Flutter app and
the previous test already finished; the runner stalls before the next scenario starts.

Observed on **small viewport (412×915)**. Same class of failure was seen earlier on
other viewport sizes during suite transitions.

## Environment

| Component | Version |
|-----------|---------|
| `patrol_cli` | 4.4.0 |
| `patrol` (package) | ^4.6.1 |
| Flutter | 3.41.2 (FVM) |
| OS | macOS 26.2 (darwin-arm64) |
| Browser | Chrome 149 (Playwright-managed) |
| Device | `patrol test --device chrome` |

## Steps to reproduce

1. Configure `.patrol.env` with valid test credentials (see `patrol_test/README.md`).
2. Boot no mobile emulator — web only.
3. Run the platform-aware runner:

   ```bash
   ./tool/run_patrol.sh web small
   ```

   Equivalent raw command (from `tool/run_patrol.sh`):

   ```bash
   DEBUG=pw:api patrol -v test --device chrome \
     --show-flutter-logs \
     --web-viewport='{"width": 412, "height": 915}' \
     --web-workers=1 \
     -t patrol_test/scenarios/contract_details_test.dart \
     -t patrol_test/scenarios/cookie_settings_test.dart \
     # … remaining scenarios …
   ```

4. Wait for the first scenario to complete.

## Expected behaviour

After `contract_details_test` passes, Patrol/Playwright should start
`cookie_settings_test` within a few seconds (DDC hot reload or page transition).

## Actual behaviour

1. `contract_details_test` **passes** (~10s).
2. Playwright starts `cookie_settings_test`: `page.goto` → `load` → DDC loads 2108
   scripts → `page.waitForFunction` starts.
3. `networkidle` fires after ~500ms.
4. **`page.waitForFunction` hangs for ~10 minutes**, then fails when
   `browserContext.close` runs (test timeout).
5. Suite crawls forward one test every ~10 minutes or appears stuck indefinitely.

### Log excerpt (2026-07-02)

```
✅ contract details page displays tariff details (/scenarios/contract_details_test.dart) (10s)
00:10 +2: scenarios.cookie_settings_test user changes cookie settings after login
…
pw:api => page.waitForFunction started +1ms
pw:api   "networkidle" event fired +499ms
pw:api => browserContext.close started +10m
pw:api <= page.waitForFunction failed +5ms
```

Full run log: terminal capture from `./tool/run_patrol.sh web small` on 2026-07-02
(~17 min, stopped manually after repeated 10m stalls).

### Secondary noise (may be unrelated)

After `contract_details_test` completed, Flutter reported a post-test async error:

```
FirebaseError: Remote Config: The stream was not able to connect to the backend
(remoteconfig/stream-error).
This test failed after it had already completed.
```

The scenario was already marked success in Patrol; the hang happens on the **next**
test's Playwright bootstrap, not inside the Flutter test body.

## Workarounds (project-side)

| Workaround | Notes |
|------------|-------|
| `./tool/run_patrol.sh web large` (1920×1080) | Sometimes completes 13/13; still flaky on other viewports |
| Run scenarios one-by-one (`-t` single file) | Avoids inter-test transition; impractical for CI |
| `--web-workers=1` | Already set; does not fix the hang |
| `DEBUG=pw:api` + `--show-flutter-logs` | Helps diagnose; does not fix |

**Do not use 1280×800** — known to hang in Patrol (separate from this report).

## Suggested upstream issue title

```
Web: page.waitForFunction hangs ~10min between bundled tests (DDC reload / 2nd test)
```

## Suggested labels

`bug`, `platform: web`, `playwright`

## References

- Patrol web docs: https://patrol.leancode.co/documentation/platforms/web
- Project runner: `tool/run_patrol.sh` → `run_patrol_web_viewport()`
- Related: Firebase/Playwright page lifecycle between sequential `patrol test` targets

Kendru98:
## Problem

Closes #3132.

Running a multi-scenario Patrol **web** suite hangs between tests: after one scenario passes, the runner stalls before the next one starts and eventually times out. It reproduces in **headed** runs (the reporter's `patrol test --device chrome` on macOS) while headless CI is unaffected.

## Root cause

The Playwright runner (shipped inside the `patrol` package under `web_runner/`) waits for the freshly-loaded Flutter app to expose its JS hooks using `page.waitForFunction(...)`. `waitForFunction` **defaults to `requestAnimationFrame` polling**, and the browser pauses `rAF` while the page/tab is not visible (an occluded or unfocused window — common in a headed run).

Each web test runs in a fresh browser context/page, so between bundled tests the tab briefly loses foreground during teardown + re-creation. That latches the `rAF`-based poll into a paused state, so the wait for `window.__patrol__runTest` (and the setup/initialise hooks) never re-evaluates — even though the Dart side has already exposed them (test discovery and `runAppService` complete via timers/microtasks, not frames). The result is a hang between tests. Headless CI never hides the page, so `rAF` keeps firing and the suite passes — which also explains why CI (which additionally passes `--web-headless true`) is green.

## Fix

Switch the three `waitForFunction` calls in the runner (`setup.ts`, `initialise.ts`, `test.spec.ts`) to **time-based polling** (`polling: 100`). Timer-based polling keeps re-evaluating regardless of tab visibility (Playwright already launches Chromium with `--disable-background-timer-throttling`), so the waits no longer stall when the window is occluded/unfocused. Timeouts and predicates are unchanged.

## Verification

- `tsc --noEmit`, `eslint`, and `prettier --check` all pass on `web_runner`.
- End-to-end: ran two web tests from `dev/e2e_app` (`web_example_navigation_test` → `web_example_dark_mode_test`) with the local CLI. Both pass and the **inter-test transition** — the exact point that hangs in the report — is clean:

  ```
  ✅ navigation (/web/web_example_navigation_test.dart) (9s)
  🧪 web.web_example_dark_mode_test dark mode
  ✅ dark mode (/web/web_example_dark_mode_test.dart) (5s)
  Total: 2  Successful: 2  Failed: 0
  ```

The original hang is timing/environment-specific (flaky, headed-macOS), so it can't be reproduced deterministically; this change targets the mechanism that makes the wait stall and is a strictly more robust poll. Running headless (`--web-headless true`, as CI does) remains a reliable workaround.

### DIFF
--- a/packages/patrol/CHANGELOG.md
+++ b/packages/patrol/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Migrate to built-in Kotlin (#3084).
 - Fix iOS tests failing on devices whose name contains a comma 
 - Migrate Japanese text resources to SPM (#3128)
+- Fix web tests hanging between bundled tests by using time-based polling instead of `requestAnimationFrame` polling in the Playwright runner, which is paused while the page is not visible in headed runs. (#3133)
 
 ## 4.6.1
 
--- a/packages/patrol/web_runner/tests/initialise.ts
+++ b/packages/patrol/web_runner/tests/initialise.ts
@@ -13,6 +13,10 @@ export async function initialise(page: Page) {
 
       return true
     },
-    { timeout: 60000 },
+    // Time-based polling instead of the default `requestAnimationFrame`
+    // polling, which is paused while the page/tab is not visible and can hang
+    // this wait in headed runs. See
+    // https://github.com/leancodepl/patrol/issues/3132.
+    { timeout: 60000, polling: 100 },
   )
 }
--- a/packages/patrol/web_runner/tests/setup.ts
+++ b/packages/patrol/web_runner/tests/setup.ts
@@ -41,7 +41,11 @@ async function setup(config: FullConfig) {
             // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
             return window.__patrol__getTests?.()!
           },
-          { timeout: 120000 },
+          // Time-based polling instead of the default `requestAnimationFrame`
+          // polling, which is paused while the page/tab is not visible and can
+          // hang this wait in headed runs. See
+          // https://github.com/leancodepl/patrol/issues/3132.
+          { timeout: 120000, polling: 100 },
         )
         .then(v => v.jsonValue()),
       setupPageErrorPromise,
--- a/packages/patrol/web_runner/tests/test.spec.ts
+++ b/packages/patrol/web_runner/tests/test.spec.ts
@@ -39,6 +39,13 @@ for (const { name, skip, tags } of tests) {
 
     await page.waitForFunction(() => window.__patrol__runTest, {
       timeout: 300000,
+      // Use time-based polling instead of the default `requestAnimationFrame`
+      // polling. `rAF` is paused by the browser while the page/tab is not
+      // visible (e.g. an occluded or unfocused window in a headed run), which
+      // can make this wait hang between tests even though the Dart side has
+      // already exposed `__patrol__runTest`. See
+      // https://github.com/leancodepl/patrol/issues/3132.
+      polling: 100,
     })
 
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion


Kendru98:
@JanGarstecki could you check whether this fixes the hang in your suite?

The fix lives entirely in the `patrol` package (the Playwright `web_runner`), so you don't need to reinstall `patrol_cli` — just point the `patrol` dependency at this branch and re-run your suite.

In your app's `pubspec.yaml`:

```yaml
dependency_overrides:
  patrol:
    git:
      url: https://github.com/leancodepl/patrol.git
      ref: claude/magical-clarke-10f928
      path: packages/patrol
```

Then:

```bash
flutter pub get
# your usual run, e.g.
patrol test --device chrome \
  --web-viewport='{"width": 412, "height": 915}' \
  --web-workers=1 \
  -t patrol_test/scenarios/contract_details_test.dart \
  -t patrol_test/scenarios/cookie_settings_test.dart
```

(The `web_runner`'s Node deps are reinstalled automatically on the next `patrol test`.)

Would be great to know if the ~10-minute stalls between tests are gone. If it still hangs, could you also try adding `--web-headless true` — that tells us whether the remaining stall is the Flutter engine pausing rendering while the window is hidden (a case interval polling alone can't cover), which would point at a follow-up.

PR #3136, opened 2026-07-03
fix(patrol_finders): enter text into edit text with hint finder

pdenert:
### Problem

When `enterText` is called with a string that matches a `TextField`'s `hintText`, the finder resolves to the `RichText` widget that renders the hint inside `InputDecorator`. This widget is purely visual and is never hit-testable, so `waitUntilVisible` (which applies `.hitTestable()` internally) always timed out — even though the field was fully visible on screen.

```dart
// Previously threw WaitUntilVisibleTimeoutException
await $('You have entered this text').enterText('Hello, Flutter!');
```

### Root cause

`createFinder` maps a `String` to `find.text(s, findRichText: true)`, which matches any `Text` or `RichText` in the tree — including hint-text nodes. The hint `RichText` and the actual `EditableText` are **siblings** under `InputDecorator` (not in an ancestor/descendant relationship), so simply searching for an `EditableText` ancestor of the matched node found nothing.

### Solution

At the start of `enterText`, before calling `waitUntilVisible`, check whether the finder already has matches but none are hit-testable. If so, walk up the element tree from the matched widget until an ancestor is found that contains an `EditableText` descendant. That `EditableText` is used as the effective target for the rest of the method — it is hit-testable, and the existing descendant-search logic that follows already handles it correctly.

No new imports are required: `EditableText` is already available from `package:flutter/widgets.dart`.


Kendru98 (commented):
Why test fail? https://github.com/leancodepl/patrol/actions/runs/28671327296/job/85035038668?pr=3136

PR #3141, opened 2026-07-07
Bump equatable

Polyphemus980:
Use with Equatable instead of the deprecated with EquatableMixin


packages/patrol_mcp/pubspec.lock
```diff
@@ -12,10 +12,9 @@ packages:
   adb:
     dependency: "direct main"
     description:
-      name: adb
-      sha256: "1c1d53b356bc9a7b89e7abcbb8b7b941f6d6ab0fc61a77eb8a4928befa294202"
-      url: "https://pub.dev"
-    source: hosted
+      path: "../adb"
+      relative: true
+    source: path
```
PiotrRogulski: Why did this change? (btw, this lock file (and other lock files) shouldn't be checked in, most probably)

Polyphemus980: No idea tbh, but removed lock files.

pdenert (commented):
Add changelog entry to all changed packages

PiotrRogulski (approved):

PR #3143, opened 2026-07-08
Add CI workflows for patrol_mcp

zoskar:
## What

`patrol_mcp` had no CI coverage. This adds workflows that actually test the MCP server, not just that it resolves.

## Cases covered

**Server lifecycle** — blocking, on `patrol_mcp` changes
- The MCP server starts and completes the `initialize` handshake.
- It shuts down cleanly on client disconnect (stdin EOF) without orphaning its process subtree.
- Verified on Linux and Windows.

**Dependency-constraint honesty** — blocking, on `patrol_mcp` changes
- `patrol_mcp` builds and starts against both the newest and the lowest `patrol_cli` its constraint allows.
- Catches the case where `patrol_mcp` starts using a newer `patrol_cli` API without bumping its constraint — which would otherwise ship broken to users on an older `patrol_cli`.

**`patrol_cli` compatibility heads-up** — non-blocking, on `patrol_cli` changes
- When a `patrol_cli` change breaks `patrol_mcp` against local source, the author gets a warning + job summary. Never fails CI — released users are unaffected because `patrol_mcp` is pinned to a published `patrol_cli`.

### DIFF
--- a/.github/workflows/patrol_mcp-cli-compat.yaml
+++ b/.github/workflows/patrol_mcp-cli-compat.yaml
@@ -1,14 +1,8 @@
 name: patrol_mcp cli-compat
 
-# Non-blocking heads-up for patrol_cli authors.
-#
-# patrol_cli exposes a public API barrel (lib/patrol_cli.dart) specifically for
-# patrol_mcp. A breaking change to it does NOT break released patrol_mcp users
-# (patrol_mcp is pinned to a *published* patrol_cli), so this must never fail CI.
-# But the author may not realize they broke patrol_mcp against the current source
-# — this job builds patrol_mcp against THIS PR's patrol_cli (local, via melos)
-# and, if it breaks, surfaces a warning annotation + job summary. It always
-# exits green.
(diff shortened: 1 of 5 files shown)


.github/workflows/patrol_mcp-cli-compat.yaml
```diff
@@ -0,0 +1,58 @@
+name: patrol_mcp cli-compat
+
+# Non-blocking heads-up: does this patrol_cli change break patrol_mcp when built
+# against local source? Released users are safe (patrol_mcp pins a published
+# patrol_cli), so on a break we only warn — never fail CI.
```
pdenert: Is it valuable to keep this explanation?

pdenert (commented):
I guess this PR should include changes in workflow.md

.github/workflows/patrol_mcp-cli-compat.yaml
zoskar: I removed it and put in workflow.md; I forgot about it

pdenert (approved):

PR #3169, opened 2026-07-15
test(patrol_finders): add regression test for keys with special characters

### LINKED ISSUE #1117: Finder can't find a widget by key, if there are special characters in key's string argument
For example, if key that we're searching has 'ł' in it, it can't be found, though there is a widget with such a key in the widget tree (I checked inspector and it appears there correctly).

pdenert:
## Summary

Closes #1117.

The issue reported that a finder couldn't find a widget by key when the key's string contained special characters (e.g. Polish `ł`). It was verified in the issue to work since Flutter 3.10.5, and the remaining action item was to add a widget test covering this case.

This PR adds a regression test to `patrol_finder_test.dart` that finds widgets by keys containing Polish diacritics, special symbols (`/`, `\`, `.`, `:`, `!@#$%^&*()`), and emoji, plus a negative case for a non-matching key.

## Verification

- New test passes; full `patrol_finders` suite green (132 tests), `flutter analyze` clean, `dart format` applied.


packages/patrol_finders/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased
```
zoskar, line 1: Are tests needed in changelog?

zoskar (approved):

packages/patrol_finders/CHANGELOG.md
pdenert: Rather not, claude went to far. Eventually I will do the cleanup before the release

PR #3170, opened 2026-07-16
Add findAttachedDevices to DevelopSessionFactory

zoskar:
Adds `DevelopSessionFactory.findAttachedDevices`, a static helper that lists the devices Flutter reports as attached (Android, iOS, macOS, web) for non-CLI callers, redirecting `Logger` output to `stderr`. Pure additive change — no effect on existing `patrol` commands.

Groundwork for multi-device support in `patrol_mcp` (#3162), which lands in a separate PR. Split out so `patrol_cli` can be released with this API first, then `patrol_mcp` bumps its constraint to it.

### DIFF
--- a/packages/patrol_cli/CHANGELOG.md
+++ b/packages/patrol_cli/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - Download only Chromium instead of all default Playwright browsers during web runner setup. (#3156)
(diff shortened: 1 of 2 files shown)


zoskar:
Superseded by #3162: patrol_mcp now lists devices by running `flutter devices --machine` directly, so this patrol_cli helper is no longer needed. Closing.

PR #3176, opened 2026-07-16
Add patrol-setup skill to user-facing catalog

zoskar:
Adds a `patrol-setup` agent skill to the published user-facing skills catalog (`skills/`) and lists it in `skills/README.md`.

The skill is deliberately **minimal** — it encodes the ordered process, the decisions/gotchas an agent would otherwise guess, and precise deep-links into the canonical docs, rather than duplicating them. It covers: configuring the pubspec `patrol:` block, wiring native Android, getting a first self-contained test passing on an emulator, and a closing pointer to Patrol MCP.

### Docs
Also fixes a real gap in the setup guide (`docs/documentation/index.mdx`) that the skill points at:
- gitignore now uses `**/test_bundle.dart` (covers the mobile test-dir **and** the web project-root location) and adds `.patrol.env`
- adds a minimal "Using AI?" pointer in *Going from here* → Patrol MCP + agent skills


skills/patrol-setup/SKILL.md
```diff
@@ -0,0 +1,58 @@
+---
+name: patrol-setup
```
Kendru98, line 2: It could be missleading by name, someone can think that it will also do whole setup for iOS etc. what is not possible I guess, maybe change name to have there that this is pubpsec and android setup

zoskar: I don't know about name change but I made it more clear in skill/desc that it's minimal setup. AI when it's done will report that ios is another step

pdenert (approved):

PR #3182, opened 2026-07-21
Add web coverage support

### LINKED ISSUE #3121: [web] patrol test --coverage crashes before Playwright starts with Bad state: No element
### Steps to reproduce

Run any Patrol web test with coverage enabled, for example:

```bash
patrol test \
  --device chrome \
  --web-headless true \
  --coverage \
  --target patrol_test/sign_in_test.dart
```

The same command without `--coverage` proceeds to the Playwright/browser test run normally. With `--coverage`, the command crashes before Playwright starts executing tests.

### Actual results

```text
Building Flutter web app...
Unhandled exception:
Bad state: No element
#0      Stream.first.<anonymous closure> (dart:async/stream.dart:1653:32)
<asynchronous suspension>
#1      CoverageTool.run.<anonymous closure> (package:patrol_cli/src/coverage/coverage_tool.dart:101:30)
<asynchronous suspension>
#2      DisposeScope.run (package:dispose_scope/src/dispose_scope.dart:46:7)
<asynchronous suspension>
#3      CoverageTool.run (package:patrol_cli/src/coverage/coverage_tool.dart:70:5)
<asynchronous suspension>
```

### Expected results

Either:

1. Patrol web coverage is supported and an LCOV report is generated for the web Patrol run, or
2. Patrol exits with a clear message that `--coverage` is not supported for `--device chrome` / web yet.

### Investigation notes

From a quick read of current Patrol CLI sources, this looks web-specific:

- `packages/patrol_cli/lib/src/commands/test.dart` starts `_coverageTool.run(...)` before `_execute(...)`.
- For `TargetPlatform.web`, the web app is not built before that point; the web backend starts the Flutter web server later in `_webTestBackend.execute(...)`.
- `packages/patrol_cli/lib/src/coverage/coverage_tool.dart` starts `flutter logs -d <device.id>` and waits for VM service details using `.first` on `vmConnectionDetailsStream`.
- With `device.id == chrome`, there are no VM service details yet, so the stream completes empty and throws `Bad state: No element`.

I found https://github.com/leancodepl/patrol/issues/2794 with a similar `Bad state: No element` coverage crash for Android/iOS simulator coverage, but I could not find an existing web-specific coverage issue.

### Latest version check

The latest stable `patrol_cli` on pub.dev is `4.4.0`. I also tested the latest `patrol_cli` prerelease currently listed on pub.dev, `4.5.0-dev.1`, and it crashes with the same stack trace.

I also checked the `patrol 4.7.0-dev.1` changelog. It does not mention web coverage or changes to the coverage tool, and there is no matching `patrol_cli 4.7.0-dev.1` version available on pub.dev to test.

### Patrol version

```text
patrol_cli v4.4.0
patrol package: 4.6.1
```

Also reproduced with:

```text
patrol_cli v4.5.0-dev.1
```

### Patrol Doctor output

```text
Patrol doctor:
Patrol CLI version: 4.4.0
Flutter command: flutter 
  Flutter 3.44.0 • channel stable
Android: 
• Program adb not found (install with `export PATH="$ANDROID_HOME/platform-tools:$PATH"`)
• Env var $ANDROID_HOME is not set. (See the link: https://developer.android.com/tools/variables#set)
iOS / macOS: 
• Program xcodebuild found in /usr/bin/xcodebuild
• Program ideviceinstaller found in /opt/homebrew/bin/ideviceinstaller
Web: 
• Program node found in /opt/homebrew/bin/node
• Program npm found in /opt/homebrew/bin/npm
```

### Flutter Doctor output

```text
Flutter 3.44.0 • channel stable
Dart 3.12.0
```

### Environment

```text
macOS 27.0 26A5368g
Node v26.4.0
npm 11.17.0
```

Komoszek:
Closes #3121

## What was implemented

`patrol test --coverage` now works on the web platform. Coverage was previously
collected only on Android/iOS from the Dart VM service; the browser has no Dart VM,
so web runs produced no report. This change collects the browser's native JavaScript
coverage during the Playwright run and maps it back to Dart sources, producing the
**same `coverage/patrol_lcov.info` LCOV report** as the mobile flow.

Notable differences, surfaced in `--help` and the docs:

- Web reports **covered / uncovered only**, without the per-line hit counts.
- Every **non-debug build** error out with `--coverage`. It previously just silently skipped coverage collection

## How it works

During a web test run the Playwright runner records Chrome's native JavaScript
coverage (with source maps) for each test. Afterwards the CLI reads that data,
maps the covered JS back to the original Dart lines through the source maps, and
writes the same LCOV report as on mobile. The VM and web paths share the report
formatting and package/workspace filtering, so the output is consistent across
platforms.

## Results

Ran the existing `text_fields_test.dart` on both an Android emulator and Chrome with
`--coverage`. Both produce a valid report over the same file set; below is the
coverage of the exercised widget (`text_fields_screen.dart`):

| Platform | Lines | Covered | % |
|---|---|---|---|
| Android (VM) | 17 | 17 | 100% |
| Web (V8 + source maps) | 21 | 20 | 95.2% |

The two agree on the widget body. The gap is the documented behaviour: web is binary
covered/uncovered (no hit counts) and the toolchains differ slightly on which lines
are "coverable" — web instruments a few more lines (e.g. the field initializer and the
`const` constructor, which it marks uncovered) that the VM omits.

### DIFF
--- a/packages/patrol_cli/lib/src/coverage/web_coverage_tool.dart
+++ b/packages/patrol_cli/lib/src/coverage/web_coverage_tool.dart
@@ -1,0 +1,258 @@
+import 'dart:convert';
+
+import 'package:coverage/coverage.dart' as coverage;
+import 'package:file/file.dart';
+import 'package:glob/glob.dart';
+import 'package:patrol_cli/src/base/logger.dart';
+import 'package:patrol_cli/src/coverage/coverage_common.dart';
+
+/// Generates a Dart coverage report for web test runs.
+///
+/// The browser has no Dart VM, so instead of a VM service the Playwright runner
+/// records Chrome's JS coverage per test (with source maps) into
+/// [dataDirectory]. This tool maps it back to Dart and writes the same
+/// `coverage/patrol_lcov.info` report as the mobile flow.
+class WebCoverageTool {
+  WebCoverageTool({
+    required FileSystem fs,
+    required Directory rootDirectory,
+    required Logger logger,
+  }) : _fs = fs,
+       _rootDirectory = rootDirectory,
+       _logger = logger;
+
+  final FileSystem _fs;
+  final Directory _rootDirectory;
+  final Logger _logger;
+
+  /// Where the Playwright runner drops one JSON file per test with the raw V8
+  /// coverage entries.
+  Directory get dataDirectory => _rootDirectory
+      .childDirectory('.dart_tool')
+      .childDirectory('patrol')
+      .childDirectory('web_coverage');
+
+  /// Clears data from previous runs and returns the directory to write to.
+  Future<Directory> prepareDataDirectory() async {
+    final dir = dataDirectory;
+    if (dir.existsSync()) {
+      await dir.delete(recursive: true);
+    }
+    await dir.create(recursive: true);
+    return dir;
+  }
+
+  Future<void> run({
+    required String flutterPackageName,
+    required Set<RegExp> packagesRegExps,
+    required Set<Glob> ignoreGlobs,
+    bool includeWorkspacePackages = false,
+  }) async {
+    final dir = dataDirectory;
+    final files = dir.existsSync()
+        ? dir
+              .listSync()
+              .whereType<File>()
+              .where((file) => file.path.endsWith('.json'))
+              .toList()
+        : <File>[];
+
+    if (files.isEmpty) {
+      _logger.warn(
+        'No web coverage data was collected, skipping coverage report. '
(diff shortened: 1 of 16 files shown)


packages/patrol_cli/lib/src/coverage/web_coverage_tool.dart
pdenert, line 1: Some of the comments are exaggerated descriptive

pdenert (approved):
LGTM

PR #3189, opened 2026-07-23
Add posthog to docs app

alszczep:
Adds posthog integration to the docs app. Main reason is getting access to feature flags, experiments and similar features. Includes skill and mcp config as an easy path of adding new flags.


.agents/skills/docs-posthog-feature-flags/SKILL.md
konowrockis, line 1: Please use skills by default 🙏

docs_app/lib/posthog/posthog-bootstrap.ts
```diff
@@ -0,0 +1,3 @@
+import type { BootstrapConfig } from "posthog-js"
+
+export type PostHogBootstrap = BootstrapConfig
```
konowrockis: Why do we need to rename that? 😅

.cursor/rules/posthog-feature-flags.mdc
```diff
@@ -0,0 +1,65 @@
+---
+description: PostHog feature flags and experiments for docs_app — create/update via PostHog MCP and wire them in code. Use when adding, changing, removing, or debugging flags or A/B tests.
+alwaysApply: true
```
konowrockis: This IMO is really... brave to enable docs posthog feature flags always in patrol repository, where probably 99.9% of work don't touch this topic 😅

docs_app/lib/posthog/posthog-server.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+async function readDistinctId() {
+  const token = process.env.NEXT_PUBLIC_POSTHOG_PROJECT_TOKEN
+  if (!token) {
+    return crypto.randomUUID()
+  }
+
+  const cookieStore = await cookies()
+  const cookie = cookieStore.get(`ph_${token}_posthog`)
```
konowrockis: Is it something documented? Or is it just some hacky-hacky workaround?

pdenert:
@alszczep can you add description that will explain what is this PR about?

PR #3197, opened 2026-07-24
Build-time test discovery + static native test codegen (experimental, opt-in)

piotruela:
## What & why

Stock Patrol discovers tests at **runtime** (dynamic registration once the app
launches), so no static list of test identifiers exists before a run. On iOS —
where XCUITest has no built-in sharding — that blocks per-test sharding entirely.

This PR adds **experimental, opt-in build-time test discovery**: a host
`flutter test` serializes the Dart test tree to a manifest at build time, and we
generate a real, statically-declared native test per Dart test. Every test becomes
individually addressable (`-only-testing` on iOS, `am instrument -e class#method`
on Android).

**This should unblock iOS test sharding on device farms that shard by an explicit
test list — SauceLabs (`testListFile`) and BrowserStack** — plus balanced Android
sharding and cleaner per-test report names.

## What's included

- iOS static XCTest codegen (new STATIC runner macros) + Android static JUnit codegen.
- Config opt-in: `patrol.emit_test_manifest: true` (the `--emit-test-manifest`
  flag overrides it); iOS build fails fast if `RunnerUITests.m` isn't switched to
  the static macro.
- `patrol test --no-build [--only "<dart name>"]`: re-run built tests without
  rebuilding (iOS `xcodebuild test-without-building`, Android `am instrument`).
- Discovery reporting (tests grouped by source file) + docs page.

## Usage

```yaml
# pubspec.yaml
patrol:
  emit_test_manifest: true
```

```bash
patrol build ios --emit-test-manifest         # or: patrol build android
patrol test --no-build --only "my group my test"   # one test, no rebuild
```

iOS also needs the static macro in `RunnerUITests.m` (see docs; the CLI prints the
exact snippet if missing).

## Verification

Unit tests (manifest/selector/codegen/config) + `dart analyze` clean. Verified
end-to-end in `dev/e2e_app`: iOS single test via `-only-testing` → `Executed 1
test`; Android via `am instrument` → `OK (1 test)`; `--no-build --only` runs one
test with no rebuild on both platforms.

## Limitations

Experimental. Host-side discovery can diverge for tests registered conditionally
on the runtime env (`Platform.isX`, dart-define-derived names). Tests must live
outside `integration_test/` (use `patrol_test/`). Parallel simulators on one host
collide on Patrol's fixed ports — shard across separate hosts/VMs (as farms do).
Initial build is slightly slower (paid once; `--no-build` re-runs skip building).


jBorkowska:
I have questions and general feedback:
1. Why opt-in?
2. [iOS] When switching between emit-test-list and no-emit-test-list, you have to modify `RunnerUITests.m`. If you forget about it, then you waste a few minutes trying to build the tests. It would be better (and I think it's doable) if patrol determined which way to go or just generate testlist always.
3. If it would stay like it is now, we need better instructions how to use it.
4. How does it interfere with tags? Tags are resolved in runtime (afaik)

pdenert:
> * Why opt-in?

There are lot's of limitations and changes that will be breaking if we made it default

piotruela:
@jBorkowska 

1. Opt-in because enabling it is breaking: existing projects' `RunnerUITests.m` uses
   the runtime macro, and discovery can't run at all for tests kept in
   `integration_test/` (flutter_tools treats that directory as integration tests, so
   it demands a device and a direct `integration_test` dependency). It also let us
   prove the thing first — iOS sharding on SauceLabs is green now on both simulators
   and real devices.
2. I'd rather keep it explicit in `pubspec.yaml` under `patrol:` than infer it — it's
   a project-level decision, not something you flip per run. You shouldn't lose
   minutes to it either: if the setting and `RunnerUITests.m` disagree, the build
   stops immediately and prints the exact macro to paste, instead of failing somewhere
   in xcodebuild.
3. Fair — the page is here:
   https://patrol.leancode.co/documentation/ci/build-time-test-discovery. I've just
   added the parts that were missing for setup: tests have to live outside
   `integration_test/`, what the discovery output looks like (that's where the name
   for `--only` comes from), and that the generated sources are build output to
   gitignore.
4. Tags are resolved when the test tree is built, not natively: they're baked into the
   generated bundle, so the manifest and the generated native tests already respect
   them. Related — `--no-build` is now its own command, `patrol test-without-building`,
   which doesn't accept build-shaping flags like `--tags` at all instead of accepting
   and ignoring them.

Polyphemus980 (approved):

PR #3200, opened 2026-07-30
fix(patrol): surface uncaught exceptions in patrol develop mode

Kendru98:
## What & why

- `patrol develop` swallowed uncaught exceptions thrown async from callbacks (e.g. a widget's `onPressed`): the step showed ✅ and the test looked passed. Same test under `patrol test` fails with a stack trace.
- Fix: after the test callback, read the framework-captured exception via `takeException()` and report it as a failure through the Patrol log.

## Info for reviewers

- Develop can't use the exception gatherer (it must not override `FlutterError.onError`, per #2613). This reads the already-captured exception instead, so it doesn't re-trigger that assertion and leaves the Hot Restart loop (`r`/`h`/`q`) untouched.
- Draft scope: surfaces the exception **message** only (no full stack yet); failing step still renders ✅; no non-zero exit (develop has no pass/fail channel). Async-case regression test for `dev/cli_tests/patrol_develop_test.dart` is a TODO (needs device CI).

## Verified

- iOS simulator on the [repro](https://github.com/alexbuntoweovum/patrol-develop-exceptions): develop now prints `❌ … / Exception: …` and stays interactive. `patrol test` unchanged (only the `hotRestartEnabled` path is touched).


packages/patrol/lib/src/common.dart
```diff
@@ -176,6 +176,32 @@ void patrolTest(
...
+      // framework but never surfaced: the develop test intentionally never
+      // completes (it waits in the loop below), so nobody reports them, and the
+      // test looks like it passed. Pull any framework-captured exception and
+      // report it through the Patrol log so it becomes visible in `patrol
+      // develop` (and via patrol_mcp), without ending the Hot Restart session.
+      if (constants.hotRestartEnabled) {
+        // Pump once so exceptions from in-flight gesture callbacks are recorded
+        // by the framework before we read them.
+        await widgetTester.pump();
+        final caughtException = patrolBinding.takeException();
```
piotruela: Would it make sense to run the same check inside this loop? The app stays interactive here, so an exception from a manual tap would get reported as it happens rather than not at all. Not sure whether pumping and reading every 10 ms causes problems.

piotruela: `takeException()` gives back only details.exception, so we lose the stack. `reportExceptionNoticed` in `binding.dart` already gets the full `FlutterErrorDetails` and throws them away, so storing them there and reading them here might be enough to get file and line into the output. I haven't checked whether that hook fires reliably in develop mode.

piotruela (commented):
Two suggestions on top of this, both aimed at making develop report the same thing patrol test would. I haven't run either on a device, so I'm not certain they hold up in practice.

Running the same `takeException()` check inside the wait loop would also catch exceptions thrown while clicking around the app after the test, which is where a develop session spends most of its time.
Keeping the `FlutterErrorDetails` that `PatrolBinding.reportExceptionNoticed` already receives (`binding.dart`) would get the stack trace into the output instead of just the message.

packages/patrol/lib/src/common.dart
Kendru98: Done in 3404a50f3 — the same `takeException()` check now runs inside the wait loop, so exceptions from manual taps while iterating get reported as they happen. Verified on an iOS simulator: the loop keeps pumping and the Hot Restart session stays attached; pumping + reading every 10 ms caused no problems in my run.

Kendru98: Tried it (3404a50f3): `reportExceptionNoticed` does fire in develop mode, so I keep the `FlutterErrorDetails` there and read it back — it gives the full stack incl. file:line.

Caveat worth deciding on: in `patrol develop` the failure is transported over stdout log-scraping (not the native result channel `patrol test` uses). On iOS the long stack line gets **truncated mid-frame** by the log transport, so the `PATROL_LOG {…}` JSON is incomplete and `patrol_log_reader` throws → `Error parsing line`. So the stack is captured fine but doesn't render cleanly over develop's channel.

Options: (a) keep message-only for clean rendering, (b) emit a compact single-line summary (message + first frame → still gives file:line, short enough to survive), or (c) make develop's log transport handle long/multi-line entries. Leaning (b); happy to go whichever way.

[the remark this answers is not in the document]
Kendru98: Agreed — the async `onPressed` path isn't covered by the existing synchronous failing-`expect` test. I'll extend `dev/cli_tests/patrol_develop_test.dart` to hot-restart into an `onPressed`-throwing test and assert the failure/error log + that the develop session stays attached, once we settle the output format (see the stack-trace truncation thread).

packages/patrol/lib/src/common.dart
```diff
@@ -176,6 +176,32 @@ void patrolTest(
         rethrow;
       }
 
+      // In develop (Hot Restart) mode the exception gatherer is disabled (see
```
Kendru98: Make this comment much shorter

packages/patrol/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Unreleased
+
+- Surface uncaught exceptions in `patrol develop` (Hot Restart) mode. Previously an exception delivered asynchronously to `FlutterError.onError` (e.g. thrown from a widget's `onPressed` callback) was captured by the test framework but never reported, so the test appeared to pass. It is now logged as a failure without ending the Hot Restart session. (#3200)
```
Kendru98: Keep 1 line in changelog entry

packages/patrol/lib/src/common.dart
Kendru98: Trimmed to 3 lines in 9c760d57e.

packages/patrol/CHANGELOG.md
Kendru98: Done — one line now (9c760d57e).

packages/patrol/lib/src/common.dart
Kendru98: Switched to this approach in e113c9c75.

Key realization: develop **already** surfaces full stacks for *rethrown* exceptions (the synchronous failing-`expect` case — the existing `patrol_develop_test.dart` asserts `'When the exception was thrown'` shows up). flutter_test dumps those to the console as normal multi-line text and `patrol develop` forwards raw console lines, so no truncation.

My earlier truncation was self-inflicted: I packed the whole multi-line stack into one long single-line `PATROL_LOG {…}` JSON, which the iOS log transport cuts mid-frame → invalid JSON → `Error parsing line`. Wrong channel for a big payload.

So instead of routing the stack through `PATROL_LOG`, develop mode now dumps the `FlutterErrorDetails` to the console in `PatrolBinding.reportExceptionNoticed` (it was a no-op override that swallowed them — which is exactly why `onPressed` was silent). That's the same path the sync case already uses, so the full stack (file+line) shows untruncated. The `PATROL_LOG` entry stays short, just for the structured ❌ status.

Net: no need to fix the log transport or keep the details plumbing. Still want to confirm the exact rendering with one develop run on a device.

Kendru98: Verified on an iOS simulator (patrol develop on the repro):

- Full `EXCEPTION CAUGHT BY GESTURE` block now prints, with the stack pointing at `example_test.dart:52:17` (the `throw`) and `:62:7` (the tap site).
- No more `Error parsing line` — the long stack goes to the console (multi-line), not into a single `PATROL_LOG` line.
- Structured `❌ … (/example_test.dart)` still renders, and Hot Restart stays attached.

piotruela (approved):

PR #3216, opened 2026-08-05
fix(patrol): make Android UiAutomation accessibility flags configurable

### LINKED ISSUE #3178: Android: native selectors can't find WebView content since 4.0 — root-caused to uiautomator 2.3.0 force-enabling FLAG_RETRIEVE_INTERACTIVE_WINDOWS
### Steps to reproduce

Upgrading patrol 3.20.0 → 4.x breaks every `$.native` selector that targets text inside
the WebView of the Plaid Link SDK (`com.plaid.link:sdk-core`, a translucent
`singleInstance` Activity hosting a WebView). Fails 100% of runs, passes 100% on 3.20.0.

Environment: Android emulator `sdk_gphone64_arm64`, Android 15 (API 35), Android System
WebView 124.0.6367.219, Flutter 3.38.5, patrol 4.7.1 / patrol_cli 4.5.1.

Reproduction (real-world): a Plaid Link flow in any app — open Link (sandbox works),
then:

```dart
await $.native.waitUntilVisible(Selector(text: 'Get started'));
```

**Honest note on a minimal public repro:** I could not reproduce this with simple
WebViews, despite trying hard. All of the following PASS on stock patrol 4.7.1
(uiautomator 2.3.0), same emulator:

- your own `$.native` webview e2e tests removed in #3113 (I restored and ran them on
  both uiautomator versions — all green, so their removal as "flaky" was probably
  unrelated to this bug),
- a WebView in a plain native Activity,
- the same native Activity with Plaid's exact manifest config (`singleInstance` +
  `windowIsTranslucent` theme),
- UiAutomator operations fired before/after WebView creation, multiple WebViews, etc.

So an additional app/window condition (present in the Plaid case) is required, which I
couldn't isolate publicly. However, the root cause inside patrol's dependency IS
isolated — see below — and is verifiable in any app with Plaid Link.

### Actual results

```
PatrolActionException: Patrol action failed: AndroidAutomatorClientException:
waitUntilVisible() failed with Invalid response: 404 selector
UiSelector[TEXT=Get started] found nothing
```

The native tree at that moment contains the WebView node **with no children** — its
virtual accessibility hierarchy is never populated — while the screen visibly shows the
fully rendered Plaid UI.

**Root cause, isolated to a single flag.** uiautomator 2.3.0's
`UiDevice.getUiAutomation()` rewrites the UiAutomation's `AccessibilityServiceInfo`
(2.2.0 never touched it). The only net change it makes is adding
`FLAG_RETRIEVE_INTERACTIVE_WINDOWS` (flags 65554 → 65618, logcat: `UiDevice: Setting
accessibility service flags: 65618`). A/B/B/A on the identical test flow:

| uiautomator | `FLAG_RETRIEVE_INTERACTIVE_WINDOWS` | result |
|---|---|---|
| 2.3.0 stock | set (by its rewrite) | ❌ fails |
| 2.3.0, rewrite suppressed¹ | not set | ✅ passes |
| 2.2.0 stock | not set | ✅ passes |
| 2.2.0 + only that flag added manually² | set | ❌ fails, identical error |

¹ suppressed by pre-syncing `UiDevice.mCachedServiceFlags` via reflection so the
`serviceInfo.flags != mCachedServiceFlags` check never fires `setServiceInfo()`.
² one `setServiceInfo(flags | FLAG_RETRIEVE_INTERACTIVE_WINDOWS)` call added at the end
of `Automator.configure()`, everything else stock 2.2.0.

Row 4 is the clincher: the old, known-good query implementation fails the moment the
flag is set, so nothing else in 2.3.0 matters. With the flag on (multi-window
accessibility tracking mode), the Plaid WebView's lazy tree population never happens
when queried; with it off, the first `findObjects` populates the tree in ~650 ms and
everything works.

Also verified NOT the cause: Chromium's accessibility state events. With WebView
verbose logging, `cr_A11yState` logs the identical `New AccessibilityState:
{...isAnyAccessibilityServiceEnabled=false...}` transition ~50 ms into the first native
query in BOTH the passing (2.2.0) and failing (2.3.0) runs — red herring.

**Fix options** (maintainers' call):

- Clear `FLAG_RETRIEVE_INTERACTIVE_WINDOWS` after uiautomator adds it. Caveat: a plain
  `setServiceInfo()` won't stick — uiautomator re-adds the flag whenever
  `serviceInfo.flags != mCachedServiceFlags`, so the private cache must also be synced
  (reflection; the field is `UiDevice.mCachedServiceFlags` in 2.3.0). I've verified the
  suppression approach green on a full real-world suite.
- Report to google/android-test that 2.3.0 unconditionally forces this flag with no
  opt-out and breaks WebView accessibility in some window configurations, and bump when
  fixed. (You're much better placed than an app developer to carry that conversation.)
- Revert to uiautomator 2.2.0 (patrol uses no 2.3.0-only APIs — 4.7.1 compiles and
  passes against 2.2.0 unmodified), at the cost of 2.3.0's gesture/display fixes.

Meanwhile the user-side workaround is:

```groovy
// android/build.gradle (allprojects)
configurations.all {
    resolutionStrategy { force 'androidx.test.uiautomator:uiautomator:2.2.0' }
}
```

### Logs

<details>
<summary>Logs</summary>

Failing run (stock 4.7.1, uiautomator 2.3.0) — flag set at test start, first query
never sees the WebView content:

```
20:37:37.836 UiAutomationManager: Registering UiTestAutomationService (flags=0x1)
20:37:37.839 UiDevice: Setting accessibility service flags: 65618      ← 2.3.0 adds FLAG_RETRIEVE_INTERACTIVE_WINDOWS
20:38:20.854 ActivityTaskManager: Displayed com.plaid.internal.link.LinkActivity
20:38:49.194 UiDevice: Retrieving nodes with selector: BySelector [TEXT='\QGet started\E']
20:38:49.241 cr_A11yState: New AccessibilityState: {... isAnyAccessibilityServiceEnabled=false ...}
(every retry finds nothing; WebView node has no children)
```

Passing control (same flow, uiautomator 2.2.0 forced) — same Chromium state event, but
the query populates the tree and succeeds:

```
20:45:54.108 ActivityTaskManager: Displayed com.plaid.internal.link.LinkActivity
20:46:20.372 PatrolServer: waitUntilVisible(): UiSelector[TEXT=Get started]
20:46:20.430 cr_A11yState: New AccessibilityState: {... isAnyAccessibilityServiceEnabled=false ...}
20:46:21.024 PatrolServer: POST /waitUntilVisible took 655 ms   ← succeeded
```

Flag-isolation run (uiautomator 2.2.0 + only the flag added) — fails identically:

```
20:54:23.072 PatrolServer: EXPERIMENT: flags before=65554
20:54:23.075 PatrolServer: EXPERIMENT: flags after=65618
...
PatrolActionException: ... 404 selector UiSelector[TEXT=Get started] found nothing
```

</details>

### Patrol version

`patrol: 4.7.1` (every 4.x affected; last known-good `patrol: 3.20.0`)

### Patrol Doctor output

<details>
<summary>Patrol Doctor output</summary>

```
Patrol doctor:
Patrol CLI version: 4.5.1
Flutter command: flutter
  Flutter 3.38.5 • channel [user-branch]
Android:
• Program adb found in /Users/<user>/Library/Android/sdk/platform-tools/adb
• Env var $ANDROID_HOME is not set. (See the link: https://developer.android.com/tools/variables#set)
iOS / macOS:
• Program xcodebuild found in /usr/bin/xcodebuild
• Program ideviceinstaller found in /opt/homebrew/bin/ideviceinstaller
Web:
• Program node found in /opt/homebrew/bin/node
• Program npm found in /opt/homebrew/bin/npm
```

</details>

### Flutter Doctor output

<details>
<summary>Flutter Doctor output</summary>

```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.38.5, on macOS 26.5 25F71 darwin-arm64, locale en-IL)
[✓] Android toolchain - develop for Android devices (Android SDK version 36.1.0)
[✓] Xcode - develop for iOS and macOS (Xcode 26.5)
[✓] Chrome - develop for the web
[✓] Connected device (3 available)
[✓] Network resources

• No issues found!
```

</details>

### LINKED ISSUE #3201: Android: UiAutomation is acquired with flags=0, suppressing all third-party AccessibilityServices for the whole session
### Problem

For the whole duration of a Patrol Android session, **every third-party `AccessibilityService` on the device is suppressed by the platform**. This makes Patrol mutually exclusive with any tool built on an `AccessibilityService` (screen readers, external UI-automation bridges, accessibility auditing tools) on the same device — they cannot be used side by side, only by tearing the Patrol session down completely.

This is not a platform inevitability: it is the consequence of one line, and uiautomator already exposes the documented knob to change it.

### Root cause

`packages/patrol/android/src/main/kotlin/pl/leancode/patrol/Automator.kt`, `initialize()`:

```kotlin
if (!this::uiDevice.isInitialized) {
    uiDevice = UiDevice.getInstance(instrumentation)
}
if (!this::uiAutomation.isInitialized) {
    uiAutomation = instrumentation.uiAutomation   // <-- Kotlin property = no-arg overload
}
```

`instrumentation.uiAutomation` resolves to `Instrumentation.getUiAutomation()`, which AOSP defines as:

```java
public UiAutomation getUiAutomation() {
    return getUiAutomation(0);
}
```

and the flags overload reuses an existing instance **only when the flags match**, otherwise it disconnects and reconnects with the new flags:

```java
public UiAutomation getUiAutomation(@UiAutomationFlags int flags) {
    boolean mustCreateNewAutomation = (mUiAutomation == null) || (mUiAutomation.isDestroyed());
    if (mUiAutomationConnection != null) {
        if (!mustCreateNewAutomation && (mUiAutomation.getFlags() == flags)) {
            return mUiAutomation;
        }
        if (mustCreateNewAutomation) { /* new UiAutomation(...) */ }
        else { mUiAutomation.disconnect(); }
        ...
        mUiAutomation.connectWithTimeout(flags, CONNECT_TIMEOUT_MILLIS);
        ...
    }
}
```

<sub>Sources: [`Instrumentation.java` master L2508/L2547](https://raw.githubusercontent.com/aosp-mirror/platform_frameworks_base/master/core/java/android/app/Instrumentation.java) · [android-14.0.0_r1 L2309/L2347](https://raw.githubusercontent.com/aosp-mirror/platform_frameworks_base/android-14.0.0_r1/core/java/android/app/Instrumentation.java) · [android-11.0.0_r1 L2130/L2161](https://raw.githubusercontent.com/aosp-mirror/platform_frameworks_base/android-11.0.0_r1/core/java/android/app/Instrumentation.java) · [android-7.0.0_r1 L1866/L1897](https://raw.githubusercontent.com/aosp-mirror/platform_frameworks_base/android-7.0.0_r1/core/java/android/app/Instrumentation.java) — the flag-decision logic is identical across all four.</sub>

`UiAutomation.connectWithTimeout(flags, …)` then does `mUiAutomationConnection.connect(mClient, flags); mFlags = flags;`, and on the server side:

```java
// services/accessibility/java/com/android/server/accessibility/UiAutomationManager.java
private boolean suppressingAccessibilityServicesLocked() {
    return (mUiAutomationService != null || !useAccessibility())
        && ((mUiAutomationFlags & UiAutomation.FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES) == 0);
}
```

`flags == 0` ⇒ suppression is on. (`FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES` is `0x00000001`, added in API 24; per [the official docs](https://developer.android.com/reference/android/app/UiAutomation#FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES) *"UiAutomation suppresses accessibility services by default"*, and with the flag set *"existing services will continue to run and new ones may start"*.)

So there are two distinct effects:

1. **Suppression is on for the entire session**, because Patrol asks for flags `0`.
2. **Patrol silently downgrades flags established earlier in the process.** Because the no-arg overload means `flags == 0`, if anything before `Automator.initialize()` obtained a `UiAutomation` with non-zero flags, Patrol's call does not reuse it — it forces a `disconnect()` + reconnect at `0`. Worth noting that the logcat in #3178 shows `UiAutomationManager: Registering UiTestAutomationService (flags=0x1)` at session start, i.e. *something* in the stack does establish `FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES` before Patrol's `Automator` runs. I have not traced which component that is, but under the reuse rule above the end state after `initialize()` is `flags = 0` either way.

Effect 2 may have a second, unrelated symptom: `UiDevice.getUiAutomation()` carries the comment *"Verify and update the accessibility service flags if necessary. These might get reset if the underlying UiAutomationConnection is recreated."* — so a redundant reconnect can also reset the `AccessibilityServiceInfo` uiautomator had just configured. #2448 (`Error while connecting UiAutomation@…[id=-1, displayId=0, flags=0]`) might be related, though I have not confirmed that.

### Suggested fix

uiautomator already provides the intended extension point, and `Automator` already holds a `Configurator` instance, so this is a one-liner in `Automator.configure()`:

```kotlin
configurator.uiAutomationFlags = UiAutomation.FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES
```

Why that is the right layer:

- `Configurator.setUiAutomationFlags(int)` is public API since uiautomator 2.2.0, and its javadoc explicitly references this exact flag (`@see android.app.UiAutomation#FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES`).
- `UiDevice.getUiAutomation()` passes `Configurator.getInstance().getUiAutomationFlags()` straight through to `Instrumentation.getUiAutomation(flags)`; the default is `Configurator.DEFAULT_UIAUTOMATION_FLAGS = 0`.
- `UiDevice.getUiAutomation()` itself is package-private, so reusing uiautomator's instance directly is not an option for Patrol — going through `Configurator` is.

Two design notes for maintainers:

- **Please consider making it opt-in rather than flipping the default.** Not suppressing accessibility services can change behaviour for apps that react to `isAnyAccessibilityServiceEnabled`, so a `patrol` config field / dart-define that defaults to today's behaviour would be the safe shape.
- Independently of the flag policy, `initialize()` arguably should not call `instrumentation.uiAutomation` with no flags *after* `UiDevice.getInstance(instrumentation)`. That is the line that clobbers whatever flags were established. Sourcing the flags from `Configurator` in both places keeps them consistent and avoids the extra disconnect/reconnect.

### Use case

Agent-driven app exploration on a single device: Patrol drives the Flutter widget tree (via `patrolTest` + `WidgetTester`), while a separate `AccessibilityService`-based CLI handles what Patrol's Flutter side cannot see — system permission dialogs, native pickers, cross-app UI.

Today those two cannot coexist. Handing off from Flutter to the accessibility-based tool requires stopping the in-app bridge, quitting `patrol develop`, and sending `SIGINT` to the `PatrolJUnitRunner` instrumentation host to release the `UiAutomation`. That handoff is **irreversible within a session** — returning to the Flutter side means building and launching a whole new Patrol session, which for us also means re-establishing transient login state. With the flag set, both could run against the same device with no handoff at all.

### Versions

- `patrol` 4.6.1 (also verified unchanged on `master` and in the 4.8.0 changelog — nothing between 4.6.1 and 4.8.0 touches `UiAutomation` flags)
- `patrol_cli` 4.4.0
- `androidx.test.uiautomator:uiautomator:2.3.0` (pinned in `packages/patrol/android/build.gradle`)
- Android real device

### What I did not verify

- Which component logs `Registering UiTestAutomationService (flags=0x1)` before `Automator.initialize()` runs.
- Whether #2448 is actually caused by the redundant disconnect/reconnect.
- I have not run a build with the suggested one-line change; the analysis above is source-level only.

Kendru98:
Make two Android `UiAutomation` accessibility flags configurable. Both opt-in, default = today's behavior.

**Problem**
- #3201: Patrol acquires `UiAutomation` with `flags=0`, so the platform suppresses every third-party `AccessibilityService` for the whole session.
- #3178: uiautomator 2.3.0 force-enables `FLAG_RETRIEVE_INTERACTIVE_WINDOWS`, which stops some WebViews (e.g. Plaid Link) from populating their a11y tree, breaking `$.native` selectors.

**Change**
- `AndroidAutomatorConfig.dontSuppressAccessibilityServices` (dart-define `PATROL_ANDROID_DONT_SUPPRESS_ACCESSIBILITY_SERVICES`) → sets `FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES` via `Configurator`.
- `AndroidAutomatorConfig.retrieveInteractiveWindows` (dart-define `PATROL_ANDROID_RETRIEVE_INTERACTIVE_WINDOWS`) → `false` clears the flag.
- Flags flow config → `ConfigureRequest` contract → `Automator.configure()`; iOS ignores them.

**Watch out**
- uiautomator re-adds the flag when `serviceInfo.flags != mCachedServiceFlags` or every `SERVICE_FLAGS_TIMEOUT` (2s), so a plain `setServiceInfo()` won't stick — `mCachedServiceFlags`/`mLastServiceFlagsTime` are pinned via reflection (guarded, degrades to default on failure).
- `FLAG_INCLUDE_NOT_IMPORTANT_VIEWS` is preserved so normal selectors keep working.

**Result**
- 14 Dart tests pass (4 new, covering config defaults + `ConfigureRequest` threading).
- Native flag effect needs on-device verification (couldn't run an emulator here).

Closes #3201
Closes #3178

### DIFF
--- a/packages/patrol/android/src/main/kotlin/pl/leancode/patrol/Automator.kt
+++ b/packages/patrol/android/src/main/kotlin/pl/leancode/patrol/Automator.kt
@@ -1,5 +1,6 @@
 package pl.leancode.patrol
 
+import android.accessibilityservice.AccessibilityServiceInfo
 import android.app.Instrumentation
 import android.app.UiAutomation
 import android.content.Context
@@ -105,11 +106,26 @@ class Automator private constructor() {
         }
     }
 
-    fun configure(waitForSelectorTimeout: Long) {
+    fun configure(
+        waitForSelectorTimeout: Long,
+        dontSuppressAccessibilityServices: Boolean = false,
+        retrieveInteractiveWindows: Boolean = true,
+    ) {
         timeoutMillis = waitForSelectorTimeout
         configurator.waitForSelectorTimeout = waitForSelectorTimeout
         configurator.waitForIdleTimeout = 5000
         configurator.keyInjectionDelay = 50
+
+        // Opt-in (#3201): keep third-party AccessibilityServices alive during the
+        // session instead of letting the platform suppress them.
+        if (dontSuppressAccessibilityServices) {
+            configureDontSuppressAccessibilityServices()
+        }
+        // Opt-in (#3178): undo uiautomator 2.3.0 force-enabling
+        // FLAG_RETRIEVE_INTERACTIVE_WINDOWS, which breaks WebView tree population.
+        if (!retrieveInteractiveWindows) {
+            clearRetrieveInteractiveWindowsFlag()
+        }
 
         Logger.i("Timeout: $timeoutMillis ms")
         Logger.i("Android UiAutomator configuration:")
@@ -119,6 +135,89 @@ class Automator private constructor() {
         Logger.i("\tactionAcknowledgmentTimeout: ${configurator.actionAcknowledgmentTimeout} ms")
         Logger.i("\tscrollAcknowledgmentTimeout: ${configurator.scrollAcknowledgmentTimeout} ms")
         Logger.i("\ttoolType: ${configurator.toolType}")
+        Logger.i("\tuiAutomationFlags: ${configurator.uiAutomationFlags}")
+        Logger.i("\tdontSuppressAccessibilityServices: $dontSuppressAccessibilityServices")
+        Logger.i("\tretrieveInteractiveWindows: $retrieveInteractiveWindows")
+    }
+
+    // See https://github.com/leancodepl/patrol/issues/3201.
+    private fun configureDontSuppressAccessibilityServices() {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
+            Logger.i("dontSuppressAccessibilityServices requires API 24+, ignoring")
+            return
+        }
+
+        // UiDevice.getUiAutomation() forwards Configurator.uiAutomationFlags to
+        // Instrumentation.getUiAutomation(flags), so setting it here is enough for
+        // uiautomator's own queries.
+        configurator.uiAutomationFlags = UiAutomation.FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES
+
+        // getUiAutomation(flags) reuses an existing instance only when the flags
+        // match, otherwise it reconnects. Re-acquiring our own reference forces the
+        // reconnect now and keeps it consistent with what UiDevice will use.
+        uiAutomation = uiAutomationForConfiguredFlags()
+        Logger.i("Acquired UiAutomation with FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES")
+    }
+
+    // Instrumentation.getUiAutomation(flags) exists only on API 24+. Below that,
+    // the flags-based opt-ins are unavailable and Configurator.uiAutomationFlags
+    // stays at its default (0), so the no-arg overload is equivalent.
+    private fun uiAutomationForConfiguredFlags(): UiAutomation {
+        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            instrumentation.getUiAutomation(configurator.uiAutomationFlags)
+        } else {
+            instrumentation.uiAutomation
+        }
+    }
+
+    // See https://github.com/leancodepl/patrol/issues/3178.
+    private fun clearRetrieveInteractiveWindowsFlag() {
+        try {
+            // Same instance UiDevice queries with (both route through
+            // Instrumentation.getUiAutomation(Configurator.uiAutomationFlags)).
+            val automation = uiAutomationForConfiguredFlags()
+            val serviceInfo = automation.serviceInfo
+            if (serviceInfo == null) {
+                Logger.e("Cannot clear FLAG_RETRIEVE_INTERACTIVE_WINDOWS: serviceInfo is null")
+                return
+            }
+
+            // Mirror uiautomator 2.3.0's own (non-compressed) flag setup, minus the
+            // interactive-windows bit. Keeping FLAG_INCLUDE_NOT_IMPORTANT_VIEWS
+            // preserves normal selector behavior.
+            var flags = serviceInfo.flags
+            flags = flags or AccessibilityServiceInfo.FLAG_INCLUDE_NOT_IMPORTANT_VIEWS
+            flags = flags and AccessibilityServiceInfo.FLAG_RETRIEVE_INTERACTIVE_WINDOWS.inv()
+            serviceInfo.flags = flags
+            automation.setServiceInfo(serviceInfo)
+            uiAutomation = automation
+
+            // UiDevice.getUiAutomation() re-adds FLAG_RETRIEVE_INTERACTIVE_WINDOWS
+            // whenever serviceInfo.flags != mCachedServiceFlags, or every
+            // SERVICE_FLAGS_TIMEOUT (2s). Pin its private cache so it never rewrites
+            // the flags we just set.
+            pinUiDeviceServiceFlags(flags)
+
+            Logger.i("Cleared FLAG_RETRIEVE_INTERACTIVE_WINDOWS (serviceInfo.flags=$flags)")
+        } catch (e: Throwable) {
+            Logger.e("Failed to clear FLAG_RETRIEVE_INTERACTIVE_WINDOWS, keeping default behavior", e)
+        }
+    }
+
+    private fun pinUiDeviceServiceFlags(flags: Int) {
+        val uiDeviceClass = uiDevice.javaClass
+        uiDeviceClass.getDeclaredField("mCachedServiceFlags")
+            .apply { isAccessible = true }
+            .setInt(uiDevice, flags)
+        try {
+            uiDeviceClass.getDeclaredField("mLastServiceFlagsTime")
+                .apply { isAccessible = true }
+                .setLong(uiDevice, Long.MAX_VALUE)
+        } catch (e: NoSuchFieldException) {
+            // Older uiautomator without the periodic re-check; syncing the cached
+            // flags alone is enough there.
+            Logger.i("mLastServiceFlagsTime not present, cache sync alone is enough")
+        }
     }
 
     private fun executeShellCommand(cmd: String) {
--- a/packages/patrol/lib/src/platform/android/android_automator_config.dart
+++ b/packages/patrol/lib/src/platform/android/android_automator_config.dart
@@ -9,6 +9,8 @@ class AndroidAutomatorConfig extends MobileAutomatorConfig {
     String? packageName,
     String? appName,
     KeyboardBehavior? keyboardBehavior,
+    bool? dontSuppressAccessibilityServices,
+    bool? retrieveInteractiveWindows,
     super.host,
     super.port,
     super.connectionTimeout,
@@ -19,7 +21,18 @@ class AndroidAutomatorConfig extends MobileAutomatorConfig {
            const String.fromEnvironment('PATROL_APP_PACKAGE_NAME'),
        appName =
            appName ?? const String.fromEnvironment('PATROL_ANDROID_APP_NAME'),
-       keyboardBehavior = keyboardBehavior ?? KeyboardBehavior.showAndDismiss;
+       keyboardBehavior = keyboardBehavior ?? KeyboardBehavior.showAndDismiss,
+       dontSuppressAccessibilityServices =
+           dontSuppressAccessibilityServices ??
+           const bool.fromEnvironment(
+             'PATROL_ANDROID_DONT_SUPPRESS_ACCESSIBILITY_SERVICES',
+           ),
+       retrieveInteractiveWindows =
+           retrieveInteractiveWindows ??
+           const bool.fromEnvironment(
+             'PATROL_ANDROID_RETRIEVE_INTERACTIVE_WINDOWS',
+             defaultValue: true,
+           );
 
   /// How the keyboard should behave when entering text.
   ///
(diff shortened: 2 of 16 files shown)


packages/patrol/android/src/main/kotlin/pl/leancode/patrol/Automator.kt
```diff
@@ -119,6 +135,89 @@ class Automator private constructor() {
...
+        // match, otherwise it reconnects. Re-acquiring our own reference forces the
+        // reconnect now and keeps it consistent with what UiDevice will use.
+        uiAutomation = uiAutomationForConfiguredFlags()
+        Logger.i("Acquired UiAutomation with FLAG_DONT_SUPPRESS_ACCESSIBILITY_SERVICES")
+    }
+
+    // Instrumentation.getUiAutomation(flags) exists only on API 24+. Below that,
+    // the flags-based opt-ins are unavailable and Configurator.uiAutomationFlags
+    // stays at its default (0), so the no-arg overload is equivalent.
+    private fun uiAutomationForConfiguredFlags(): UiAutomation {
```
jBorkowska, line 165: Is this method used anywhere else than in the method above? Seems redundant

packages/patrol/lib/src/platform/android/android_automator_config.dart
```diff
@@ -19,7 +21,18 @@ class AndroidAutomatorConfig extends MobileAutomatorConfig {
            const String.fromEnvironment('PATROL_APP_PACKAGE_NAME'),
        appName =
            appName ?? const String.fromEnvironment('PATROL_ANDROID_APP_NAME'),
-       keyboardBehavior = keyboardBehavior ?? KeyboardBehavior.showAndDismiss;
+       keyboardBehavior = keyboardBehavior ?? KeyboardBehavior.showAndDismiss,
+       dontSuppressAccessibilityServices =
+           dontSuppressAccessibilityServices ??
+           const bool.fromEnvironment(
```
jBorkowska, line 27: No default value?

Kendru98:
Superseded by splitting into two PRs per team request: #3227 (accessibility-services default, targets master) and #3228 (FLAG_RETRIEVE_INTERACTIVE_WINDOWS, stacked on #3227).

PR #3222, opened 2026-08-07
feat: native failure screenshots on Android for device farms

Kendru98:
Opt-in native failure screenshots on Android for device farms, captured from the Dart failure path (before teardown) so the image is the real failing screen, not a teardown frame.

**Change**
- New `takeNativeScreenshot` on the `AndroidAutomator` contract; native `Automator` writes a PNG via `androidx.test` `Screenshot.capture()` to `Download/screenshots/<class>/<method>/` (MediaStore on API 29+, so scoped storage doesn't drop it).
- The `<class>/<method>` folder is the running test's JUnit identity, read from its `Description` by a `RunListener` — so it matches what the farm reports for both the parameterized host and the `emit_test_manifest` generated class, with no reconstruction.
- Triggered in the `patrolTest` failure `catch`, gated by `screenshot_on_failure`; swallows all errors so it never flakes a test. A native `TestWatcher` would fire too late.
- On-demand `await $.takeNativeScreenshot('tag')` (Android only; no-op on iOS).
- `adbPull`/`adbRemove` extracted and shared with the video recorder (no behavior change to video).

**Where it runs**
- `patrol test` (Android) pulls screenshots into `<test-directory>/screenshots` after the run (override `--screenshots-output-dir`), mirroring `--record-video` — works on emulators, independent of any farm.
- On farms patrol only writes the file — the farm collects it (BrowserStack `debugscreenshots`, Firebase Test Lab `--directories-to-pull`).
- BrowserStack renders the image only for URL-safe names, so use `emit_test_manifest` there (its generated per-test JUnit names are URL-safe); the default `runDartTest[...]` names contain spaces.

**Watch out**
- Kept `androidx.test:runner` at 1.5.1 (already ships `Screenshot`); `screenshot_on_failure` off by default, other runs unaffected.

**Verification**
- Dart/CLI: analyze + format clean; screenshot/adb/config unit tests pass.
- Capture + local pull verified end-to-end on an Android 16 emulator. The `RunListener` that records the JUnit name isn't locally compilable — validate on a BrowserStack build.


docs/documentation/integrations/firebase-test-lab.mdx
```diff
@@ -52,6 +52,12 @@ Farm][aws_device_farm].
       app, not other data on the device, e.g. Chrome.
     </Info>
 
+    <Info>
```
Kendru98, line 55: This is not tested but should work

Kendru98: Verified on Firebase Test Lab (MediumPhone.arm / API 34, --use-orchestrator + clearPackageData=true): the failure screenshot is pulled and also shows in the console Screenshots tab. Run: https://console.firebase.google.com/project/leancode-exampleapp/testlab/histories/bh.7e1ea131966f185b/matrices/4889775389655572934

packages/patrol_cli/lib/src/android/android_test_backend.dart
```diff
@@ -823,6 +833,52 @@ class AndroidTestBackend {
...
 
+  /// Where patrol writes native screenshots on the device.
+  static const _deviceScreenshotsDir = '/sdcard/Download/screenshots';
+
+  /// Pulls native screenshots from [device] into [outputDir]. Best-effort: never
+  /// throws; a missing directory (nothing captured) is not an error.
+  @visibleForTesting
+  Future<void> pullDeviceScreenshots(Device device, String outputDir) async {
+    try {
+      // adb pull recreates the `screenshots` basename under [parent].
```
piotruela: This honors only the parent of `--screenshots-output-dir`: `adb pull` recreates the `screenshots` basename, so `artifacts/my_shots` ends up as `artifacts/screenshots`. Pull straight into `outputDir` when it doesn't exist (adb renames the dir then), or rename after the pull.

packages/patrol_cli/lib/src/android/adb_commands.dart
piotruela, line 1: I'd add `pull`/`shell rm` to `package:adb` (it's ours, and this class already holds `_adb`) instead of new raw-`adb` helpers. One place for adb resolution and device handling; the video recorder started the raw pattern, this is a good moment to stop it spreading.

docs/documentation/index.mdx
```diff
@@ -132,6 +132,45 @@ Check out our video version of this tutorial on YouTube!
         ```
     </Info>
 
+    <Info>
```
piotruela, line 135: This whole naming-limitation story is already in `test.mdx` (and the changelog). Keep the full version in one place and link it, duplicated docs drift. Should the getting-started tutorial get this much of a device-farm feature at all? A short pointer would do.

packages/patrol_cli/lib/src/android/android_test_backend.dart
```diff
@@ -375,6 +376,8 @@ class AndroidTestBackend {
     required bool clearTestSteps,
     void Function(Entry entry)? onLogEntry,
     VideoRecordingConfig? videoConfig,
+    bool pullScreenshots = false,
+    String screenshotsOutputDir = 'screenshots',
```
piotruela: nit: this default is never used, `test.dart` always passes it.

packages/patrol_cli/lib/src/commands/test.dart
```diff
@@ -551,6 +555,8 @@ See https://github.com/leancodepl/patrol/issues/1316 to learn more.
           flavor: flutterOpts.flavor,
           clearTestSteps: clearTestSteps,
           videoConfig: videoConfig,
+          pullScreenshots: true,
```
piotruela, line 558: Leftovers in `/sdcard/Download/screenshots` from an aborted run or from `develop` (which captures but never pulls) land in this run's artifacts. Maybe clear the device dir at run start too, so screenshots always belong to the run that reports them?

packages/patrol_cli/lib/src/android/android_test_backend.dart
Kendru98: Done

packages/patrol_cli/lib/src/android/adb_commands.dart
Kendru98: Done

docs/documentation/index.mdx
Kendru98: Done

packages/patrol_cli/lib/src/android/android_test_backend.dart
Kendru98: Done

packages/patrol_cli/lib/src/commands/test.dart
Kendru98: Done

packages/patrol_cli/CHANGELOG.md
```diff
@@ -14,6 +14,8 @@
   without rebuilding, optionally just one of them with `--only`. Requires build-time test
   discovery. (#3197)
 - Add `--record-video` flag to `patrol test` and `patrol develop` to record a video per test case (Android and iOS simulators). (#2741)
+- Add `screenshot_on_failure` option to the pubspec's `patrol` section, forwarded to the app (via a dart-define) for `build` and `test` so patrol can capture native failure screenshots on Android device farms (e.g. BrowserStack, Firebase Test Lab). Not collected by `patrol develop`. Off by default. (#3222)
+- `patrol test` (Android) now pulls native screenshots (failure and on-demand) from the device into `<test-directory>/screenshots` after the run (override with `--screenshots-output-dir`), so they are available from local/emulator runs, not only device farms. (#3222)
```
piotruela: Something is wrong here. `## Unreleased` is below 4.6.1. Also some of the things mentioned in this section were already released

Kendru98: Done — merged master, so the released items are now under `## 4.7.0`; this PR's entries sit under a fresh `## Unreleased` at the top.

piotruela (approved):

PR #3225, opened 2026-08-11
Support latest package_config, cli_completion and pub_updater in patrol_cli

PiotrRogulski:
`patrol_cli` was scoring 0/10 on pana's *"All of the package dependencies are supported in the latest version"* check. It wasn't only `package_config` — three direct dependencies were capped below their newest major:

| Dep | Was | Latest | Latest needs |
|---|---|---|---|
| `cli_completion` | `^0.5.1` | 0.6.0 | Dart ^3.12.0 |
| `package_config` | `^2.1.1` | 3.0.0 | Dart ^3.11.0 |
| `pub_updater` | `^0.5.0` | 0.6.0 | Dart ^3.12.0 |

Every new major raises only *its own* SDK floor, so bumping the carets would have forced `patrol_cli`'s `sdk:` from `>=3.8.0` up to `^3.12.0` — dropping everyone below Flutter 3.44.

## What this does

Constrains each dependency across two majors instead, so pub picks the newest on new SDKs and the previous one on older SDKs. **No SDK constraint changes, here or in any other package.** `coverage` already constrains `package_config` as `>=2.0.0 <4.0.0`, so there's precedent in our own dependency graph.

No code changes were needed. `patrol_cli` only calls `loadPackageConfig(File)` and `findPackageConfig(Directory)`, whose signatures are identical in 2.x and 3.0.0, and 3.0.0's breaking change (dropping `.packages`) was never used here. The `cli_completion` and `pub_updater` 0.6.0 releases are SDK bumps with no API changes.

## Verification

Ran against Dart 3.12.2 (latest stable, what pana runs on) and Dart 3.8.1 (our declared floor):

- A path consumer resolves **3.0.0 / 0.6.0 / 0.6.0** on Dart 3.12.2 and **2.2.0 / 0.5.1 / 0.5.0** on Dart 3.8.1, and AOT-compiles on both.
- `dart analyze`, `dart run custom_lint` and `dart format` clean — both at the normal resolution and with `package_config` force-overridden to 3.0.0.
- `dart test`: failure set byte-identical to `master` (63 pre-existing failures, all `FlutterVersion.fromCLI` — no Flutter SDK in the sandbox).
- `dart pub downgrade` resolves the same lower bounds as before (2.1.1 / 0.5.1 / 0.5.0) and analyzes clean, so pana's 20-point lower-bound check is unaffected.
- pana locally: **140/160 → 150/160**, dependency section **30/40 → 40/40**.

Note that in the dev resolution `package_config` still lands on 2.2.0, because `custom_lint` caps it at `^2.0.2`. pana's check uses `--no-dev-dependencies`, so the points land regardless, and consumers on Dart 3.11+ get 3.0.0.

## Other packages

Checked all of them; `patrol_cli` was the only one losing points.

- `adb`, `patrol`, `patrol_finders`, `patrol_log`, `patrol_mcp` — all direct deps already resolve to latest. Their only outdated entries are `leancode_lint` and `custom_lint`, which are dev_dependencies and don't count toward pub points. Bumping `leancode_lint` ^17 → ^26 would need Dart ≥3.11 plus a lint-rule migration, so I left it as a separate call.
- `patrol_devtools_extension` and `patrol_gen` are `publish_to: none`, so no pub score. They do have genuinely stale deps needing code migration (`analyzer` ^7 → ^14, `devtools_extensions` ^0.3 → ^0.5, `flutter_hooks` ^0.20 → ^0.21) — out of scope here, but worth flagging.

## Unrelated thing spotted

`patrol_cli` declares `sdk: '>=3.8.0 <4.0.0'` but can't actually resolve on Dart 3.8.1, because `leancode_lint: ^19.0.1` needs ≥3.9.0. Dev-only, so consumers are unaffected and pana doesn't see it, but the declared floor is a little optimistic.

The CHANGELOG's `## Unreleased` section was also moved above `## 4.6.1`, where it belongs.

https://claude.ai/code/session_01RdBaZuNJ1qnwki2fxm93Qh

---
_Generated by [Claude Code](https://claude.ai/code/session_01RdBaZuNJ1qnwki2fxm93Qh)_


packages/patrol_cli/pubspec.yaml
```diff
@@ -14,12 +14,15 @@ screenshots:
...
 
+# Deps spanning two majors keep `sdk:` low: pub picks the newest major on new
+# SDKs, the previous one on old ones.
 dependencies:
   adb: ^0.6.0
   ansi_styles: ^0.3.2+1
   args: ^2.4.2
   ci: ^0.1.0
-  cli_completion: ^0.5.1
+  # 0.6.x needs Dart 3.12.
```
pdenert, line 24: remove this comment

packages/patrol_cli/pubspec.yaml
```diff
@@ -29,12 +32,14 @@ dependencies:
   http: ^1.1.0
   mason_logger: ^0.3.5
   meta: ^1.10.0
-  package_config: ^2.1.1
+  # 3.x needs Dart 3.11.
```
pdenert, line 35: ditto

packages/patrol_cli/pubspec.yaml
```diff
@@ -29,12 +32,14 @@ dependencies:
...
+  package_config: '>=2.1.1 <4.0.0'
   path: ^1.8.3
   patrol_log: ^0.10.0
   platform: ^3.1.3
   process: ^5.0.1
-  pub_updater: ^0.5.0
+  # 0.6.x needs Dart 3.12.
```
pdenert, line 41: ditto

packages/patrol_cli/pubspec.yaml
```diff
@@ -14,12 +14,15 @@ screenshots:
 environment:
   sdk: '>=3.8.0 <4.0.0'
 
+# Deps spanning two majors keep `sdk:` low: pub picks the newest major on new
+# SDKs, the previous one on old ones.
```
pdenert, line 18: ditto

pdenert (commented):
Please update the branch @PiotrRogulski

packages/patrol_cli/pubspec.yaml
PiotrRogulski: Hmm, I'd keep at least one, to explain why these deps are versioned differently from the other ones 🤔 (at least, that was the reason behind them originally)

pdenert (approved):

Issue #951, opened 2023-02-17
Test reports don't contain useful info ("Multiple exceptions were thrown")


bartekpacia:
This is unacceptable. We want to know what these exceptions are.

![Screenshot 2023-02-17 at 10 52 02 AM](https://user-images.githubusercontent.com/40357511/219613785-c500fc6e-a03a-4df6-84aa-0cb748868271.png)

The code that prints this message [lives here in `flutter_test`](https://github.com/flutter/flutter/blob/3.7.3/packages/flutter_test/lib/src/binding.dart#L790-L808) – in the `_runTest()` method of the abstract class [`TestWidgetsFlutterBinding`](https://api.flutter.dev/flutter/flutter_test/TestWidgetsFlutterBinding-class.html).

[`TestWidgetsFlutterBinding`](https://api.flutter.dev/flutter/flutter_test/TestWidgetsFlutterBinding-class.html) is extended by [`LiveTestWidgetsFlutterBinding`](https://api.flutter.dev/flutter/flutter_test/LiveTestWidgetsFlutterBinding-class.html).

[`LiveTestWidgetsFlutterBinding`](https://api.flutter.dev/flutter/flutter_test/LiveTestWidgetsFlutterBinding-class.html) is extended by [`IntegrationTestWidgetsFlutterBinding`](https://api.flutter.dev/flutter/package-integration_test_integration_test/IntegrationTestWidgetsFlutterBinding-class.html) from `integration_test`.

We should be able to override `FlutterError.onError` to print full exception message and stack trace.

cc @mateuszwojtczak, @jBorkowska

jeremiahlukus:
would be nice also to have print statements output in the patrol output

jBorkowska:
size: M

Noxiver:
@bartekpacia any solutions?

MickaelHrndz:
Please consider prioritizing this. It's the most upvoted bug issue. And it's been up for more than a year.
This testing package currently **doesn't tell you what the errors that make your test fail are**. 
If you can't, at least please make it not fail the test.

piotruela:
Hello everyone. We've just published a prerelease version of [patrol 3.16.0-dev.1](https://pub.dev/packages/patrol/versions/3.16.0-dev.1) and [patrol_cli 3.6.0-dev.1](https://pub.dev/packages/patrol_cli/versions/3.6.0-dev.1) that should fix this issue. Please let us know if it helps

PR #3244, opened 2026-08-14
fix: restore FlutterError.onError in a finally so a failing test always reports

Kendru98:
Fixes a failing test being able to hang an entire test shard.

### The bug

`_wrapTestBodyWithExceptionGatherer` installs a `FlutterError.onError` override and restores it only *after* `await testBody()`:

```dart
await testBody();                        // throws here…
FlutterError.onError = previousOnError;  // …so this never runs
```

A body that throws leaves the override installed. A later async error then trips flutter_test's *"a test overrode FlutterError.onError but failed to return it to its original state"* invariant, the test never completes, the Dart side never replies, and the native runner blocks on `runDartTest` until its 2h timeout.

Two things make this expensive:

- **Every test queued behind it on that shard never runs.**
- **The dead shard reports zero failures**, so CI looks green.

The leaked override also belongs to whichever *earlier* test threw, so the test that appears to hang is often an innocent bystander.

### The fix

`try`/`finally` around the body, plus a `catch` that records the thrown error.

The `catch` is not optional. A thrown failure — a failed `expect`, a finder timeout, a `PatrolActionException` — is only reported by flutter_test *after* it has unwound out of the body, via the zone handler calling `FlutterError.reportError`. The `finally` has already run by then, so the override is gone and nothing records the details; the native side would print the test name followed by a bare `null`.

I deliberately did **not** restore the `reportTestException` hook removed in #2362. That hook receives `_pendingExceptionDetails`, which collapses to the `Multiple exceptions were thrown` placeholder as soon as a second error arrives — reintroducing exactly the regression #2362 fixed (#951). The thrown object is always the real one.

A `Failure` already gathered by `onError` accumulates every framework error seen during the test, so it keeps precedence; the `catch` only fills the gap where there was nothing.

Passing tests are unaffected — the `finally` runs at the same point the old assignment did. Develop mode is unchanged, as the gatherer is not installed there.

### Verification

Measured on a 359-test Android suite on BrowserStack (Galaxy S24 Ultra, 2 shards, espresso with orchestrator):

| | executed | reported |
|---|---|---|
| before | **45 / 359** | `failures="0"` (two dead shards, killed at the 2h cap) |
| after | **365** | 4 real failures, with messages and stack traces |

The four failures were pre-existing and had been invisible for days behind the hang.

**Controlled A/B on the same commit.** Same 10 test files, same device, same shard count and BrowserStack config; the only variable is the patrol version:

| patrol | executed | shards | wall clock |
|---|---|---|---|
| 4.7.1 (unfixed) | **0 / 10** | both dead, 1 `timedout` each | 2h (killed at the cap) |
| this PR | **10 / 10** | both clean | 18 min |

On the unfixed build each shard died on its *first* test, and in both cases that test was one of the four that legitimately fail — `classic_pair_first_then_ble_pairing_patrol` and `ble_pairing_requires_classical_connection_patrol`. The remaining four tests on each shard never ran.

This also rules out 4.9.0 as the cause: 4.7.1 carries the same unfixed code and hangs the same way.

### What I could not reproduce

I could not trigger the wedge on a local x86 emulator via `patrol test`. Four shapes against unfixed 4.9.0, all reported normally instead of hanging:

1. one-shot `Timer` firing after the throw
2. `Timer.periodic` every 100 ms
3. unawaited `Future.error` from a `tearDown` — this one *did* make the **next** test fail with no message, i.e. the cross-test contamination, but no wedge
4. failing while a live `WebView` platform view is mounted

### DIFF
--- a/packages/patrol/CHANGELOG.md
+++ b/packages/patrol/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- Fix the whole test shard hanging when a test fails while an asynchronous error source (e.g. a still-running `WebView` platform view) is active. `FlutterError.onError` is now restored in a `finally`, so a failing test reports its result immediately instead of leaving the native runner blocked on `runDartTest` until its timeout. (#3243)
 - Android: keep third-party `AccessibilityService`s running during the test session again. `AndroidAutomatorConfig.dontSuppressAccessibilityServices` now defaults to `true` (it was effectively `false` since 4.8.0) and is configurable, also via the `PATROL_ANDROID_DONT_SUPPRESS_ACCESSIBILITY_SERVICES` dart-define. (#3227)
 
 ## 4.9.0
--- a/packages/patrol/CHANGELOG.md
+++ b/packages/patrol/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- Fix a failing test hanging the whole shard: `FlutterError.onError` was not restored when a test body threw, so the native runner could block on `runDartTest` until its timeout and every test queued behind it never ran. The failure is now reported, with its details. (#3243)
 - Android: keep third-party `AccessibilityService`s running during the test session again. `AndroidAutomatorConfig.dontSuppressAccessibilityServices` now defaults to `true` (it was effectively `false` since 4.8.0) and is configurable, also via the `PATROL_ANDROID_DONT_SUPPRESS_ACCESSIBILITY_SERVICES` dart-define. (#3227)
 
 ## 4.9.0
--- a/packages/patrol/lib/src/binding.dart
+++ b/packages/patrol/lib/src/binding.dart
@@ -243,9 +243,15 @@ class PatrolBinding extends LiveTestWidgetsFlutterBinding {
       }
     };
 
-    await testBody();
-
-    FlutterError.onError = previousOnError;
+    try {
+      await testBody();
+    } finally {
+      // Restore even if the body throws. Otherwise a later async error trips
+      // flutter_test's "overrode FlutterError.onError" invariant and the test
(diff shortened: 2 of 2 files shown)


Kendru98:
Closing this — after tracing the failure to its actual cause, this PR is not the right fix.

The hang was on our side. Our app registers Crashlytics error handlers that install their own `FlutterError.onError` and report without forwarding to the handler they replaced. That buries the binding's handler, `_pendingExceptionDetails` stays null, `flutter_test/src/binding.dart:1018` asserts, and the test future never completes — so `runDartTest` blocks until its timeout and the whole shard dies, reported as zero failures.

Patrol's documentation already covers this exactly, in *Initializing app inside a test*:

> DO NOT modify `FlutterError.onError`. Sometimes it is done by some monitoring tools (like Crashlytics). Those tools rely on intercepting errors by modifying `FlutterError.onError` callback and it causes that the test engine can't see any exceptions, thus can't end a test if it fails.

We were violating that. It became visible only recently, when a change on our side made handler registration unconditional instead of gated on analytics consent — under patrol there is no consent, so previously the handler was never installed.

Not installing those handlers under patrol fixes it on its own. Verified against unmodified patrol 4.9.0: the same 10 targets that previously executed 0 of 10 with both shards dead at the 2h cap now run 10/10 with no timeouts in about two minutes.

The `finally` here does make a failing test report again, but it does so by restoring the handler patrol captured — clobbering whatever the app installed afterwards. That is working around `FlutterError.onError` rather than fixing anything, and it would silently disable an app's crash reporting mid-test. Not something to land.

Sorry for the noise. Leaving the analysis here in case it helps anyone who hits the same symptom: a shard that dies at the runner timeout and reports `failures="0"` is worth checking against that docs rule first.

PR #3251, opened 2026-08-21
Add some macOS native automations

Polyphemus980:
Adds macOS platform automation via $.platform.macos: native tap/wait, NSAlert helpers, and tapMenu for nested application menu bar items.


dev/e2e_app/macos/Runner/AppDelegate.swift
```diff
@@ -3,6 +3,13 @@ import FlutterMacOS
 
 @main
 class AppDelegate: FlutterAppDelegate {
+  private(set) static var patrolMenuActionCount = 0
+
+  @IBAction
+  func patrolMenuAction(_ sender: Any?) {
+    Self.patrolMenuActionCount += 1
+  }
```
pdenert: Why do we need that?

packages/patrol/lib/src/platform/platform_automator.dart
```diff
@@ -57,15 +69,21 @@ class PlatformAutomatorConfig {
 
     /// Bundle identifier name of the application under test.
     ///
-    /// iOS only.
+    /// iOS only. For macOS, prefer `macosBundleId`.
```
pdenert: 'prefer' or just 'use'?

packages/patrol/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- Add macOS platform automation via `$.platform.macos`: native tap / wait, `NSAlert` helpers, and `tapMenu` for the application menu bar.
```
pdenert: You can list in changelog available options like it's done for the Playwright flags below

dev/e2e_app/macos/Runner/AppDelegate.swift
Polyphemus980: It was leftover from a test, deleted

packages/patrol/lib/src/platform/platform_automator.dart
Polyphemus980: True, it doesn't make sense to reuse bundleId from iOS. Changed to use

Kendru98:
After this macos config is needed, or it is still not needed? Im refering to https://github.com/leancodepl/patrol/pull/3230 @Polyphemus980

pdenert (approved):
LGTM

